PR #19, opened 2020-08-30
Implement UDS support and switch services to parse their endpoints from their configs

arsing:
Also change identityd to create a single instance of KS client, CS client and
KS openssl engine, instead of recreating them in each function that needed them.

---

See the changes to the three docs files for how to specify the endpoints.


cert/aziot-certd/src/config.rs
```diff
@@ -3,19 +3,21 @@
...
 	/// Configuration of how new certificates should be issued.
-	pub(crate) cert_issuance: CertIssuance,
+	pub cert_issuance: CertIssuance,
 
 	/// Map of preloaded certs from their ID to their location.
 	#[serde(default)]
-	pub(crate) preloaded_certs: std::collections::BTreeMap<String, PreloadedCert>,
+	pub preloaded_certs: std::collections::BTreeMap<String, PreloadedCert>,
+
+	pub endpoints: Endpoints,
```
gordonwang0, line 15: Missing doc comment? (Same in other config.rs files)

identity/aziot-identityd/src/settings.rs
```diff
@@ -9,17 +9,15 @@ use crate::error::InternalError;
```
gordonwang0, line 9: Is there a reason this file is called settings.rs in identityd but config.rs in certd/keyd? I feel like the file names should be consistent.

cert/aziot-certd/src/config.rs
arsing: Fixed.

identity/aziot-identityd/src/settings.rs
massand: This is a hangover from iotedged using settings.rs for parsing YAML. We can rename this to be consistent.

http-common/src/connector.rs
```diff
@@ -0,0 +1,335 @@
...
+#[derive(Debug)]
+pub enum Incoming {
+	Http(tokio::net::TcpListener),
+	Unix(tokio::net::UnixListener),
+}
+
+impl Connector {
+	pub fn new(uri: &url::Url) -> Result<Self, ConnectorError> {
+		match uri.scheme() {
+			"http" => {
```
massand, line 30: Can we put a test feature flag to enable http? I don't think we need to support http by default after UDS is enabled. 
This also reminds me - I should rip out the https protocol config in identityd. I can handle that in my UDS PR for identityd though.

massand (changes requested):
Http should be in test mode only

http-common/src/connector.rs
```diff
@@ -0,0 +1,335 @@
...
+
+impl<'de> serde::Deserialize<'de> for Connector {
+	fn deserialize<D>(deserializer: D) -> Result<Self, D::Error> where D: serde::de::Deserializer<'de> {
+		struct Visitor;
+
+		impl<'de> serde::de::Visitor<'de> for Visitor {
+			type Value = Connector;
+
+			fn expecting(&self, formatter: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+				formatter.write_str("Connector")
```
massand, line 170: According to `serde::de::Visitor` dev docs - 
> The message should not be capitalized and
    should not end with a period

http-common/src/connector.rs
```diff
@@ -0,0 +1,335 @@
...
+impl std::error::Error for ConnectorError {
+	fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+		Some(&*self.inner)
+	}
+}
+
+#[cfg(test)]
+mod tests {
+	#[test]
+	fn create_connector() {
```
massand, line 319: add negative test case for say ftp://

http-common/src/connector.rs
arsing: `http` is needed because `iotedge` uses reqwest to invoke it directly, which in turn is because there's no `Connector`-based client for IS. That's why I left the IS docs and default config to still use `http`. If you're going to add a `Connector`-based client in your PR, then you can remove `http` entirely in that PR.

http-common/src/connector.rs
arsing: Right, I misremembered what `#[derive(serde::Deserialize)]` expands to. It should be `"struct Connector"`. But anyway, `"an endpoint URI"` is better. Fixed.

http-common/src/connector.rs
arsing: Done, and added serde tests too.

identity/aziot-identityd/src/settings.rs
massand: @arsing - are you renaming this? If not, I can do this in the identityd UDS PR.

http-common/src/connector.rs
massand: I was initially thinking we could use it for E2E tests, but that isn't useful either. I can remove it after IS gets UDS. Thanks.

identity/aziot-identityd/src/settings.rs
arsing: I'm not.

We should probably consolidate some of the "entrypoint" stuff, like reading env vars for configs, env_logger, etc. But I talked with Cindy about a thing that would affect this anyway. Let's talk offline.

http-common/src/connector.rs
arsing: Yes, E2E can use UDS too.

We will need http/https eventually for Kubernetes, but let's worry about that when that happens.

PR #20, opened 2020-09-04
EST fix

massand:
Identity Service will attempt to create a new cert and keypair during provisioning

### DIFF
--- a/identity/aziot-identityd/src/lib.rs
+++ b/identity/aziot-identityd/src/lib.rs
@@ -203,7 +203,6 @@ impl Server {
 						aziot_identity_common::Credentials::SharedPrivateKey(device_id_pk)
 					},
 					settings::ManualAuthMethod::X509 { identity_cert, identity_pk } => { 
-						
 						self.create_identity_cert_if_not_exist_or_expired(&identity_pk, &identity_cert, &device_id).await?;
 						aziot_identity_common::Credentials::X509{identity_cert, identity_pk}
 					}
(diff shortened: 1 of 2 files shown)


identity/aziot-identityd/src/lib.rs
```diff
@@ -203,6 +203,34 @@ impl Server {
 						aziot_identity_common::Credentials::SharedPrivateKey(device_id_pk)
 					},
 					settings::ManualAuthMethod::X509 { identity_cert, identity_pk } => { 
+						let identity_pk_key_handle = self.key_client.create_key_pair_if_not_exists(&identity_pk, Some("rsa-2048:*")).await
+							.map_err(|err| Error::Internal(InternalError::CreateCertificate(Box::new(err))))?;
+								
+						let csr = {
```
arsing, line 209: As I said in the chat, you need to do this only if the cert doesn't already exist, or exists but has expired.

massand: My thought behind generating a new cert was as this - provisioning is a one-time operation and generating a new cert before each device provision, while not being necessary, isn't happening too often. 
I was planning on cracking open the cert and doing expiration validation when I implement the `/principal` API, because its the only other operation where the identity cert expiration matters.

arsing: DPS-X509 doesn't mean the cert *has* to be issued dynamically. When doing a manually-registered device, the cert's thumbprint is in DPS / IoT Hub already. Creating a new cert is wrong if it already exists and the user preloaded it.

massand: You are right. I forgot that create_cert does clobber the cert, even if the device-id cert is configured as self-signed (in which mode I mistakenly thought CS will ignore the CSR). Fixing. Thanks for clarifying.

arsing: In the case where the user wants the cert to be used as-is, there will be no cert_issuance specified for the cert in the first place, "self-signed" or otherwise. So it'll fail with the error that no issuance has been specified for this cert.

arsing (approved):

PR #21, opened 2020-09-04
Add IdentityClient trait and implement for Hub identities

gordonwang0:
Adds a common trait for local and hub identities and moves the Hub identity implementation to use that trait.

### DIFF
--- a/identity/aziot-identity-common/src/lib.rs
+++ b/identity/aziot-identity-common/src/lib.rs
@@ -14,38 +14,38 @@ pub struct GenId(pub String);
 #[serde(rename_all = "lowercase")]
 #[serde(tag = "type", content = "spec")]
 pub enum Identity {
-    Aziot(AzureIoTSpec),
+	Aziot(AzureIoTSpec),
 }
 
 #[derive(Clone, Debug, serde::Deserialize, serde::Serialize)]
 pub struct AzureIoTSpec {
-    #[serde(rename = "hubName")]
-    pub hub_name: String,
-    #[serde(rename = "deviceId")]
-    pub device_id: DeviceId,
-    #[serde(rename = "moduleId", skip_serializing_if = "Option::is_none")]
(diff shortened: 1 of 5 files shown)


identity/aziot-identity-common/src/lib.rs
```diff
@@ -70,81 +70,133 @@ pub struct IoTHubDevice {
...
+		#[serde(skip_serializing_if = "Option::is_none")]
+		pub authentication: Option<AuthMechanism>,
+	}
+}
+
+/// Common trait for identity clients.
+#[async_trait::async_trait]
+pub trait IdentityClient {
+	/// Type of identity, i.e. Hub or local.
+	type IdentityType;
```
massand, line 156: This appears to be the wrong layer for having traits. We don't need traits for the hub vs local clients, because there is a coincidental overlap in some APIs. We need to leverage the `Identity` enum to distinguish between the two representations, not associated types. From our conversations, my understanding was that a trait was needed for the higher-level `IdentityManager`, whose operations are invoked by the API layer.

identity/aziot-identity-common/src/lib.rs
```diff
@@ -70,81 +70,133 @@ pub struct IoTHubDevice {
...
+
+	/// Device context for constructor, unique per identity type.
+	type DeviceInfo;
+
+	/// Context for creating and updating identities, unique per identity type.
+	type IdentityInfo;
+
+	/// Constructor: create a new identity client.
+	fn new(
```
massand, line 165: This is an implementation specific constructors, which shouldn't be part of a trait.

massand (changes requested):
The trait was applied at the wrong layer.

massand (commented):
Thanks - the traits are defined at the right level. Let us define the API shape for Local Identity and then make a call on whether a trait might be useful.

gordonwang0:
Per our discussion yesterday, let's revisit this in the future.

PR #22, opened 2020-09-07
Add API version to all requests, and reduce HTTP server boilerplate

arsing:


gordonwang0 (approved):

http-common/src/server.rs
```diff
@@ -0,0 +1,319 @@
...
+												Ok(result) => result,
+												Err(err) => return Ok(err.to_http_response()),
+											};
+											http_common::server::json_response(status_code, response.as_ref())
+										},
+
+										http::Method::PUT => {
+											let content_type = headers.get(hyper::header::CONTENT_TYPE).and_then(|value| value.to_str().ok());
+											if content_type.as_deref() != Some("application/json") {
+												return Ok((http_common::server::Error {
```
massand, line 149: Is a JSON response required for all PUT requests per RFC?

massand (changes requested):

cert/aziot-cert-client-async/src/lib.rs
```diff
@@ -154,7 +160,7 @@ where
 		},
 
 		res_status_code if res_status_code.is_client_error() || res_status_code.is_server_error() => {
-			let res: aziot_cert_common_http::Error = serde_json::from_slice(&body).map_err(|err| std::io::Error::new(std::io::ErrorKind::Other, err))?;
+			let res: http_common::server::ErrorBody<'static> = serde_json::from_slice(&body).map_err(|err| std::io::Error::new(std::io::ErrorKind::Other, err))?;
```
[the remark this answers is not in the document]
arsing: Good point. Done.

http-common/src/server.rs
arsing: `s/JSON response/request body/` ?

If so, we don't have any route that has a `PUT` with no body. If we ever have one, we can make it an `Option` just like `POST`'s arm.

massand: I am good with that. Thanks.

massand (approved):

PR #23, opened 2020-09-11
Change IS config.toml idtype to vector

gordonwang0:


identity/aziot-identityd/src/lib.rs
```diff
@@ -379,14 +379,24 @@ impl auth::authorization::Authorizer for SettingsAuthorizer
 			crate::auth::OperationType::GetModule(m) => {
 				if let crate::auth::AuthId::LocalPrincipal(creds) = o.auth_id {
 					if let Some(p) = self.pmap.get(&crate::auth::Uid(creds.0)) {
-						return Ok(p.name.0 == m && p.id_type == Some(aziot_identity_common::IdType::Module))
+						let allow_id_type = match &p.id_type {
+							Some(i) => i.contains(&aziot_identity_common::IdType::Module),
```
CindyXing, line 383: We would also allow device type host process to query module identity. 
Does this line mean that device type can't?

gordonwang0: No, this just requires the principal to specify idtype = module in order to authorize.

CindyXing: I see. So for a host process to access both device and module identity, it needs to specify idtype = ["device", "module"]. Got you.

CindyXing (approved):

identity/aziot-identityd/src/main.rs
```diff
@@ -90,19 +90,24 @@ async fn run() -> Result<(), Box<dyn std::error::Error>> {
...
-    if let Some(v) = principal.as_ref() { v.iter()
-            .for_each(|p| {
-                pmap.insert(p.uid, p.clone());
-                if p.id_type == Some(aziot_identity_common::IdType::Module) { mset.insert(p.clone().name); }
-            })};
+				if let Some(id_type) = &p.id_type {
+					if id_type.contains(&aziot_identity_common::IdType::Module) {
+						mset.insert(p.clone().name);
+					}
+				}
```
massand, line 107: More idiomatically, we can do this - 
```
let mset: std::collections::BTreeSet<aziot_identity_common::ModuleId> = 
        principal
            .as_ref()
            .map_or(
            std::collections::BTreeSet::new(), 
                |v| v.iter().filter(|&p| p.id_type.clone().map_or(false, |t| t.contains(&aziot_identity_common::IdType::Module)))
                    .map(|p| p.name.clone()).collect::<std::collections::BTreeSet<_>>());

    let pmap: std::collections::BTreeMap<aziot_identityd::auth::Uid, aziot_identityd::settings::Principal> =
        principal
            .as_ref()
            .map_or(
                std::collections::BTreeMap::new(),
                |v| v.iter()
                .map(|p| (p.uid, p.clone())).collect::<std::collections::BTreeMap<_,_>>()
            );
```

identity/aziot-identityd/src/lib.rs
```diff
@@ -379,14 +379,24 @@ impl auth::authorization::Authorizer for SettingsAuthorizer
...
+							None => false,
+						};
```
massand, line 385: More idiomatically, this is better (for this instance and other instances below) - 
```
let allow_id_type = p.id_type.clone()
    .map_or(false, |i| i.contains(&aziot_identity_common::IdType::Module));
```

massand (changes requested):

identity/aziot-identityd/src/lib.rs
gordonwang0: Changed

identity/aziot-identityd/src/main.rs
gordonwang0: I got a clippy warning with the code in your comment, so made a small modification.

massand (approved):

PR #24, opened 2020-09-14
Add local ID settings struct

gordonwang0:


identity/aziot-identityd/src/error.rs
```diff
@@ -19,84 +19,87 @@ impl Error {
...
-    ParseSettings(toml::de::Error),
-    SaveDeviceInfo(std::io::Error),
-    SaveSettings(std::io::Error),
+	CreateCertificate(Box<dyn std::error::Error + Send + Sync>),
+	CreateHomeDir(std::io::Error),
+	InvalidUri(http::uri::InvalidUri),
+	LoadKeyOpenslEngine(openssl2::Error),
+	LoadDeviceInfo(std::io::Error),
+	LoadSettings(std::io::Error),
+	BadSettings(std::io::Error),
```
gordonwang0, line 60: Addition of BadSettings is the only non-spacing change in this file.

CindyXing: Ok. Are we using rustfmt when submitting PR?

CindyXing (approved):

identity/aziot-identityd/test/good_auth_settings.toml
```diff
@@ -4,6 +4,10 @@
 hostname = "iotedge"
 homedir = "/var/lib/aziot/identityd"
 
+# Local ID settings
```
massand, line 7: Please add a comment indicating that this is concatenated (suffixed) to id.name for producing the CN in the local identity's X.509 certificate.

gordonwang0: Done

massand (approved):

PR #25, opened 2020-09-16
Add makefile targets to build packages.

arsing:
This commit adds support for building deb packages for Debian and Ubuntu,
and rpm packages for CentOS.

A new script, `ci/package.sh` builds the packages. There is a CI job in
`packages.yaml` that invokes the script.

Other changes:

- Implement support for systemd socket activation by supporting the `fd://`
  scheme for listening endpoints.

- Remove manual dependency resolution from makefile.

  It was a pain to duplicate information that cargo already has, and redundant
  since cargo already no-ops if outputs are up-to-date.

  As a downside, if aziot-keys has a syntax error, that will lead to poor
  error messages from bindgen instead of nice errors from rustc, which
  the makefile-based method was a solution to. But in that case the developer
  can just temporarily switch to building with `cargo build -p aziot-keys`
  while they fix that.

- Use only the latest version of openssl for each OS in CI and scripts.
  We don't care about older versions of openssl.

- Change IS to take full DPS global endpoint URI instead of the URI without
  the scheme. This makes it consistent with iotedged, consistent with
  the default config file, consistent with other URIs in that and
  other config files, and allows for flexibility in the future.

- Merge softhsm tests CI job into main test CI job, since they both build
  the pkcs11-test binary.


docs/running/pkcs11/_index.md
```diff
@@ -216,34 +216,36 @@ KEY_3_TYPE='rsa-2048'
 # export PKCS11_SPY_PATH='/usr/lib64/pkcs11/pkcs11-spy.so'
 # export PKCS11_SPY_PATH='/usr/lib/arm-linux-gnueabihf/pkcs11/pkcs11-spy.so'
 
+PKCS11_TEST_PATH="$PWD/target/x86_64-unknown-linux-gnu/debug/pkcs11-test"
```
CindyXing, line 219: Nit pick, document the paths for arm as well.

ci/package.sh
```diff
@@ -0,0 +1,94 @@
...
+
+
+mkdir -p packages
+
+
+case "$CONTAINER_OS" in
+    'centos:7')
+        case "$ARCH" in
+            'arm32v7'|'aarch64')
+                echo 'Cross-compilation on CentOS 7 is not supported' >&2
```
CindyXing, line 17: Add more loggings can be useful in case of issue.

ci/install-build-deps.sh
```diff
@@ -1,62 +1,107 @@
 #!/bin/bash
 
 # This script is meant to be sourced.
+#
```
CindyXing, line 4: Instead of comments in the file, can we show the warning to users and prompt for continue?

contrib/debian/preinst
```diff
@@ -0,0 +1,61 @@
...
+		if ! /usr/bin/getent group aziotks >/dev/null; then
+			/usr/sbin/groupadd -r aziotks
+		fi
+		if ! /usr/bin/getent passwd aziotks >/dev/null; then
+			/usr/sbin/useradd -r -g aziotks -c 'aziot-keyd user' -s /sbin/nologin -d /var/lib/aziot/keyd aziotks
+		fi
+		mkdir -p /var/lib/aziot/keyd
+		;;
+
+	abort-upgrade)
```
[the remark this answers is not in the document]
arsing: This is from the default deb template, so we don't need to change it and don't need to worry about it.

contrib/debian/postinst
```diff
@@ -0,0 +1,61 @@
...
+		fi
+
+		for d in /var/lib/aziot/certd /var/lib/aziot/identityd /var/lib/aziot/keyd; do
+			if [ -d "$d" ]; then
+				chmod 0700 "$d"
+			fi
+		done
+		;;
+
+	abort-upgrade|abort-remove|abort-deconfigure)
```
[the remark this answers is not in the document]
arsing: This is from the default deb template, so we don't need to change it and don't need to worry about it.

ci/package.sh
arsing: The log already tells you what the problem is. What more do you want it to say?

ci/install-build-deps.sh
arsing: A script cannot know whether it's being sourced or not.

identity/aziot-identityd/config/unix/default.toml
```diff
@@ -4,38 +4,32 @@
...
+# device_id = "deviceid"
 
-[provisioning.authentication]
-method = "sas"
-device_id_pk = "device-id"
-dynamic_reprovisioning = false
+# [provisioning.authentication]
+# method = "sas"
+# device_id_pk = "device-id"
+# dynamic_reprovisioning = false
```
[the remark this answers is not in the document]
arsing: That feature does not exist in this repo.

.github/workflows/test.yaml
```diff
@@ -16,32 +16,118 @@ jobs:
...
-        - '1.1.0'
-        - '1.1.1'
+        - 'ubuntu:18.04'
+        - 'ubuntu:20.04'
+        arch:
+        - 'amd64'
+        - 'arm32v7'
+        - 'aarch64'
         exclude:
+        # TODO: support running CI tests for arm32v7 and aarch64
```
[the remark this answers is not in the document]
arsing: It's part of other tests tasks that are already on the board.

docs/running/pkcs11/_index.md
arsing: Done.

README.md
```diff
@@ -3,7 +3,17 @@ This repository contains the code of the Azure IoT Identity Service and related
...
+
+```sh
+yum install aziot-identity-service
+```
+
+## Debian, Ubuntu
+
+```sh
+apt install aziot-identity-service
+```
```
gordonwang0, line 16: Both these commands probably need sudo, but that should be obvious

contrib/centos/aziot-identity-service.spec
```diff
@@ -0,0 +1,146 @@
+# spec file for package aziot-identity-service
+#
+# Copyright (c) Microsoft. All rights reserved.
+
+Name: aziot-identity-service
+Version: @version@
+Release: @release@
+Summary: Azure IoT Identity Service and related services
+License: Proprietary
```
gordonwang0, line 9: Is this correct? The repo (and the license file below) are MIT.

arsing: Yeah, I copied this from iotedge and was confused by this too, because the license in `debian/copyright` is still MIT. It might be something to do with the source / product separation that we have with iotedge vs azure-iotedge, but then why doesn't it apply to the Debian package...

I'll follow up with Mike.

README.md
arsing: I think I'll just remove this section. There's a lot more that goes into installing and configuring them than just the distro package, and we'll have docs.microsoft.com for that.

arsing: Removed.

massand (approved):
Identity service specific changes LGTM

contrib/centos/aziot-identity-service.spec
arsing: I didn't hear from Mike and Dave doesn't remember, so I've made this MIT to match the Debian package.

CindyXing (approved):

CindyXing (commented):
Sounds good. Thanks Anarv. 

Given the open source of this repo was approved, it is right to mark as the MIT license.

gordonwang0 (approved):

PR #29, opened 2020-09-30
(iotedged) Add CLI flags to configure socket/url paths

daprilik:
While getting set up, I decided to lump all of my sockets + config files under `~/Work/fake_root` to avoid futzing about with permissions. Unfortunately, `iotedged` uses hard-coded socket paths / URLs to talk to the various daemons.

This PR removes the hard-coded socket paths / URLs from `iotedged` by adding a few more CLI flags.

### DIFF
--- a/iotedged/src/main.rs
+++ b/iotedged/src/main.rs
@@ -17,16 +17,16 @@ async fn main() -> Result<(), Error> {
 	let Options {
 		hub_id,
 		device_id,
-		keyd_url,
-		certd_url,
-		identityd_url,
+		keyd_uri,
+		certd_uri,
+		identityd_uri,
 		preloaded_device_id_ca_cert,
 		preloaded_device_id_cert,
 		sas_key,
 	} = structopt::StructOpt::from_args();
 
 
-	let key_connector = http_common::Connector::new(&keyd_url).unwrap();
+	let key_connector = http_common::Connector::new(&keyd_uri).unwrap();
 	let mut key_engine = {
 		let key_client = aziot_key_client::Client::new(aziot_key_common_http::ApiVersion::V2020_09_01, key_connector.clone());
 		let key_client = std::sync::Arc::new(key_client);
@@ -42,7 +42,7 @@ async fn main() -> Result<(), Error> {
 	};
 
 	let cert_client = {
-		let cert_connector = http_common::Connector::new(&certd_url).unwrap();
+		let cert_connector = http_common::Connector::new(&certd_uri).unwrap();
 		let cert_client = aziot_cert_client_async::Client::new(aziot_cert_common_http::ApiVersion::V2020_09_01, cert_connector);
 		let cert_client = std::sync::Arc::new(cert_client);
 		cert_client
@@ -326,7 +326,7 @@ async fn main() -> Result<(), Error> {
 	let body = serde_json::json! {{ "type": "aziot" }};
 	let client = reqwest::Client::new();
 	let res =
-		client.post(identityd_url.join("/identities/device").unwrap())
+		client.post(identityd_uri.join("/identities/device").unwrap())
 		.json(&body)
 		.send()
 		.await.map_err(Error::Reqwest)?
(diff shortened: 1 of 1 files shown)


iotedged/src/main.rs
```diff
@@ -378,6 +381,18 @@ struct Options {
...
 
+	/// `keyd` API endpoint
+	#[structopt(long, default_value = "unix:///var/lib/aziot/keyd.sock")]
+	keyd_url: url::Url,
+
+	/// `certd` API endpoint
+	#[structopt(long, default_value = "unix:///var/lib/aziot/certd.sock")]
+	certd_url: url::Url,
+
+	/// `identityd` API endpoint
```
arsing, line 392: Would be useful to add a note here that it has to be an `http://` URI. Not just because it gets used with `reqwest`, but because the URL concatenation on line 339 etc only works for URIs handled by `hyper::HttpConnector`.

iotedged/src/main.rs
```diff
@@ -378,6 +381,18 @@ struct Options {
 	#[structopt(long)]
 	device_id: String,
 
+	/// `keyd` API endpoint
+	#[structopt(long, default_value = "unix:///var/lib/aziot/keyd.sock")]
+	keyd_url: url::Url,
```
arsing, line 386: To fit with the naming scheme, these are "endpoints" (as in the config files) or "URIs" (as in the code), not "URLs".

arsing, line 386: Update `Cargo.toml` to enable the `"serde"` feature on the `url` dependency. It'll compile fine already because one of the other deps enables it, but it's good to be explicit.

arsing (changes requested):

iotedged/src/main.rs
daprilik: Yeah, I considered adding that comment, but reconsidered given that it should be able to run over UDS at some point anyways. I'll add it in though.

Also, AFAIK, `Url::join` works fine regardless of the underlying Connector (the method is implemented on `Url` directly, and isn't generic .

iotedged/src/main.rs
daprilik: IIRC, structopt uses `FromStr` to parse custom data structures, not serde, so I don't think this should be necessary

arsing: Oh right, derp.

iotedged/src/main.rs
arsing: >Also, AFAIK, `Url::join` works fine regardless of the underlying Connector (the method is implemented on `Url` directly, and isn't generic.

No, my point is the unix connector needs to use `http://foo/rest/of/the/uri` as opposed to `unix:///path/to/socket/rest/of/the/uri`, as you can see in https://github.com/Azure/iot-identity-service/blob/fe4a7a7fb5378558a99469150fb8b2c0ca5ba913/key/aziot-key-client-async/src/lib.rs#L41

arsing (approved):

iotedged/src/main.rs
daprilik: uhhh, not gonna lie, I'm not _entirely_ sure what's going on here, but I'll take your word for it.

arsing: `unix:///path/to/socket/rest/of/the/uri` is ambiguous, yes? There is a need to be able to know what the socket path is and what the URI is.

`hyperlocal`'s connector encodes the socket path in the hostname by converting it to hex, but the downside is you need to know it's being used because you have to use its special function to construct the URI with the encoded hostname. So you can't just concatenate the base URI with the request URI path.

This repo uses its own connector instead of `hyperlocal` that takes in the socket path at construction time instead, so it doesn't need the hostname to tell it what the socket path is. So the actual value of the hostname is irrelevant and can just be `foo`. But still, this too doesn't let you just concatenate the base URI with the request URI path.

daprilik: Ahhhh, gotcha. I didn't know that. Makes sense though.

PR #30, opened 2020-09-30
Add ProvisioningType none to Identity Service

gordonwang0:
Adds the option to skip provisioning with IoT Hub for devices that only use local identities.
```
[provisioning]
source = "none"
```

Calls to IS APIs when unprovisioned will return HTTP 400 with the error "device not found".


identity/aziot-identityd/src/lib.rs
```diff
@@ -362,7 +363,10 @@ impl Server {
 					}
 				};
 				device
-			}
+			},
+			settings::ProvisioningType::None => {
+				return Err(Error::DeviceNotFound);
```
massand, line 368: If this returns an error, the `main function` (or `reprovision_device` function) will fail fast, which is not what we want. We want the opposite effect for the IS process with the introduction of the `None` provisioning type, IIRC. The `provision_device` needs to be modified to return an enum (with possible values of `NoProvisioning` and `Provisioned`(IoTHubDevice) instead of returning a `IoTHubDevice` struct with `"none"` values .

identity/aziot-identityd/src/main.rs
```diff
@@ -45,29 +45,43 @@ async fn run() -> Result<(), Box<dyn std::error::Error>> {
...
-		log::info!("Provisioning complete.");
+		let curr_device_info_state;
 
-		let curr_device_info_state = aziot_identityd::settings::DeviceInfo { hub_name: device.iothub_hostname, device_id: device.device_id };
+		if let aziot_identityd::settings::ProvisioningType::None = server_.settings.provisioning.provisioning {
+			log::info!("Skipping provisioning with IoT Hub.");
 
-		if prev_device_info_path.exists() {
-			let prev_device_info_state = aziot_identityd::settings::DeviceInfo::new(&prev_device_info_path)?;
+			curr_device_info_state = aziot_identityd::settings::DeviceInfo {
```
massand, line 54: We shouldn't do this, otherwise we will return invalid information to callers of `/identities/device`

massand (changes requested):

identity/aziot-identityd/src/lib.rs

identity/aziot-identityd/src/main.rs
gordonwang0: Fixed

identity/aziot-identityd/src/main.rs
```diff
@@ -45,31 +45,43 @@ async fn run() -> Result<(), Box<dyn std::error::Error>> {
 	let server = aziot_identityd::Server::new(settings, authenticator, authorizer)?;
 	let server = std::sync::Arc::new(futures_util::lock::Mutex::new(server));
 	{
-		log::info!("Provisioning starting..");
-
 		let mut server_ = server.lock().await;
-		let device = server_.provision_device().await?;
+		let device_status;
```
massand, line 49: You don't need to declare this here. See below

identity/aziot-identityd/src/main.rs
```diff
@@ -45,31 +45,43 @@ async fn run() -> Result<(), Box<dyn std::error::Error>> {
...
+
+		log::info!("Provisioning starting..");
+		let provisioning = server_.provision_device().await?;
 		log::info!("Provisioning complete.");
 
-		let curr_device_info_state = aziot_identityd::settings::DeviceInfo { hub_name: device.iothub_hostname, device_id: device.device_id };
+		if let aziot_identity_common::ProvisioningStatus::Provisioned(device) = provisioning {
```
massand, line 55: You can assign the result of a block's execution like so 
```suggestion
		let device_state = { if let aziot_identity_common::ProvisioningStatus::Provisioned(device) = provisioning {
```

identity/aziot-identityd/src/main.rs
```diff
@@ -45,31 +45,43 @@ async fn run() -> Result<(), Box<dyn std::error::Error>> {
...
-			let () = server_.init_identities(prev_module_set, module_set).await?;
-		}
+			log::info!("Identity reconciliation with IoT Hub complete.");
 
-		log::info!("Identity reconciliation complete.");
+			device_status = toml::to_string(&curr_hub_device_info)?;
+		}
+		else {
+			device_status = aziot_identityd::settings::HubDeviceInfo::unprovisioned();
+		}
```
massand, line 82: And you can end the block here, returning instead of assigning the returned value like so
```suggestion
			toml::to_string(&curr_hub_device_info)?;
		}
		else {
			aziot_identityd::settings::HubDeviceInfo::unprovisioned();
		}
	}
```

massand (changes requested):
Code style comment

identity/aziot-identityd/src/main.rs
gordonwang0: Good catch. fixed.

massand (approved):

PR #31, opened 2020-10-01
Rename `KEYGEN_*` things exported from libaziot-keys to `AZIOT_KEYS_*`

arsing:
4c6a200f6f92592d2c3560349c19d7421a042b6d renamed many things as part of
renaming libiothsm-keygen to libaziot-keys, but forgot the names of
the exported items. This commit finishes the job.

One special case is `KEYGEN_get_function_list` - it was renamed not to
`AZIOT_KEYS_get_function_list` but to `aziot_keys_get_function_list`.
The upper-case prefix was incorrect for function names.

One more related change in this commit is to stop compiling
the `cbindgen_unused_` functions into the final libaziot-keys cdylib.
These fns are a hack to get cbindgen to emit structs that are otherwise
not referenced by any function. But before this change, these functions would
still end up being compiled into the final binary.

### DIFF
--- a/key/aziot-keyd/src/keys.rs
+++ b/key/aziot-keyd/src/keys.rs
@@ -2,14 +2,14 @@
 
 #[allow(clippy::module_name_repetitions)]
 #[derive(Clone, Copy, Debug)]
-pub struct KeysRawError(pub(crate) sys::KEYGEN_ERROR);
+pub struct KeysRawError(pub(crate) sys::AZIOT_KEYS_STATUS);
 
 impl std::fmt::Display for KeysRawError {
 	fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
 		match self.0 {
-			sys::KEYGEN_ERROR_FATAL => f.write_str("KEYGEN_ERROR_FATAL"),
-			sys::KEYGEN_ERROR_INVALID_PARAMETER => f.write_str("KEYGEN_ERROR_INVALID_PARAMETER"),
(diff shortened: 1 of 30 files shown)


gordonwang0 (approved):

CindyXing (approved):

key/aziot-keyd/src/keys.rs
```diff
@@ -798,26 +798,26 @@ impl std::fmt::Display for DecryptError {
 impl std::error::Error for DecryptError {
 }
 
-fn keys_fn(f: impl FnOnce() -> sys::KEYGEN_ERROR) -> Result<(), KeysRawError> {
-	match f() {
-		sys::KEYGEN_SUCCESS => Ok(()),
+fn keys_ok(result: sys::AZIOT_KEYS_ERROR) -> Result<(), KeysRawError> {
+	match result {
+		sys::AZIOT_KEYS_SUCCESS => Ok(()),
```
massand, line 803: The naming on the result type is misleading. It should AZIOT_KEYS_STATUS since one of the return values is AZIOT_KEYS_SUCCESS and the rest are AZIOT_KEYS_ERROR.

massand (changes requested):
While we are refactoring, the AZIOT_KEYS_ERROR nomenclature can be fixed

key/aziot-keyd/src/keys.rs
arsing: Good point. Fixed.

massand (approved):

PR #32, opened 2020-10-03
Fix package scripts to add CS user to KS group and IS user to KS and CS groups.

arsing:


gordonwang0 (approved):

contrib/centos/aziot-identity-service.spec
```diff
@@ -78,29 +78,29 @@ make -j \
...
+fi
+
 if ! %{_bindir}/getent group aziotcs >/dev/null; then
     %{_sbindir}/groupadd -r aziotcs
 fi
 if ! %{_bindir}/getent passwd aziotcs >/dev/null; then
     %{_sbindir}/useradd -r -g aziotcs -c 'aziot-certd user' -s /sbin/nologin -d /var/lib/aziot/certd aziotcs
+    %{_sbindir}/usermod -aG aziotks aziotcs
 fi
-mkdir -p /var/lib/aziot/certd
```
massand, line 87: Please double check this line can be removed. According to the RPM packaging guide, `%dir` only owns the folder but folder still needs to be created. Thanks.

arsing: I confirmed installing the package does create the homedirs.

massand (approved):

PR #33, opened 2020-10-05
Add management of local identities

gordonwang0:
Currently, `convert_to_map` parses settings and returns a set of module IDs that have Hub identities. This PR modifies this function to also return a set of local identities.

Also changes the `main` function to always parse the previous settings regardless of provisioning state (needed for unprovisioned devices that only use local identities).


identity/aziot-identityd/src/main.rs
```diff
@@ -60,26 +70,23 @@ async fn run() -> Result<(), Box<dyn std::error::Error>> {
...
 
-				if let Some(prev_state) = prev_hub_device_info {
-					let mut prev_module_set: std::collections::BTreeSet<aziot_identity_common::ModuleId> = std::collections::BTreeSet::default();
-					if prev_state.eq(&curr_hub_device_info) && prev_settings_path.exists() {
-						let settings = aziot_identityd::settings::Settings::new(&prev_settings_path)?;
-						let (_, p) = convert_to_map(&settings.principal);
-						prev_module_set = p;
+				if let Some(prev_device) = prev_hub_device_info {
+					if prev_device == curr_hub_device_info {
+						let () = server_.init_hub_identities(prev_hub_mset, hub_mset).await?;
```
massand, line 75: `init_hub_identities` should happen regardless of the `prev_device` and `curr_hub_device_info` match. The prev_module_set was only set if `prev_device` and `curr_hub_device_info` matched.

identity/aziot-identityd/src/main.rs
```diff
@@ -60,26 +70,23 @@ async fn run() -> Result<(), Box<dyn std::error::Error>> {
...
 
-			log::info!("Identity reconciliation with IoT Hub complete.");
-
 			toml::to_string(&curr_hub_device_info)?
 		}
 		else {
 			aziot_identityd::settings::HubDeviceInfo::unprovisioned()
 		};
 
+		let () = server_.init_local_identities(prev_local_mset, local_mset).await?;
```
massand, line 87: I cannot review this logic until I see how you are handling reconciliation in init_local_identities. I think you need to be aware of your EST endpoint before reconciling your local identities on it. Otherwise, identity service will clobber all previous created identities after a server switch (which is likely not what we want).

massand (changes requested):
I haven't seen any reconciliation logic yet. Please open a PR when that part is ready because it is key to understanding what we should do with old local identities.

identity/aziot-identityd/src/main.rs
gordonwang0: Pushed the commits with the reconciliation logic.

identity/aziot-identityd/src/main.rs
gordonwang0: Fixed

gordonwang0: Note that I've changed the logic so that init_hub_identities will be called if the device is currently provisioned. Before, it was called if the device was provisioned AND the device was previously provisioned.

identity/aziot-identityd/src/lib.rs
```diff
@@ -430,28 +488,78 @@ impl Server {
...
 	}
 }
 
+fn deconstruct_local_opts(options: Option<settings::LocalIdOpts>, module_id: String) ->
+	(aziot_identity_common::LocalIdAttr, String, String) {
+	let defaults = (aziot_identity_common::LocalIdAttr::default(), module_id.clone(), module_id.clone());
+
+	options.map_or(defaults, |opts|
+		match opts {
+			settings::LocalIdOpts::X509 {attributes, cert_id, key_id,} => {
```
massand, line 511: Since module names are unique, can we not have the key id and cert id specified as options?

identity/aziot-identityd/src/main.rs
```diff
@@ -57,21 +67,24 @@ async fn run() -> Result<(), Box<dyn std::error::Error>> {
 				device_id: device.device_id
 			};
 
+			// Only consider the previous Hub modules if the current and previous Hub devices match.
+			let mut use_prev = false;
```
massand, line 71: The prev_hub_mset can be set directly where this flag is set. We don't need this right?

identity/aziot-identityd/src/lib.rs
gordonwang0: Key id and cert id are specified as options. The module name is used for both key id and cert id if either is not specified.

identity/aziot-identityd/src/main.rs
gordonwang0: Both prev_local_mmap and prev_hub_mset are set from the call to convert_to_map on line 51. prev_local_mmap is needed regardless of provisioning status, while prev_hub_mset is only needed for provisioned devices. So prev_hub_mset is generated regardless of whether it's needed because device provisioning status isn't known until after convert_to_map is called.

IMO, it's more efficient (the list of principals only needs to iterate once) to generate them both from a single convert_to_map and ignore prev_hub_mset using this flag if it's not needed.

identity/aziot-identityd/src/lib.rs
massand: My suggestion was for not making key_id and cert_id configurable because I am not aware of any scenario where CS and KS are preloaded.

gordonwang0: Ah got it. Ok, I'll remove the options.

massand (approved):
:shipit:

PR #35, opened 2020-10-07
Add designs and reorganize documents into github.io style docs

massand:


arsing (approved):

docs/keys-service.md
```diff
@@ -0,0 +1,333 @@
...
+
+#### RSA-PKCS1
+
+```json
+{
+    "keyHandle": "string",
+    "parameters": {
+        "algorithm": "RSA-PKCS1",
+    },
+    "plaintext": "base64-encoded-string"
```
micahl, line 275: This should be "ciphertext", right?

docs/keys-service.md
```diff
@@ -0,0 +1,333 @@
...
+
+#### RSA-NO-PADDING
+
+```json
+{
+    "keyHandle": "string",
+    "parameters": {
+        "algorithm": "RSA-NO-PADDING",
+    },
+    "plaintext": "base64-encoded-string"
```
micahl, line 287: Same as above

docs/keys-service.md
```diff
@@ -0,0 +1,333 @@
...
+
+Only valid for ECDSA keys.
+
+Note that the request takes the message digest, ie it must be calculated by the client.
+
+```json
+{
+    "keyHandle": "string",
+    "parameters": {
+        "algorithm": "ECDSA",
```
micahl, line 158: Is there a change planned to have 'algorithm' in the 'parameters'?  It appears to currently be a sibling

docs/keys-service.md
massand: Fixed

docs/keys-service.md
massand: Fixed

docs/keys-service.md
massand: It should not be the case - although we have been primarily using HMAC-SHA256 for IoT Hub keys. Please file an issue on this repo for this.

arsing: `algorithm` is a sibling of `parameters`. What Micah wrote is correct.

PR #37, opened 2020-10-07
Avoid unnecessary IoT Hub call to list modules

damonbarry:
In `identity::Server::init_identities(...)`, we query IoT Hub for a list of modules to decide whether any identities should be added or removed. But if the function arguments (`prev_module_set` and `current_module_set`) are both empty, there's no need to make the cloud API call because the function is effectively a no-op.

This change adds an early-exit check to the function, to avoid the cloud API call when it's not necessary.


identity/aziot-identityd/src/lib.rs
```diff
@@ -504,3 +508,61 @@ impl auth::authorization::Authorizer for SettingsAuthorizer
...
+				},
+			},
+			localid: Default::default(),
+		}
+	}
+
+	#[test]
+	fn init_identities_with_empty_args_exits_early() {
+		let server = Server::new(
+			make_empty_settings(),
```
damonbarry, line 561: This test is a bit underhanded in that it relies on an empty config to ensure cloud API calls fail (it assumes that if it somehow got past the early exit, _something_ will trip over the empty config before the function completes). A better option would be to create a nullable version of `aziot_hub_client_async::Client`, but what I have here is probably a lot less work.

CindyXing (approved):
Nice to see the test.

identity/aziot-identityd/src/lib.rs
```diff
@@ -504,3 +508,61 @@ impl auth::authorization::Authorizer for SettingsAuthorizer
...
+			Box::new(|_| Ok(AuthId::Unknown)),
+			Box::new(|_| {Ok(true)})
```
massand, line 563: I know this closure is similar to the one in main.rs. Could you please change this and the main.rs instance to look like this? Thanks.
```suggestion
			Box::new(|_| Ok(true))
```

massand (changes requested):
LGTM - waiting on a nit

identity/aziot-identityd/src/lib.rs
damonbarry: Yeah this was copy/paste. I'll change them...

identity/aziot-identityd/src/lib.rs
```diff
@@ -504,3 +508,61 @@ impl auth::authorization::Authorizer for SettingsAuthorizer
...
+				aziot_keyd: Connector::Fd {
+					original_specifier: Default::default(),
+					fd: Default::default(),
+				},
+			},
+			localid: Default::default(),
+		}
+	}
+
+	#[test]
```
arsing, line 558: https://docs.rs/tokio/0.2.22/tokio/attr.test.html

And then you don't need the macro that wraps `block_on`.

Or is there a reason you want to use `tokio_test` instead?

arsing (changes requested):

identity/aziot-identityd/src/lib.rs
damonbarry: I just did a google search and that's what turned up. What you have looks way better, fixed.

identity/aziot-identityd/src/lib.rs
damonbarry: Fixed

identity/aziot-identityd/src/lib.rs
```diff
@@ -555,14 +549,14 @@ mod tests {
...
+	#[tokio::test]
+	async fn init_identities_with_empty_args_exits_early() {
 		let server = Server::new(
 			make_empty_settings(),
 			Box::new(|_| Ok(AuthId::Unknown)),
 			Box::new(|_| Ok(true))
 		).unwrap();
 
-		assert!(wait!(server.init_identities(BTreeSet::new(), BTreeSet::new())).is_ok());
+		assert!(server.init_identities(BTreeSet::new(), BTreeSet::new()).await.is_ok());
```
arsing, line 560: You can get a better panic message with `let _ = result.unwrap();` compared to `assert!(result.is_ok());`, because the panic message will include the `Err` value.

arsing (approved):

identity/aziot-identityd/src/lib.rs
damonbarry: Yeah that's better. I'll change it.
```
---- tests::init_identities_with_empty_args_exits_early stdout ----
thread 'tests::init_identities_with_empty_args_exits_early' panicked at 'called `Result::unwrap()` on an `Err` value: DeviceNotFound', identity/aziot-identityd/src/lib.rs:563:28
```

damonbarry: Updated

massand (approved):

PR #39, opened 2020-10-08
Ensure aziot-keys is built before aziot-keyd.

arsing:
aziot-keyd requires aziot-keys to have been built first so that the linker
can link it in. But before this change this was not guaranteed to happen.

This went unnoticed until now because the parallel build would usually finish
building aziot-keys before it got around to linking aziot-keyd.
But Micah happened to have a machine that ended up linking aziot-keyd first
and complained about having no `libaziot_keys.so` to link to.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -310,6 +310,7 @@ version = "0.1.0"
 dependencies = [
  "aziot-key-common",
  "aziot-key-common-http",
+ "aziot-keys",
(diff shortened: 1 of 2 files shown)


damonbarry (approved):

gordonwang0 (approved):

massand (approved):
:shipit:

PR #41, opened 2020-10-09
Disallow custom endpoints and TCP servers in release builds.

arsing:
We want to ensure production devices use unix sockets for the transport
in order to have functional authentication of the caller. Furthermore,
in order for host modules to be able to reach the services, the endpoints need
to be in fixed locations. Otherwise every host module would need to provide
its own configurations that the user would have to keep in sync with
any customization they made to the services' endpoints. (Unlike with iotedged,
there is no env var that could give this information to an arbitrary
host module, because host modules run independently of our services.)

To that end, the endpoints configuration for all services will be ignored
in release builds, and the hard-coded defaults of unix sockets will be used.
Sockets inherited via systemd socket activation will also be checked to be
unix sockets.

Custom endpoints can still be configured for debug builds, so local development
and testing can still use unix sockets in custom locations, or TCP sockets.

The sockets directory has been standardized to be `/run/aziot`, rather than
the mix of `/var/lib/aziot` on distros that did not use
systemd socket activation (CentOS 7) and `/var/run/aziot` on distros that did
(everything else). All our supported distros already have `/run`,
so there's no reason to continue using `/var/run`.

Next, this commit enables systemd socket activation on CentOS 7. The original
decision to disable socket activation on CentOS 7 was inherited from iotedged,
which does it because CentOS 7 doesn't support `LISTEN_FDNAMES` and thus
cannot be used for socket-activated services that want more than one socket.
This is the case for iotedged with its mgmt and workload sockets, but is not
the case for the services in this repo. With this commit, socket activation
only uses the `LISTEN_FDS` env var now, and uses the first fd as the socket
that the service should bind to.

Because the endpoints are not configurable, this means the servers can
no longer use *either* systemd socket activation *or* bind a socket themselves.
They must do the former and *fall back* to the latter if the systemd env vars
haven't been set. This is how socket-activated services should work anyway,
so this only makes our services consistent with other software.
Therefore this commit removes the `fd://` URI scheme entirely, and
the default configuration value is the `unix://` URI. If there is
no inherited socket as described above, then the service falls back to
the endpoint URI in the configuration.

Thus, putting all these changes together, the default endpoint URIs are
`unix:///run/aziot/<service_name>.sock` on all distros.

Lastly, the aziot-key-openssl-engine-shared engine no longer uses
the `AZIOT_KEY_SERVICE_ENDPOINT` env var, and expects to connect to
`/run/aziot/keyd.sock`


http-common/src/connector.rs
```diff
@@ -173,7 +81,18 @@ impl Connector {
...
 	pub async fn incoming(self) -> std::io::Result<Incoming> {
 		match self {
-			Connector::Fd { fd, .. } => {
+			Connector::Fd { fallback } => {
+				let fd = match get_systemd_socket() {
+					Ok(Some(fd)) => fd,
+
+					Ok(None) => {
+						let fallback_incoming: std::pin::Pin<Box<dyn std::future::Future<Output = _>>> = Box::pin(fallback.incoming());
+						return fallback_incoming.await;
```
daprilik, line 90: Why not just `return fallback.incoming().await;`?

http-common/src/connector.rs
```diff
@@ -461,25 +402,102 @@ impl std::error::Error for ConnectorError {
...
+	//
+	// >sd_listen_fds parses the number passed in the $LISTEN_FDS environment variable, then sets the FD_CLOEXEC flag
+	// >for the parsed number of file descriptors starting from SD_LISTEN_FDS_START. Finally, it returns the parsed number.
+	//
+	// Note that this function always returns the first fd. It cannot be used for processes which expect more than one socket.
+	// CS/IS/KS only expect one socket, so this is fine, but it is not the case for iotedged (mgmt and workload sockets) for example.
+	//
+	// If obtaining more than one fd is required in the future, keep in mind that it requires getting fds by name (by inspecting the LISTEN_FDNAMES env var)
+	// instead of by number, since systemd does not pass down multiple fds in a deterministic order. The complication with LISTEN_FDNAMES is that
+	// CentOS 7's systemd is too old and doesn't support it, which would mean CS/IS/KS would have to forego systemd socket activation on CentOS 7
```
gordonwang0, line 419: Very nit (I'm sorry): The correct word here should be "forgo". "forgo" = "to do without"; "forego" = "to precede"

Although I just googled it and it seems like enough people mix them up that "forego" is now defined as a variant spelling of "forgo".

http-common/src/connector.rs
arsing: Recursive futures would be infinitely-sized.

http-common/src/connector.rs
```diff
@@ -214,6 +141,49 @@ impl Connector {
 	}
 }
 
+impl std::fmt::Display for Connector {
+	fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+		let url = match self {
+			Connector::Fd { fallback, .. } => {
```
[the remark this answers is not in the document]
arsing: Of all the things clippy lints about, it doesn't lint about this one huh.

Fixed.

http-common/src/connector.rs
arsing: TIL. Fixed.

http-common/src/connector.rs
daprilik: ahhhh, didn't see it was recursive.

I take it that's why you've got an explicit type annotation (to coerce it to a `dyn` Future)?

arsing: Yeah.

`cert/aziot-certd/src/lib.rs` has another example of doing a recursive async fn in its `fn create_cert` with a different approach, but I think this approach is better. I might switch that one to this approach too if I ever edit it for another reason.

cert/aziot-certd/src/config.rs
```diff
@@ -349,8 +359,36 @@ aziot_certd = "unix:///var/run/aziot/certd.sock"
 			].into_iter().collect(),
 
 			endpoints: super::Endpoints {
-				aziot_certd: http_common::Connector::new(&"unix:///var/run/aziot/certd.sock".parse().unwrap()).unwrap(),
-				aziot_keyd: http_common::Connector::new(&"unix:///var/run/aziot/keyd.sock".parse().unwrap()).unwrap(),
+				aziot_certd: http_common::Connector::Fd {
+					fallback: Box::new(http_common::Connector::Unix { socket_path: std::path::Path::new("/run/aziot/certd.sock").into() }),
+				},
+				aziot_keyd: http_common::Connector::Unix { socket_path: std::path::Path::new("/run/aziot/keyd.sock").into() },
```
massand, line 365: Where do we `set_permissions` on `/run/aziot` for IS/KS/CS groups?

http-common/src/connector.rs
damonbarry: TIL too! 😄

iotedged/src/main.rs
```diff
@@ -382,11 +382,11 @@ struct Options {
 	device_id: String,
 
 	/// `keyd` API endpoint.
-	#[structopt(long, default_value = "unix:///var/lib/aziot/keyd.sock")]
+	#[structopt(long, default_value = "unix:///run/aziot/keyd.sock")]
```
massand, line 385: It seems unlikely because /run is `rwx` to `root` user only. But, what's the behavior if `/run/aziot` already exists? I assume the service will fail to start?

massand (changes requested):
Questions about permissions

cert/aziot-certd/src/config.rs
arsing: That's systemd's job. The `/run/foo` directories are `r-x` to world, and the socket files themselves are ACLed based on the `SocketUser` and `SocketGroup` of the individual socket unit.

```sh
$ ls -l /run/ | grep aziot

drwxr-xr-x.  2 root root   60 Oct  9 00:26 aziot


$ ls -l /run/aziot/

total 0
srw-rw----. 1 aziotcs aziotcs 0 Oct  9 11:29 certd.sock
srw-rw----. 1 aziotid aziotid 0 Oct  9 11:29 identityd.sock
srw-rw----. 1 aziotks aziotks 0 Oct  9 11:29 keyd.sock
```

iotedged/src/main.rs
arsing: For socket-activation, the directory and the .sock itself will have been created by systemd, as I wrote in the other comment.

For non-socket-activation, the user has to ensure it's been created to be `rwx` to the three users, and `r-x` to the three groups. The socket files themselves will have the correct permissions when the services bind them, because they'll follow the user and group of the service, which is set in the service file (the group isn't, but the primary group of the user was set at user-creation time).

massand (approved):
:shipit:

arsing:
Daniel suggested a better way of doing this by eliminating the `fd://` scheme entirely. What do you think?

massand:
@arsing I am ok with the rest of the changes. I am also ok with removing the fallback via query string on an `fd://`. Signing off with the assumption that CentOS 7 works with systemd the same way as CentOS 8.

http-common/src/connector.rs
```diff
@@ -172,45 +66,92 @@ impl Connector {
...
-				let sock_addr = nix::sys::socket::getsockname(fd).map_err(|err| std::io::Error::new(std::io::ErrorKind::Other, err))?;
-				match sock_addr {
-					nix::sys::socket::SockAddr::Inet(_) => {
-						let listener = unsafe { std::os::unix::io::FromRawFd::from_raw_fd(fd) };
-						let listener = tokio::net::TcpListener::from_std(listener)?;
+		let systemd_socket = get_systemd_socket().map_err(|err| std::io::Error::new(std::io::ErrorKind::Other, err))?;
+		if let Some(fd) = systemd_socket {
+			let sock_addr = nix::sys::socket::getsockname(fd).map_err(|err| std::io::Error::new(std::io::ErrorKind::Other, err))?;
+			match sock_addr {
+				nix::sys::socket::SockAddr::Inet(_) => {
```
massand, line 73: Do we support TCP socket configuration for systemd for release builds?

arsing: Good point. Fixed.

massand:
:shipit:

damonbarry (approved):
I'm not familiar with the specifics of packaging and socket activation, but this PR looks fine to me. I like that we're [1] making config simpler for users, and [2] aligning to common conventions.

gordonwang0 (approved):

PR #44, opened 2020-10-13
Add local identities to get_identity

gordonwang0:
Modifies the Identity Service get_identity API (`/identities/modules/moduleId`) to return a module's local identity (if any).


docs/identity-service.md
```diff
@@ -229,37 +229,43 @@ The response depends on the principal used to authenticate with the Identity Ser
...
-    "hubname": "myhub.net",
-    "deviceid": "device01",
-    "moduleid": "module01",
-    "genid": "12345",
-    "auth": {
-        "type": "sas",
-        "keyHandle": "string"
-   }
-  }
+  "identities":[
```
massand, line 232: There are two issues with making this array - 1) it breaks the MR interface which expects one API (this can be fixed although I would like a schema to reflect there is one identiyy of each type per module) and 2) this isn't in line with the caller identity API which should also receive the same list.

massand (changes requested):
:clock1:

PR #45, opened 2020-10-13
Add aziot binary with `aziot init` command.

arsing:
`aziot init` takes in basic provisioning information from the user and
creates the config files for the three services. It's expected to be run
after a fresh install of the services on a new device.


arsing:
@CindyXing, @micahl: Please try it out. (Packages are in the GH Actions results.)

aziot/src/init.rs
```diff
@@ -0,0 +1,867 @@
...
+	}
+}
+
+fn prompt(stdin: &mut impl std::io::BufRead, question: &str) -> Result<String, crate::Error> {
+	println!("{}", question);
+
+	let mut line = String::new();
+
+	loop {
+		print!("> ");
```
daprilik, line 688: A small QOL improvement would be to drop in something like https://github.com/kkawakam/rustyline
People aren't perfect, and seeing `^[[D` when you've fat-fingered a filename is never fun

aziot/src/init.rs
```diff
@@ -0,0 +1,867 @@
...
+
+fn create_dir_all(
+	path: &(impl AsRef<std::path::Path> + ?Sized),
+	user: &nix::unistd::User,
+	mode: u32,
+) -> Result<(), crate::Error> {
+	let path = path.as_ref();
+	let path_displayable = path.display();
+
+	let () =
```
daprilik, line 791: any reason for these `let () =` bindings?

aziot/src/init.rs
```diff
@@ -0,0 +1,867 @@
...
+fn write_file(
+	path: &(impl AsRef<std::path::Path> + ?Sized),
+	content: &[u8],
+	user: &nix::unistd::User,
+	mode: u32,
+) -> Result<(), crate::Error> {
+	let path = path.as_ref();
+	let path_displayable = path.display();
+
+	let () =
```
daprilik, line 812: ditto

aziot/test-files/init/dps-symmetric-key/device-id
```diff
@@ -0,0 +1 @@
+aziot-identity-service|aziot-identity-service|aziot-identity-ser
```
daprilik, line 1: nit: missing newline

daprilik: looks like this applies to all `device-id`'s across the various tests

cert/aziot-certd/src/config.rs
```diff
@@ -157,29 +157,64 @@ impl<'de> serde::Deserialize<'de> for Est {
 	}
 }
 
+impl serde::Serialize for Est {
+	fn serialize<S>(&self, serializer: S) -> Result<S::Ok, S::Error> where S: serde::ser::Serializer {
+		let mut inner = EstInner {
```
daprilik, line 162: I haven't looked too deeply into where this struct is used, but is there any reason you've got this manually-flattened inner struct vs. using something like `#[serde(flatten)]`?

cert/aziot-certd/src/config.rs
```diff
@@ -194,45 +229,23 @@ fn deserialize_expiry_days<'de, D>(deserializer: D) -> Result<Option<u32>, D::Er
 }
 
 /// The method used to issue a certificate.
-#[derive(Clone, Copy, Debug, PartialEq)]
-pub(crate) enum CertIssuanceMethod {
+#[derive(Clone, Copy, Debug, PartialEq, serde::Deserialize, serde::Serialize)]
+pub enum CertIssuanceMethod {
```
daprilik, line 233: `#[serde(rename_all = "snake_case")]`?

daprilik (commented):
Haven't had the chance to test it locally, but I left a couple small comments down below

aziot/src/init.rs
arsing: It wouldn't work, for the same reason as https://github.com/Azure/iot-identity-service/blob/1499fb72dbc2ffc89ef5a78e474db50f17672a8c/aziot/src/init.rs#L451-L453

aziot/src/init.rs
arsing: https://github.com/Azure/iot-identity-service/blob/1499fb72dbc2ffc89ef5a78e474db50f17672a8c/CONTRIBUTING.md See the note for `clippy::let_unit_value`

aziot/test-files/init/dps-symmetric-key/device-id
arsing: It's not supposed to have a newline. It's a binary file. The test data just happens to look like an ASCII string.

cert/aziot-certd/src/config.rs
arsing: I don't think `EstAuthX509` can be represented even with `flatten`ing.

cert/aziot-certd/src/config.rs
arsing: Ah yes. Done.

aziot/src/init.rs
daprilik: Oof. I guess you did that so we could mock input in tests?
That's a shame, since it'd be really nice...

Theoretically, it could be possible to use `#[cfg(test)]` to swap out the implementations, but it might be annoying.

Easy solution: make `aziot` a wrapper around a shell script that runs `rlwrap aziot-real` :wink:

aziot/src/init.rs
daprilik: Huh, I can't say I've ever seen that pattern before 🤔

While I don't think it's something I'd do myself, I can see how it might make it easier for users unfamiliar with the specific function understand that the return value is unit.

aziot/test-files/init/dps-symmetric-key/device-id
daprilik: oh, I totally wouldn't have guessed that. okie all good then, carry on

aziot/src/init.rs
arsing: >Oof. I guess you did that so we could mock input in tests?

Yes.

arsing: Added a trait.

aziot/src/init.rs
```diff
@@ -0,0 +1,948 @@
...
+				Err(rustyline::error::ReadlineError::Utf8Error) => println!("Input could not be parsed as UTF-8"),
+
+				Err(rustyline::error::ReadlineError::Errno(err)) => return Err(std::io::Error::new(std::io::ErrorKind::Other, err)),
+
+				Err(err) => return Err(std::io::Error::new(std::io::ErrorKind::Other, err)),
+			}
+		}
+	}
+
+	fn read_secret(&mut self, prompt: &str, line: &mut String) -> std::io::Result<usize> {
```
daprilik, line 690: theoretically, you could remove the rpassword dependency and re-use rustyline: https://github.com/kkawakam/rustyline/pull/266/files

I haven't checked how much extra "security" rpassword offers, but if it's "good enough" to simply replace the password chars with '*', we might be able to cut down on a dependency

aziot/src/init.rs
```diff
@@ -0,0 +1,948 @@
...
+
+	let symmetric_key = symmetric_key.ok_or(r#"required parameter "SharedAccessKey" is missing"#)?;
+	let symmetric_key =
+		base64::decode(symmetric_key)
+		.map_err(|err| format!(r#"connection string's "SharedAccessKey" parameter could not be decoded from base64: {}"#, err))?;
+
+	Ok((iothub_hostname.to_owned(), device_id.to_owned(), symmetric_key))
+}
+
+fn convert_path_or_uri_to_uri(value: String) -> Option<String> {
```
daprilik, line 798: super minor, but there's technically no reason why a user can't use a `pkcs11:` prefixed string when they're not supposed to. I know there's no reason why someone _would_ do that, but maybe it makes sense to add a `allow_pkcs11: bool` to the function signature.

aziot/src/init.rs
```diff
@@ -0,0 +1,948 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+// This subcommand interactively asks the user to give out basic provisioning information for their device and
```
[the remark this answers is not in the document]
arsing: Oh yeah, it was supposed to be one. Fixed.

aziot/src/init.rs
```diff
@@ -0,0 +1,948 @@
...
+		}
+	}};
+}
+
+/// Returns the KS/CS/IS configs, and optionally the contents of a new /var/secrets/aziot/keyd/device-id file to hold the device ID symmetric key.
+fn run_inner(stdin: &mut impl Reader) -> Result<(
+	Vec<u8>,
+	Vec<u8>,
+	Vec<u8>,
+	Option<Vec<u8>>,
```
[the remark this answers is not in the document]
arsing: Yeah, I was being lazy. Fixed.

aziot/src/init.rs
arsing: `rpassword` disables the echo attribute on the TTY in the standard way. `rustyline` instead puts the TTY into raw mode and uses the highlighter to emit the keystrokes. Both approaches are basically the same, so it should be fine.

massand (approved):
:shipit:

aziot/src/init.rs
arsing: Fixed. I made it into two functions that properly parse URLs of the target type (`aziot_keys::PreloadedKeyLocation` for keyd and `url::Url` for certd).

aziot/src/init.rs
```diff
@@ -1,15 +1,18 @@
...
-// creates the config files for the three services based on that information.
-//
-// Notes:
-//
-// - Provisioning with a symmetric key (manual or DPS) requires the key to be preloaded into KS, which means it needs to be
-//   saved to a file. This subcommand uses a file named `/var/secrets/aziot/keyd/device-id` for that purpose.
-//   It creates the directory structure and ACLs the directory and the file appropriately to the KS user.
-//
-// - `dynamic_reprovisioning` is enabled by default in IS provisioning settings.
+/// This subcommand interactively asks the user to give out basic provisioning information for their device and
```
daprilik, line 3: uhh, this is now a doc-comment on `const DPS_GLOBAL_ENDPOINT`.

I think you meant to use `//!`

daprilik (changes requested):

aziot/src/init.rs
arsing: Derp.

daprilik (approved):

aziot/src/init.rs
daprilik: nice commit message lol

PR #47, opened 2020-10-15
Add DPS-TPM Provisioning

daprilik:
Add DPS-TPM Provisioning support to `identityd`.

This PR contains a complete E2E implementation of DPS-TPM provisioning, including integration with the `aziot init` command.

To test the new functionality: create a TPM individual enrollment in DPS and configure `identityd` to use TPM provisioning:

```toml
[provisioning]
dynamic_reprovisioning = true
source = "dps"
global_endpoint = "https://global.azure-devices-provisioning.net"
scope_id = "0ne0018EEC2"
  [provisioning.attestation]
  method = "tpm"
  registration_id = "wd5pvdv6w272j5alm3di2y5ydebh3f5mjyu5zxgxktwsinri5waa"
```

See the newly added docs for more implementation details, including the new `tpmd` HTTP API.

### Implementation Roadmap

- [x] Draft initial design docs for the TPM service
- [x] Port over the `azuire-iot-hsm-c` library + strip all non-TPM related functionality
- [x] Create Rust bindings to C library via `aziot-tpm-sys` and `aziot-tpm-rs`
- [x] Expose `aziot-tpm-rs` API over HTTP via the `aziot-tpmd` daemon
- [x] Update `aziot-identityd` with support for DPS-TPM provisioning, using `aziot-tpmd`'s HTTP API
- [x] Plumb-through the TPM key source throughout identityd
- [x] Fix packaging on `centos7`
- [x] Update `aziot init`


tpm/aziot-tpm-rs/src/lib.rs
```diff
@@ -0,0 +1,32 @@
...
+#![allow(clippy::missing_errors_doc, clippy::module_name_repetitions)]
+#![deny(missing_docs)]
+
+mod error;
+mod tpm;
+
+pub use crate::error::Error;
+pub use crate::tpm::{Tpm, TpmDigest, TpmKey, TpmKeys};
+
+// Traits
```
arsing, line 16: You don't need to replicate this from libiothsm. Two traits is unnecessary when one will suffice, and one trait is unnecessary when there's only one impl. Just make them crate-level fns.

tpm/aziot-tpm-rs/src/tpm.rs
```diff
@@ -0,0 +1,177 @@
...
+    aziot_tpm_get_keys, aziot_tpm_import_auth_key, aziot_tpm_init, aziot_tpm_sign_with_auth_key,
+    AZIOT_TPM_HANDLE,
+};
+
+use crate::Error;
+use crate::{ManageTpmKeys, SignWithTpm};
+
+/// Hsm for TPM
+///
+/// Create an instance of this to use the TPM interface of an HSM.
```
arsing, line 20: There's only one "instance" of this implementation that needs to exist, so why does it need a handle? Why does it have to be `new()`'d ?

Eg this API makes it possible to call `new()` twice. What happens then?

tpm/aziot-tpm-sys/README-TPM.md
```diff
@@ -0,0 +1,118 @@
+
+# Setting up a TPM device with iotedged access
```
arsing, line 2: There is no `iotedged` in this repo.

tpm/aziot-tpm-sys/README.md
```diff
@@ -0,0 +1,60 @@
+# HSM-SYS crate
```
arsing, line 1: Wrong name. (Same for `HSM-RS` below.)

tpm/aziot-tpm-sys/azure-iot-hsm-c/inc/hsm_client_data.h
```diff
@@ -0,0 +1,135 @@
+// Copyright (c) Microsoft. All rights reserved.
```
arsing, line 1: "HSM" is a misnomer since this lib is only for TPMs. That applies to all the files being added.

For this header specifically, why does this header need to exist? The `aziot_tpm_*` functions just forward to it, so you could merge them?

tpm/aziot-tpm-sys/azure-iot-hsm-c/src/hsm_log.c
```diff
@@ -0,0 +1,96 @@
+#include <stdarg.h>
+#include <stdbool.h>
+#include <stdio.h>
+#include <time.h>
+
+#include "hsm_log.h"
+
+#if defined __WINDOWS__ || defined _WIN32 || defined _WIN64 || defined _Windows
```
arsing, line 8: There's no Windows support in this repo.

arsing (changes requested):

tpm/aziot-tpm-rs/src/lib.rs
daprilik: Yeah, I was debating about whether or not to leave the two traits as separate entities. One should do.

That said, I think it makes sense to use a Trait, as it'll make testing/mocking easier.

tpm/aziot-tpm-sys/README-TPM.md
daprilik: Ah, whoops, this README (and the other one as well) were copied wholesale from the old repo.

That said, I'm not sure how much value these READMEs add over the existing in-repo docs regarding setting up a TPM, and I was thinking of just deleting them entirely. Thoughts?

tpm/aziot-tpm-sys/azure-iot-hsm-c/inc/hsm_client_data.h
daprilik: Whoops, I thought I left a DEVNOTE somewhere that explains why I did that. Oh, yep, it's part of the 
63ec387 commit message. Copied here for reference:

> at the moment, this is simply a facade over the existing hsm_sys functions, in an effort to keep tests passing.

You're absolutely right - the two headers could be merged into one. It's moreso that I'd rather not spend time fixing the [one remaining] C test in-tree. We can re-visit whether or not this API should be consolidated as the PR gets closer to being merged.

tpm/aziot-tpm-sys/azure-iot-hsm-c/src/hsm_log.c
daprilik: Good catch. I thought I got everything windows related, but looks like I missed a spot. Will fix.

tpm/aziot-tpm-rs/src/tpm.rs
daprilik: I mean, what would be the alternative? Having a single static global TPM handle instance in the C code which is "implicitly" passed to each function? I can't say I'm a huge fan of that...

Plus, while I'm all for deleting unused C code, I can't say I'm too excited to re-write the existing C code. The Rust-side bookkeeping required to store + eventually free a TPM handle is super straightforward, and doesn't make things all that more complicated.

As for what happens when `new()` is called twice? Well, based on the fact that this was also possible with the old API, and that the documentation never explicitly states this is a one-time global `new` function, I'm _assuming_ that nothing too bad would happen. i.e: it would simply return a fresh handle, and things would Just Work. That's a pretty naive view though, and I could be totally wrong.

If you've got any suggestions, I'm all ears, but as mentioned above, I'm of the opinion that if the code works, it'd be best to leave it as-is and focus on the rest of the task at hand.

tpm/aziot-tpm-sys/README-TPM.md
arsing: This file `README-TPM.md` has useful info, but also is redundant with a similar page in docs.microsoft.com, so I'd be down for removing it.

The other one has useful info that is not duplicated anywhere, so I would keep that one.

tpm/aziot-tpm-sys/azure-iot-hsm-c/tests/test_utils/test_utils.c
```diff
@@ -0,0 +1,198 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+#include <stdlib.h>
+
+#include "testrunnerswitcher.h"
+#include "azure_c_shared_utility/crt_abstractions.h"
+#include "azure_c_shared_utility/strings.h"
+#include "azure_c_shared_utility/uniqueid.h"
+
+#if (defined __WINDOWS__ || defined _WIN32 || defined _WIN64 || defined _Windows)
```
arsing, line 11: Remove the Windows-only code in this file too.

tpm/aziot-tpm-rs/src/lib.rs
arsing: As we'd discussed, if you mock the crate API, then there's no point writing such a test, because the API that has been mocked was the only thing worth testing.

We can test with `ibmswtpm2` and run the real code against it.

daprilik: My one concern with that approach is that if we parallelize the tests, we might run into problems where one test will register a key, and before it has a chance to sign something, another test registers another key, leading to weird shenanigans.

Then again, I've never actually written a proper end-to-end test in our repo, so I'm not sure how in-depth we'd go.

I'm cool with nixing the trait, just lmk

tpm/aziot-tpm-rs/src/tpm.rs
arsing: >As for what happens when `new()` is called twice? Well, based on the fact that this was also possible with the old API, and that the documentation never explicitly states this is a one-time global new function, I'm assuming that nothing too bad would happen. i.e: it would simply return a fresh handle, and things would Just Work. That's a pretty naive view though, and I could be totally wrong.

It's not quite as straightforward as reading the documentation. `initialize_tpm_device` internally mutates two statics to store the parameters for creating a new TPM connection. They're always updated in exactly the same way, and once the statics have been used to create a connection they can be mutated freely without causing problems to the already opened connection. So all it needs from you is to ensure that `aziot_tpm_init` cannot be called from two threads simultaneously. Of course you don't *need* to do that, because tpmd will only call `aziot_tpm_init` once, which is exactly my point.

You can keep `Tpm::new()` and have it return a handle, but ensure from the Rust side that calls to `aziot_tpm_init` are serialized.

To be safe it might be worth ensuring every `aziot_tpm_` call is serialized in general. The KS does this with `libaziot-keys` by having the bindings take `&mut self`, which eventually means the HTTP layer has a mutex to ensure only one request runs at a time, which is good enough. You can do that for the other fns of `Tpm` that currently take `&self`, but you can't do that for `Tpm::new()` itself, so you might need a `lazy_static` `Mutex` there.

>I'd like to avoid re-writing the existing C code as much as possible.

What *I'd* like to avoid is having all this C code of external provenance in the first place, but if we're going to have it I want to ensure it has the same thought put into it as if we were writing it ourselves.

tpm/aziot-tpm-rs/src/lib.rs
arsing: Any tests involving `ibmswtpm2` will have to be serialized, as we'd already discussed.

For E2E tests they wouldn't be run in parallel anyway, because they all operate on global resources. The first step of an E2E test is installing the deb / rpm package and setting up the config for the test.

tpm/aziot-tpm-rs/src/tpm.rs
daprilik: I'm going to assume that when you say `aziot_tpm_init` you actually mean `aziot_tpm_create`. `_init` doesn't do much aside from setup library-side logging. The `_create` method is the one that actually returns a new handle. 

> To be safe it might be worth ensuring every aziot_tpm_ call is serialized in general. The KS does this with libaziot-keys by having the bindings take &mut self, which eventually means the HTTP layer has a mutex to ensure only one request runs at a time, which is good enough. You can do that for the other fns of Tpm that currently take &self, but you can't do that for Tpm::new() itself, so you might need a lazy_static Mutex there.

Yeah, I thought that might be an issue, so I followed the same pattern you used in `aziot-keyd`. I'll be pushing those changes up soon-ish.

I'll also go ahead and add some sort of `lazy_static` / `std::sync::Once` based code to `aziot-tpm-rs` that ensures the methods you pointed out are only called once / serialized.

> What I'd like to avoid is having all this C code of external provenance in the first place, but if we're going to have it I want to ensure it has the same thought put into it as if we were writing it ourselves.

I feel that on a deep and visceral level, but based on my understanding of what you've said, the only benefit in changing the C API would be to remove the Rust-side handle book-keeping. The fundamental issue of ensuring certain methods are only called once / called by a single thread at a time won't be solved, right?

tpm/aziot-tpm-rs/src/lib.rs
daprilik: Alrighty then, I shall make the traits b e g o n e

tpm/aziot-tpm-rs/src/tpm.rs
arsing: >I'm going to assume that when you say `aziot_tpm_init` you actually mean `aziot_tpm_create`.

Right, sorry. I had written a paragraph about merging `_init` and `_create` into one to further reinforce that there is only one TPM connection, but lost it in editing.

>the only benefit in changing the C API would be to remove the Rust-side handle book-keeping. The fundamental issue of ensuring certain methods are only called once / called by a single thread at a time won't be solved, right?

For `_create` it would be solved by not having the Rust wrapper give the indication it could safely be called more than once, which it currently does by virtue of being called `new()` and returning a handle.

For the others, the lack of thread-safety is independent of whether there is a handle or not. It just requires something on the Rust side to ensure a `Mutex` is used somewhere. Having a handle does make it easier by making the handle methods take `&mut self` so that callers like the HTTP layer that have shared references are forced to use a `Mutex`.

daprilik: Okay, I feel like we're on the same page, and I'm just bad at formulating questions. Just to make sure: we can leave the C API as-is (i.e: using handles), and just tweak the Rust API to enforce the C API's soft invariants, yes?

arsing: Yes.

daprilik: Addressed in 67fc30d

tpm/aziot-tpm-client-async/src/lib.rs
```diff
@@ -0,0 +1,168 @@
...
+                self.api_version
+            ),
+            Some(&body),
+        )
+        .await?;
+        Ok(res.digest.0)
+    }
+}
+
+async fn request<TRequest, TResponse>(
```
arsing, line 82: Perhaps these should be moved to `http-common` since this is the third definition now. Doesn't have to be in this PR though.

tpm/aziot-tpm-common-http/src/lib.rs
```diff
@@ -0,0 +1,77 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+//! Request/Response types used by the TPM Service's HTTP API.
+
+#![deny(rust_2018_idioms)]
+#![warn(clippy::all, clippy::pedantic)]
+#![allow(clippy::wildcard_imports)] // to use `super::*`
```
arsing, line 7: But the `use super::*;` isn't doing anything in the first place?

tpm/aziot-tpm-rs/README.md
```diff
@@ -0,0 +1,4 @@
+# HSM Crate
```
arsing, line 1: Doc is wrong.

tpm/aziot-tpm-sys/README.md
```diff
@@ -0,0 +1,36 @@
...
+You may need additional setup for a TPM device see the in-tree dev docs for
+more details.
+
+## Memory allocation
+
+The current TPM API functions expect the calling function to allocate memory for
+the the caller to use. The caller (in this case, the Rust crate) is expected to
+free this memory.
+
+## Build dependencies
```
arsing, line 19: Just leave this section out. Apart from being incomplete ("On Debian based linux systems"), it's redundant with `install-build-deps.sh` and would have to be kept in sync.

tpm/aziot-tpm-sys/README.md
```diff
@@ -0,0 +1,36 @@
...
+This crate is dependent on CMake being installed. On Debian based linux systems,
+this can be installed with
+
+```
+sudo apt-get install build-essential cmake libcurl4-openssl-dev uuid-dev valgrind
+```
+
+### Valgrind
+
+Valgrind was added to the linux build dependencies. We are using Valgrind for
```
arsing, line 30:
```diff
-Valgrind was added to the linux build dependencies. We are using Valgrind for
+We use Valgrind for
```

tpm/aziot-tpm-sys/README.md
```diff
@@ -0,0 +1,36 @@
...
+detecting memory leaks, unassigned variables, and overruns in the dev mode
+iothsm library.
```
arsing, line 32: `s/iothsm/aziot-tpm/`

tpm/aziot-tpm-sys/README.md
```diff
@@ -0,0 +1,36 @@
...
+
+Valgrind slows down the tests considerably, so the iothsm library in hsm-sys
```
arsing, line 34: `s/iothsm library in hsm-sys/aziot-tpm library/`

tpm/aziot-tpm-sys/build.rs
```diff
@@ -0,0 +1,134 @@
...
+use std::env;
+use std::path::Path;
+use std::process::Command;
+
+use cmake::Config;
+
+const SSL_OPTION: &str = "use_openssl";
+const USE_EMULATOR: &str = "use_emulator";
+
+trait SetPlatformDefines {
```
arsing, line 15: Why make a trait with just one implementation instead of a free function?

tpm/aziot-tpmd/config/unix/default.toml
```diff
@@ -0,0 +1 @@
+#
```
arsing, line 1: Maybe

```toml
# This file intentionally left empty because aziot-tpmd has no configurable settings.
```

tpm/aziot-tpmd/aziot-keyd.socket
```diff
@@ -0,0 +1,13 @@
+[Unit]
+Description=Azure IoT TPM Service API socket
+PartOf=aziot-tpmd.service
+
+[Socket]
+ListenStream=/run/aziot/tpmd.sock
+SocketMode=0660
+DirectoryMode=0755
+SocketUser=aziotks
```
arsing, line 9: User and group should be `aziottpm`

tpm/aziot-tpmd/src/error.rs
```diff
@@ -0,0 +1,73 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+#[derive(Debug)]
+pub enum Error {
+    Internal(InternalError),
+    InvalidParameter(Option<(&'static str, Box<dyn std::error::Error + Send + Sync>)>),
+}
+
+impl Error {
+    // pub(crate) fn invalid_parameter<E>(name: &'static str, err: E) -> Self
```
arsing, line 10: Remove commented-out code, and presumably the `InvalidParameter` variant entirely since you're not using it.

arsing (changes requested):

tpm/aziot-tpm-sys/build.rs
daprilik: As with many things in this file, it's a remnant from when Windows support was a thing.

That said, having it as a trait _does_ make things a bit cleaner, as it plays nicely with the other `.define()` chains down below.

I see not real harm leaving it as an extension trait.

identity/aziot-dps-client-async/src/lib.rs
```diff
@@ -143,10 +222,21 @@ impl Client {
         let mut req = req.expect("cannot fail to create hyper request");
 
         let connector = match auth_kind {
-            DpsAuthKind::SymmetricKey { sas_key } => {
+            DpsAuthKind::Tpm { .. } => hyper_openssl::HttpsConnector::new()?,
+            DpsAuthKind::SymmetricKey { sas_key: ref key }
```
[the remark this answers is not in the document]
daprilik: turns out `&auth_kind` does the trick as well

key/aziot-key-client-async/src/lib.rs
```diff
@@ -127,10 +127,7 @@ impl Client {
             http::Method::GET,
             &format!(
                 "http://foo/key/{}?api-version={}",
-                percent_encoding::percent_encode(
-                    id.as_bytes(),
-                    http_common::PATH_SEGMENT_ENCODE_SET
-                ),
+                percent_encoding::percent_encode(id.as_ref(), http_common::PATH_SEGMENT_ENCODE_SET),
```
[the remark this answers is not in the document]
daprilik: I was playing around with switching the signature to `impl AsRef<[u8]>` and forgot tho change it back. Will fix.

tpm/aziot-tpm-client-async/src/lib.rs
daprilik: Yep, I would be in favor of extracting it out

tpm/aziot-tpm-common-http/src/lib.rs
daprilik: It imports `Serialize` and `Deserialize` into scope.

This is a stylistic choice, since I'm not a fan of using qualified paths for derive macros (especially those as foundational as `serde`), and would rather not `use serde::{Deserialize, Serialize}` in every sub-block.

tpm/aziot-tpm-rs/README.md
daprilik: I went ahead and deleted the file. It should be pretty obvious that this is the Rust half corresponding to `aziot-tpm-sys`.

tpm/aziot-tpmd/src/http/mod.rs
```diff
@@ -0,0 +1,45 @@
...
+
+fn to_http_error(err: &crate::Error) -> http_common::server::Error {
+    let error_message = http_common::server::error_to_message(err);
+
+    // TODO: When we get distributed tracing, associate these logs with the tracing ID.
+    for line in error_message.split('\n') {
+        log::log!(
+            match err {
+                crate::Error::Internal(_) => log::Level::Error,
+                // _ => log::Level::Info,
```
arsing, line 30: Remove commented-out line (but keep the `match` so that it's future-proof).

arsing (changes requested):

Makefile
```diff
@@ -189,7 +189,7 @@ test:
 		grep -v '^\./tpm/aziot-tpm-sys/azure-iot-hsm-c/' | \
 		grep -v '\.generated\.rs$$' | \
 		while read -r f; do \
-			if [ "$$(tail -c 1 "$$f" | wc -l)" -eq '0' ]; then \
+			if [[ -s "$$f" && "$$(tail -c 1 "$$f" | wc -l)" -eq '0' ]]; then \
```
arsing, line 192: Instead of this, just do `grep -v '^\.aziot/test-files/'`

arsing (changes requested):

Makefile
daprilik: I feel like my solution is a bit more robust, since if we ever add an empty-file elsewhere in the repo, things will continue to work okay. That said, I'm not dead-set on leaving it in, so I can change it if you like.

arsing: If we add an empty file, it should have a copyright header unless we decide otherwise.

daprilik: This doesn't have anything to do with copyright. This is only for the newline check

arsing: Derp. Sorry. In that case keep your original change.

contrib/centos/aziot-identity-service.spec
```diff
@@ -185,6 +182,8 @@ fi
 %doc THIRD-PARTY-NOTICES
 %license LICENSE
 
+%exclude /root/rpmbuild/BUILD/aziot-identity-service-%{version}/target/x86_64-unknown-linux-gnu/release/build/
```
arsing, line 185: This assumes the buildroot is `/root/rpmbuild`, which it doesn't have to be if someone else is using the spec file to build the package.

Also you can use `%{name}-%{version}` instead of `aziot-identity-service-%{version}`

arsing (changes requested):

contrib/centos/aziot-identity-service.spec
daprilik: I've played around with a whole bunch of prefixes (%{buildroot}, %{builddir}, both), but none of them seemed to work. I can keep iterating, if you've got another suggestion.

arsing: `%exclude %{_builddir}/%{name}-%{version}/target/x86_64-unknown-linux-gnu/release/build/`

works for me. Just tried it locally.

daprilik: The CI hath disagreed

arsing: Yes, as I told you on Teams already :)

identity/aziot-identityd/src/lib.rs
```diff
@@ -390,49 +399,6 @@ impl Api {
         Ok(())
     }
 
-    async fn dps_get_registered_device(
```
massand, line 393: This is a loop that calls get_operation_status in a loop after the register call has been made. This is how DPS provisioning works and we seem to be removing it. Is this intentional?

identity/aziot-identityd/src/lib.rs
```diff
@@ -475,12 +441,32 @@ impl Api {
                 scope_id,
                 attestation,
             } => {
+                async fn operation_to_iot_hub_device(
+                    credentials: aziot_identity_common::Credentials,
+                    operation: aziot_dps_client_async::model::RegistrationOperationStatus,
+                ) -> Result<aziot_identity_common::IoTHubDevice, Error> {
+                    let status = operation.status;
+                    assert!(!status.eq_ignore_ascii_case("assigning"));
```
massand, line 449: See other comment in the removed `dps_get_registered_device`. This isn't a runtime invariant, it is one of 2-3 possible value that can be returned by DPS

identity/aziot-cloud-client-async-common/src/lib.rs
```diff
@@ -30,18 +31,15 @@ pub async fn get_sas_connector(
...
-            )
+            .sign_with_key(&key_handle, sig_data.as_bytes())
             .await?;
 
         let signature = base64::encode(&signature);
 
         let token = url::form_urlencoded::Serializer::new(format!("sr={}", resource_uri))
             .append_pair("sig", &signature)
             .append_pair("se", &expiry)
+            .append_pair("skn", "registration")
```
massand, line 42: This isn't required for non-TPM DPS registration or IoT Hub connection. This should only be set when provisioning type is Dps and attestation type is Tpm, right?

massand (changes requested):
The DPS flow has changed in this PR. We need to keep the flows the same for non-TPM registration

identity/aziot-identityd/src/lib.rs
daprilik: I've moved the looping logic into the DPS client itself

identity/aziot-identityd/src/lib.rs
daprilik: Yes, it's not obvious why this would uphold, but it's valid because the underlying `register` call in the DPS client now checks for this case: https://github.com/Azure/iot-identity-service/pull/47/files#diff-50183279457368e456cbe124fbd69c2c4d1d2ebbeb988737299d06b93e78bd4dR151

identity/aziot-cloud-client-async-common/src/lib.rs
daprilik: The old repo unconditionally set it (see https://github.com/Azure/iotedge/blob/7831fc5ad76fa3bd33b1dc1e4630ac4db771e9ff/edgelet/dps/src/registration.rs#L84), so it seems fine leaving it in?

massand: Sure, but it applies to all DPS requests. Currently, you have it for all tokens (including hub).

identity/aziot-identityd/src/lib.rs
massand: I missed that in my first pass at the giant review :) LGTM

identity/aziot-identityd/src/lib.rs
massand: Makes sense in the context of the new code.

identity/aziot-dps-client-async/src/lib.rs
```diff
@@ -70,68 +87,130 @@ impl Client {
...
+                let aziot_tpm_common::TpmKeys {
+                    endorsement_key,
+                    storage_root_key,
+                } = self.tpm_client.get_tpm_keys().await?;
+
+                model::DeviceRegistration {
+                    registration_id: Some(registration_id.into()),
+                    tpm: Some(model::TpmAttestation {
+                        endorsement_key: base64::encode(&endorsement_key),
+                        storage_root_key: Some(base64::encode(&storage_root_key)),
```
massand, line 101: Any reason this is an Option? It is always specified in TpmAttestation

daprilik: Ah, I had copied the definition from the old repo, which probably auto-generated these structures. But yes, we always pass a string here, so I'll go ahead and change the defn

identity/aziot-cloud-client-async-common/src/lib.rs
daprilik: Yeah, you're right. I'm trying to think of the best way to fix this... The quick-and-dirty approach would be to add some sort of `is_tpm_registration: bool` to the method, but that's pretty ugly. Any suggestions?

massand (approved):
IS changes LGTM :shipit:

identity/aziot-cloud-client-async-common/src/lib.rs
massand: Yes - that is what I was thinking of too. Please change it because it might cause inadvertent failures in IoT Hub client logic if `skn` becomes a valid parameter.

daprilik: I added the boolean, and checked that provisioning still works :+1:

Arnavion (approved):

arsing (approved):

PR #43, opened 2020-10-12
Doc updates

micahl:
Linked the main README to the docs/README, fixed a few minor broken links, and added some additional intro with pointers to relevant conceptual docs.

PR #48, opened 2020-10-19
Fix broken links in docs and miscellaneous cleanup.

arsing:
- Jekyll doesn't like `_index.md`, so rename them to `index.md`. Also convert
  the inter-doc links to `.html`.

- Jekyll uses a top-level h1 heading for the page title, so add one to the pages
  that didn't have it.

- Add API version to all KS/CS/IS API docs.

- Fix KS API docs to match the current implementation.

- Rename occurrrences of `libiothsm-keygen` to `libaziot-keys` in docs
  and plantuml diagrams.

Co-authored-by: Arnav Singh <arsing@microsoft.com>


arsing:
This is a subset of https://github.com/Azure/iot-identity-service/pull/43 (it doesn't include the material about "edge devices" that is blocking that PR), plus additional fixes.

cc @micahl

massand (approved):

docs/README.md
```diff
@@ -2,20 +2,21 @@
...
 
-    The Identity Service provides access to host-level processes and container modules to connect to the cloud using tokens or X.509 certificate credentials, corresponding to their respective identities.
+    The Identity Service also provides access to native processes ("host processes") to connect to the cloud. Depending on how the user configures the Identity Service, host processes can get their own *module identities* provisioned by the Identity Service, or use the device identity, or both.
 
-    Note that this service performs the entirety of provisioning.
+    The Identity Service also provisions module identities for containerized modules, ie IoT Edge modules. However container modules do not communicate with the Identity Service directly; instead the provisioning of their identities is handled by the IoT Edge runtime on the modules' behalf. Refer to <https://github.com/Azure/iotedge> for more details about the IoT Edge runtime.
```
CindyXing, line 11: Given we purposely separating this repo out for both edge and non-edge devices, it would be helpful to clarify what is generic.

docs/keys-service.md
```diff
@@ -46,11 +48,11 @@
 
 ---
 
-## Get Existing Symmetric Key
+### Get Existing Symmetric Key
 
-`GET /key/{keyId}`
+`GET /key/{keyId}?api-version=2020-09-01`
```
CindyXing, line 53: 2020-09-01 is our current version. How about we specify the format here only?

CindyXing (approved):

docs/keys-service.md
arsing: The point of the API version is to have URLs that are not necessarily the same across all API versions. So the URL is not generic on the API version.

Eg if 2020-11-01 removes this URL, then it won't be correct to use `{apiVersion}` for this one.

The downside is that if this URL is *the same* for 2020-11-01, we'd have to duplicate the URL, but that's something we'd have to do anyway because there'd be separate docs for each API version.

docs/README.md
arsing: That's what I've already done. The previous paragraph talks about host processes. This one talks about IoT Edge modules. What do you want changed?

CindyXing: Yes, I saw that. Thanks. 

How about this: 
"IoT Identity Service is made up of three services that provide provisioning and cryptographic services for both edge and non-edge devices:"

"Specially for edge devices, the identity service also provision identities for containerized modules. ... "

docs/keys-service.md
CindyXing: If we are to have separate docs for each API version, we are good.

docs/README.md
arsing: Okay, I've done that, and also moved it to its own section to be clearer how the MR talks to all three.

docs/keys-service.md
arsing: Yes. I expect we'll remove these docs once we have the swagger specs and generate them from the swagger specs, and the swagger specs have to be one spec per API version like we have in the iotedge repo.

PR #49, opened 2020-10-19
[http-common] use DynRangeBounds for ApiVersion

daprilik:
This PR removes the need to have a dummy "MAX" API Version, and allows us to use the standard Rust range syntax (albeit through a `&'static dyn` trait object)

### DIFF
--- a/cert/aziot-cert-common-http/src/lib.rs
+++ b/cert/aziot-cert-common-http/src/lib.rs
@@ -6,14 +6,12 @@
 #[derive(Clone, Copy, Debug, Eq, Ord, PartialEq, PartialOrd)]
 pub enum ApiVersion {
 	V2020_09_01,
-	Max,
 }
 
 impl std::fmt::Display for ApiVersion {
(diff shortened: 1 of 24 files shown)


arsing:
I'd rather not make a new `Box` for every request.

Is a trait object const-promotable? Then you can return `&'static dyn` instead of `Box<dyn>`

daprilik:
Yep, that's very reasonable. I switched the signature to `&'static dyn`, and things Just Worked :tm:

arsing:
`identity/aziot-identity-common-http/src/lib.rs`'s `ApiVersion::MAX` also needs to be removed.

arsing (approved):
Thanks.

The bot takes the commit message from the PR OP, so please fix that to say `&'static` instead of `Box` before you add the merge label.

daprilik:
Done. 

I've never actually used the Bot, how does it work? 
Do you just add the "ready-to-merge" label and it merges for you?

arsing:
Yes.

daprilik:
neato

PR #50, opened 2020-10-19
[http-common] Use `async_trait` for http_common::Route

daprilik:
[`async_trait`](https://docs.rs/async-trait/0.1.41/async_trait/) is basically just some syntactic sugar over explicitly returning Pinned Boxed Futures syntax.

It's not a particularly heavy dependency, and really helps cut-down on the boilerplate required to define a route.

Don't be alarmed by the line number diff, it's mostly just whitespace removal :sweat_smile:

### DIFF
--- a/http-common/src/lib.rs
+++ b/http-common/src/lib.rs
@@ -10,6 +10,7 @@
 	clippy::must_use_candidate,
 	clippy::similar_names,
 	clippy::type_complexity,
+	clippy::too_many_lines,
 )]
 
 mod dynrange;
--- a/http-common/src/lib.rs
+++ b/http-common/src/lib.rs
@@ -9,8 +9,8 @@
 	clippy::module_name_repetitions,
 	clippy::must_use_candidate,
 	clippy::similar_names,
-	clippy::type_complexity,
 	clippy::too_many_lines,
+	clippy::type_complexity,
 )]
 
 mod dynrange;
--- a/http-common/src/server.rs
+++ b/http-common/src/server.rs
@@ -319,7 +319,11 @@ pub fn json_response(status_code: hyper::StatusCode, body: Option<&impl serde::S
 	res
 }
 
-mod fake_server {
+/// This server is never actually used, but is useful to ensure that the macro
+/// works as expected.
+mod test_server {
+	use crate as http_common;
+
 	#[derive(Clone, Copy, Debug, Eq, Ord, PartialEq, PartialOrd)]
 	pub enum ApiVersion {
 		FAKE,
@@ -344,24 +348,23 @@ mod fake_server {
 		}
 	}
 
-	use crate as http_common;
 
 	http_common::make_server! {
 		server: Server,
 		api_version: ApiVersion,
 		routes: [
-			fake_route::Route,
+			test_route::Route,
 		],
 	}
 
-	pub(crate) struct Server;
+	pub struct Server;
 
-	mod fake_route {
+	mod test_route {
 		use crate as http_common;
 
 		use super::ApiVersion;
 
-		pub(super) struct Route {}
+		pub(super) struct Route;
 
 		#[async_trait::async_trait]
 		impl http_common::server::Route for Route {
@@ -376,7 +379,7 @@ mod fake_server {
 				_path: &str,
 				_query: &[(std::borrow::Cow<'_, str>, std::borrow::Cow<'_, str>)],
 			) -> Option<Self> {
-				Some(Route {})
+				Some(Route)
 			}
 
 			type DeleteBody = serde::de::IgnoredAny;
@@ -390,7 +393,5 @@ mod fake_server {
 			type PutBody = serde::de::IgnoredAny;
 			type PutResponse = ();
 		}
-
 	}
-
 }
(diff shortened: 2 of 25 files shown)


arsing:
>Don't be alarmed by the line number diff, it's mostly just whitespace removal

https://github.com/Azure/iot-identity-service/pull/50/files?w=1

http-common/src/server.rs
```diff
@@ -316,3 +318,79 @@ pub fn json_response(status_code: hyper::StatusCode, body: Option<&impl serde::S
...
+	}
+
+	pub(crate) struct Server;
+
+	mod fake_route {
+		use crate as http_common;
+
+		use super::ApiVersion;
+
+		pub(super) struct Route {}
```
arsing, line 364: Just make it a unit struct.

http-common/src/server.rs
```diff
@@ -316,3 +318,79 @@ pub fn json_response(status_code: hyper::StatusCode, body: Option<&impl serde::S
...
+
+			type GetResponse = ();
+
+			type PostBody = serde::de::IgnoredAny;
+			type PostResponse = ();
+
+			type PutBody = serde::de::IgnoredAny;
+			type PutResponse = ();
+		}
+
```
arsing, line 393: Remove this blank line and 395

http-common/src/server.rs
```diff
@@ -316,3 +318,79 @@ pub fn json_response(status_code: hyper::StatusCode, body: Option<&impl serde::S
...
+		fn from_str(s: &str) -> Result<Self, Self::Err> {
+			match s {
+				"fake" => Ok(ApiVersion::FAKE),
+				_ => Err(()),
+			}
+		}
+	}
+
+	use crate as http_common;
```
arsing, line 347: `use` goes at the top of the module.

http-common/src/server.rs
```diff
@@ -316,3 +318,79 @@ pub fn json_response(status_code: hyper::StatusCode, body: Option<&impl serde::S
 	let res = res.expect("cannot fail to build hyper response");
 	res
 }
+
+mod fake_server {
```
arsing, line 322: `s/fake/test` would be nicer. Same for the route mod.

http-common/src/server.rs
```diff
@@ -316,3 +318,79 @@ pub fn json_response(status_code: hyper::StatusCode, body: Option<&impl serde::S
...
+
+	http_common::make_server! {
+		server: Server,
+		api_version: ApiVersion,
+		routes: [
+			fake_route::Route,
+		],
+	}
+
+	pub(crate) struct Server;
```
arsing, line 357: Does it need to be `pub(crate)` ?

arsing (changes requested):

http-common/src/server.rs
```diff
@@ -20,6 +20,7 @@ macro_rules! make_server {
 				std::task::Poll::Ready(Ok(()))
 			}
 
+			#[allow(clippy::too_many_lines)]
```
arsing, line 23: Move this to the crate-level allows.

identity/aziot-identityd/src/http/get_update_or_delete_module_identity.rs
```diff
@@ -36,75 +37,70 @@ impl http_common::server::Route for Route {
...
-			let identity = match inner.update_identity(auth_id, "aziot", &self.module_id).await {
-				Ok(v) => v,
-				Err(err) => return Err(super::to_http_error(&err)),
-			};
-			let res = aziot_identity_common_http::update_module_identity::Response {
-				identity,
-			};
-			Ok((hyper::StatusCode::OK, res))
-		})
+	#[allow(clippy::needless_pass_by_value)]
```
arsing, line 86: As discussed, add a comment here about why this is being done. Something like this is sufficient:

```rust
// clippy fires this lint for the `_body` parameter of the inner fn in the `async-trait` expansion.
// It's not clear why clippy does this, especially since it doesn't raise it for other functions
// that also ignore their `_body` parameter `fn delete` above.
//
// So suppress it manually.
```

http-common/src/server.rs
```diff
@@ -344,24 +348,23 @@ mod fake_server {
...
 		server: Server,
 		api_version: ApiVersion,
 		routes: [
-			fake_route::Route,
+			test_route::Route,
 		],
 	}
 
-	pub(crate) struct Server;
+	pub struct Server;
```
arsing, line 360: Err, what I meant is, can it be just `struct Server;`, ie neither `pub(crate)` nor `pub`?

Same for `ApiVersion`

arsing (changes requested):

http-common/src/server.rs
daprilik: oopsie, yep, true.

http-common/src/lib.rs
```diff
@@ -10,6 +10,7 @@
 	clippy::must_use_candidate,
 	clippy::similar_names,
 	clippy::type_complexity,
+	clippy::too_many_lines,
```
arsing, line 13: Keep them sorted please.

arsing (approved):

PR #51, opened 2020-10-20
Use ephemeral (in-memory) keys for local identity

gordonwang0:
To support container modules (which won't be able to interact with KS), local identity keys must be available as files. They can then be copied into the container.

This change stores local identity keys under IS's homedir/local_id_keys directory.


identity/aziot-identityd/src/lib.rs
```diff
@@ -28,11 +28,25 @@ const DPS_ASSIGNMENT_RETRY_INTERVAL_SECS: u64 = 10;
 /// This is the number of seconds to wait for DPS to complete assignment to a hub
 const DPS_ASSIGNMENT_TIMEOUT_SECS: u64 = 120;
 
+/// Private key types when creating a new identity certificate.
+#[derive(Clone)]
+enum KeyType<'a> {
+    /// Key handle managed by keyd.
+    Handle(&'a String),
```
daprilik, line 35: There's rarely a good reason to take a const-reference to a `String` directly instead of using a `&str`.

With `&String`, you end up doing a double pointer indirection whenever you access the data: once to dereference the `&String` into a `String`, and then one more time to access the underlying byte array.

* * *

That said, in this case, I think you'd be better off switching to `Handle(String)` directly, and not messing with any lifetimes.

gordonwang0: Fixed.

PR #53, opened 2020-10-21
Add aziot-keyd and aziot-certd OpenAPI specs.

arsing:
Other changes:

- Fix naming of `preferredAlgorithms` field of keyd's
  create-key-pair-if-not-exists API to use camel case instead of snake case.

- Merge certd's import-cert API into the get-or-delete-cert API so that
  the import-cert API can actually be invoked. Previously it would never get
  invoked because the get-or-delete-cert route would match the same URLs
  and end up returning HTTP 405.


cert/aziot-certd/src/http/get_or_import_or_delete.rs
```diff
@@ -73,6 +73,21 @@ impl http_common::server::Route for Route {
...
+    ) -> http_common::server::RouteResponse<Self::PutResponse> {
+        let mut inner = self.inner.lock().await;
+        let inner = &mut *inner;
+
+        match inner.import_cert(&self.cert_id, &body.pem.0) {
+            Ok(()) => (),
+            Err(err) => return Err(super::to_http_error(&err)),
+        };
+
+        let res = aziot_cert_common_http::import_cert::Response { pem: body.pem };
```
daprilik, line 90: Is there any particular reason we're parroting back the pem?

cert/aziot-certd/openapi/2020-09-01.yaml
```diff
@@ -0,0 +1,128 @@
...
+info:
+  title: 'aziot-certd API specification'
+
+  version: '2020-09-01'
+
+  description: |
+    This is the specification of the HTTP API of the aziot-certd service.
+
+  license:
+    name: 'Apache 2.0'
```
gordonwang0, line 15: License should be MIT (same for keyd file below)

cert/aziot-certd/openapi/2020-09-01.yaml
```diff
@@ -0,0 +1,128 @@
...
+    url: 'https://www.apache.org/licenses/LICENSE-2.0.html'
+
+
+servers:
+- url: 'http://foo/'
```
gordonwang0, line 20: I believe it's valid to specify unix:// as the scheme, so `unix://run/aziot/certd.sock`.

arsing: No, because the spec says about Path objects that:

>The path is appended to the URL from the Server Object in order to construct the full URL.

cert/aziot-certd/openapi/2020-09-01.yaml
arsing: Fixed.

cert/aziot-certd/src/http/get_or_import_or_delete.rs
arsing: Just for consistency. It's quite common for "create an object with this spec" API to return the spec to you.

cert/aziot-certd/openapi/2020-09-01.yaml
gordonwang0: Links (the line after license name) is still pointing to Apache license.

arsing: Ah shit, I thought I was pushing to my private branch.

arsing: Anyway, also fixed now.

gordonwang0 (approved):

cert/aziot-certd/openapi/2020-09-01.yaml
```diff
@@ -0,0 +1,127 @@
...
+      summary: 'Imports a PEM as a certificate with the given ID.'
+      requestBody:
+        content:
+          'application/json':
+            schema:
+              $ref: '#/components/schemas/ImportCertRequest'
+        required: true
+      responses:
+        '200':
+          description: 'HTTP 201 response'
```
massand, line 71: Nit: Description isn't matching status code.

massand (approved):

cert/aziot-certd/openapi/2020-09-01.yaml
arsing: Fixed. 201 is the right one.

PR #54, opened 2020-10-22
Fix docs some more.

arsing:
Fix docs some more.

- Move the overview diagram in README.md back to the overview section.

- Update the package contents listed in packaging.md to match
  the current implementation.

- Update packaging.md references to `aziot init` and `iotedge init` to match
  the current / expected implementation.

- Change dev/running/pkcs11/index.md to use lists instead of tables.
  The table is too wide to fit in the slate theme's narrow page.

- Make the code blocks use even shorter lines because the GitHub Pages versions
  of these pages is even narrower.

- Remove mentions of 1.0.10 and 1.0.11 as the versions of the iotedge and
  aziot-identity-service packages, since those are no longer
  the correct numbers.

- Revert inter-doc links from `.html` to `.md`. Jekyll does actually handle
  them fine and converts them to `.html` links, so this way the inter-doc links
  work for people reading the `.md` files on github.com too.


massand (approved):

massand:
The packaging.md has references to the old .sock file locations - `/var/run/aziot/` . Also the binaries are now in `/usr/libexecdir/aziot-identity-service`
Please fix those references

massand (changes requested):
nit and some old package references

arsing:
Ah, but that page has the same broken formatting from markdown in HTML tables, and I don't really want to change *that* page to be a list too...

arsing:
Gordon's PR #55 sets the markdown processor to GFM, so the packaging.md table problem will be solved by that. I've rebased this PR on top of that one. Will rebase it again once that PR is merged.

massand (approved):
My comment about old binary paths has been addressed. :shipit:

gordonwang0 (commented):
The code blocks on https://arsing.github.io/iot-identity-service/dev/running/ should also be split into shorter lines.

gordonwang0 (changes requested):
See comment above

arsing:
Done.

gordonwang0 (approved):

PR #57, opened 2020-10-25
Merge service binaries into a single binary.

arsing:
aziot-{cert,identity,key}d are now symlinks to a new aziotd binary.
This new binary looks at its `argv[0]` to figure out which service it's being
invoked as, and runs the corresponding service's code.

One advantage of this approach is that the common code for
reading config env vars, binding the service connector and
creating an HTTP server can be made common to all services.

The second advantage is that the three services can now share a single copy of
tokio, hyper, etc dependencies, that previously were statically linked into
each individual binary. This has savings for both disk space as well as
memory usage (from sharing code pages).

Making the startup sequence code common also means all services now
consistently use `env_logger` to set up logging. keyd and certd have been
updated to use the `log` crate instead of `{e,}println!`

Other changes:

- Add support for reading multiple config files. The base file is still
  config.toml. In addition any files found under `config.d` will be applied
  as patches to the base config.

  This will be needed in the near future when `iotedge init` needs to add
  additional preloaded keys and certs to keyd's and certd's config.
  The same or a similar mechanism will also be useful for third-party packages
  to register principals with identityd.

- Make naming of servers vs services vs APIs clearer.

  hyper owns the server. The server uses instances of a service to
  handle requests. The service implements the HTTP routes using an inner API.

- Fix some tabs that rustfmt did not convert.

- Don't run `cargo fmt -- --check` with `--quiet` regardless of the verbose flag,
  because it doesn't print errors if `--quiet` is specified.

- Separate the rlib components of aziot-keys into aziot-keys-common.
  This is to avoid building aziot-keys as an rlib, because it triggers
  a cargo bug that causes it to be rebuilt over and over again even if it
  isn't modified.

  There were three things causing it to be built as an rlib:

  - Tests. The test target now excludes aziot-keys.

  - aziot used the `PreloadedKeyLocation` type from the crate. That type is
    now in the aziot-keys-common crate.

  - aziotd had it as a dependency to force it to be built first, because it
    has extern C fns that link to it. This is now enforced by the makefile
    instead.


Makefile
```diff
@@ -168,7 +175,7 @@ test:
 	$(CARGO) clippy --all --exclude aziot-key-openssl-engine-shared --tests $(CARGO_PROFILE) --target $(CARGO_TARGET) $(CARGO_VERBOSE)
 	$(CARGO) clippy --all --examples $(CARGO_PROFILE) --target $(CARGO_TARGET) $(CARGO_VERBOSE)
 
-	$(CARGO) fmt --all $(CARGO_VERBOSE) -- --check
+	$(CARGO) fmt --all -- --check
```
daprilik, line 178: oopsie, my bad :sweat_smile:

docs/dev/running/aziot-certd.md
```diff
@@ -205,9 +205,13 @@ With this basic file, fill it out depending on what workflow you want to test:
...
 
     ```sh
+    export LD_LIBRARY_PATH="$LD_LIBRARY_PATH:$PWD/target/x86_64-unknown-linux-gnu/debug"
+
+    export AZIOT_LOG=aziot=debug
+
     export AZIOT_CERTD_CONFIG='...'
 
-    cargo run -p aziot-certd
+    (exec -a aziot-certd target/x86_64-unknown-linux-gnu/debug/aziotd)
```
daprilik, line 214: Does `cargo run -p aziotd -- aziot-certd` not work?

daprilik (commented):
Some general comments:

cert/aziot-certd/Cargo.toml
```diff
@@ -5,9 +5,9 @@ authors = ["Azure IoT Edge Devs"]
 edition = "2018"
 build = "build/main.rs"
 
+
```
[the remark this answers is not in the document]
arsing: I did them explicitly, because the `[dependencies]` section has one-line gaps between groups of deps, so it seems nicer (?) to have two-line gaps between the sections themselves.

cert/aziot-certd/src/http/mod.rs
```diff
@@ -3,36 +3,45 @@
...
 
-fn to_http_error(err: &aziot_certd::Error) -> http_common::server::Error {
+fn to_http_error(err: &crate::Error) -> http_common::server::Error {
     let error_message = http_common::server::error_to_message(err);
 
     // TODO: When we get distributed tracing, associate these logs with the tracing ID.
     for line in error_message.split('\n') {
-        eprintln!("!!! {}", line);
+        log::log!(
+            if matches!(err, crate::Error::Internal(_)) {
```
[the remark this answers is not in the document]
arsing: The intent is that `Internal` is the only one that maps to `Error` [1] and everything else maps to `Info`. It doesn't matter for certd and keyd because they only have two variants, but identityd has many more and so the if-else makes that clearer.

Or do you mean to merge it into the `match` below? If so, I left it as-is because it's better than repeating the loop in every variant's arm.

[1]: I just noticed I've written it backwards. I'll fix that.

aziotd/src/main.rs
```diff
@@ -0,0 +1,243 @@
...
+
+            return;
+        }
+    }
+
+    *base = patch;
+}
+
+#[derive(Debug)]
+struct Error(ErrorKind, backtrace::Backtrace);
```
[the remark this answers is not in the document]
arsing: Done.

aziotd/src/main.rs
```diff
@@ -0,0 +1,243 @@
...
+
+    Ok(())
+}
+
+fn merge_toml(base: &mut toml::Value, patch: toml::Value) {
+    // Similar to JSON patch, except that maps are called tables, and
+    // there is no equivalent of null that can be used to remove keys from an object.
+
+    if let toml::Value::Table(base) = base {
+        if let toml::Value::Table(patch) = patch {
```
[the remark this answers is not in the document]
arsing: It can't. Try it yourself.

cert/aziot-certd/src/http/mod.rs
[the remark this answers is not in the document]
daprilik: Oh, I meant it moreso as why are you using `matches!` if there's also an `else` branch

Right now, the code expands to:

```rust
    if match err {
        crate::Error::Internal(_) => true,
        _ => false,
    } {
        // ...
    } else {
        // ..
    }
```

Which just seems redundant

aziotd/src/main.rs
```diff
@@ -0,0 +1,243 @@
...
+            for (k, v) in patch {
+                merge_toml(base.entry(k).or_insert(toml::Value::Boolean(false)), v);
```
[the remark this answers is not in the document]
arsing: Done.

docs/dev/running/aziot-certd.md
arsing: `cargo run` passes those as `std::process::Command::arg`s, ie `argv[1..]`. Overriding `argv[0]` is an OS-specific thing in the first place, and `std::process::Command` only supports it for unix.

cert/aziot-certd/Cargo.toml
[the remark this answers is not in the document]
daprilik: Uhh, sure? In that case, you might want to skim through the rest of the TOML files and keep them consistent.

Either way though, not sure if it makes sense to have these changes in this PR, but s'aight.

inb4 tomlfmt when??

aziotd/src/main.rs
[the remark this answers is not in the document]
daprilik: oof, that sucks. I didn't realize that `base` was borrowed, so it's an implicit `ref mut` binding.

docs/dev/running/aziot-certd.md
daprilik: oh, I see, took me a second to remember that `argv[0]` is (traditionally) the executable's name. I now see why this bit of rigmarole is required.

Nonetheless, given that (afaik) we aren't passing any other parameters over the CLI, could we have the code _also_ check for `argv[1]` as well?

cert/aziot-certd/Cargo.toml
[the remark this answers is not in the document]
arsing: I've fixed the ones that are being edited in this PR. We can fix the rest when we edit them.

cert/aziot-certd/src/http/mod.rs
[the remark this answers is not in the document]
arsing: Ah. I was proactvely correcting for https://rust-lang.github.io/rust-clippy/master/index.html#match_like_matches_macro ("one meaningful arm + one wildcard arm => use `matches!`"), but it only fires for bool exprs.

Done.

docs/dev/running/aziot-certd.md
arsing: Yeah, I can add it as a fallback if `argv[0]` is `aziotd`

arsing: Done.

aziotd/src/main.rs
```diff
@@ -0,0 +1,243 @@
...
+        std::process::exit(1);
+    }
+}
+
+async fn main_inner() -> Result<(), Error> {
+    let argv0 = std::env::args_os()
+        .next()
+        .ok_or_else(|| ErrorKind::GetProcessName("argv[0] not set".into()))?;
+
+    // argv[0] could be a single component like "aziot-certd", or a path that ends with "aziot-certd",
```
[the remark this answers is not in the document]
arsing: I don't follow. The point is that a simple check for `match argv0 { Some("aziot-certd") => { ... }, ... }` is not sufficient since it might be a full path.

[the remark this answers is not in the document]
daprilik: yep, sorry, this comment stemmed from the fact that I forgot that we were checking `argv[0]` and not `argv[1]`.

aziot/src/init.rs
```diff
@@ -1065,6 +1065,7 @@ fn write_file(
     Ok(())
 }
 
+#[cfg(debug_assertions)]
```
daprilik, line 1068: Is there any reason why the tests wouldn't work in release mode?

aziotd/src/main.rs
```diff
@@ -186,58 +227,154 @@ where
...
-    // Similar to JSON patch, except that maps are called tables, and
-    // there is no equivalent of null that can be used to remove keys from an object.
+    // Similar to JSON patch, except that:
+    //
+    // - Maps are called tables.
+    // - There is no equivalent of null that can be used to remove keys from an object.
+    // - Arrays are merged via concatenating the patch to the base, rather than replacing the base with the patch.
 
-    if let toml::Value::Table(base) = base {
+    if let toml::Value::Table(base) = &mut *base {
```
daprilik, line 236: Is the `&mut *base` here to clarify that `base` has type `&mut`, or is there some other reason I'm missing?

If it's the former, then it probably best to just leave it as `base`, and keep things terse.

aziotd/src/main.rs
```diff
@@ -186,58 +227,154 @@ where
...
-            ErrorKind::ReadConfig(Some(path), _) => {
-                write!(f, "could not read config from {}", path.display())
-            }
-            ErrorKind::ReadConfig(None, _) => f.write_str("could not read config"),
-            ErrorKind::Service(_) => f.write_str("service encountered an error"),
+#[cfg(test)]
+mod tests {
+    #[test]
+    fn process_name_from_args() {
+        for &(input, expected) in &[
```
daprilik, line 262: super minor, but it would be nice to avoid repeating `#[cfg(debug_assertions)]` by using a Vec. i.e:

```rust
let mut test_cases = Vec::new();
test_cases.extend(&[ /* ... */ ]);
#[cfg(debug_assertions)]
test_cases.extend(&[ /* ... */ ])
```

docs/dev/running/aziot-certd.md
```diff
@@ -211,7 +211,7 @@ With this basic file, fill it out depending on what workflow you want to test:
 
     export AZIOT_CERTD_CONFIG='...'
 
-    (exec -a aziot-certd target/x86_64-unknown-linux-gnu/debug/aziotd)
+    cargo run --target x86_64-unknown-linux-gnu -p aziotd -- aziot-certd
```
daprilik, line 214: Is there any reason you're explicitly stating the `--target` here? 
Is it just to make make extra sure that the `LD_LIBRARY_PATH` is correct?

aziot/src/init.rs
arsing: https://github.com/Azure/iot-identity-service/pull/57/commits/e36e1334cd8c4a37f4506e5dcb3c1845e3c3dd24

>restrict argv[1] fallback to debug builds.

aziotd/src/main.rs
arsing: `&mut` is not `Copy` so it moves by default unless it's reborrowed.

docs/dev/running/aziot-certd.md
arsing: The docs tell you to build using `make` which always sets `--target` (for ease when cross-compiling), so if `--target` wasn't specified here it'd do the whole build from scratch.

aziotd/src/main.rs
arsing: Yeah, I was just avoiding an allocation, but it's not a big deal for a test.

Done.

aziot/src/init.rs
daprilik: sorry, I don't follow :grimacing: 

What does this test block have to do with `argv[1]`?

aziotd/src/main.rs
daprilik: Would `if let toml::Value::Table(ref mut base) = base` do the right thing in this case?

aziot/src/init.rs
arsing: Oh my bad, you're asking about the init test.

The answer to that is https://github.com/Azure/iot-identity-service/blob/becac6833628aece78cddffe9052f5c81d624d92/aziot/src/init.rs#L170-L172

daprilik: Ahhh, I see.

That's somewhat non-obvious though, maybe add a comment on the test module explaining why it's disabled in release mode?

aziotd/src/main.rs
arsing: Actually, I must've shuffled some of the code around after that because I don't see that error any more, so I'll just take it out.

aziot/src/init.rs
arsing: I'll just move the check to the `run()` function.

daprilik (approved):

massand (approved):
:shipit:

PR #58, opened 2020-10-28
Build improvements.

arsing:
- Use a mktemp'd file for cbindgen output rather than creating one in
  the repo directory. This ensures directory mtime isn't updated from
  the temp file, though it doesn't matter at this point since neither make
  nor cargo care about that.

- Detect the OS from /etc/os-release instead of needing the CONTAINER_OS env var
  to be set.

- Consistently print error messages when failing due to unsupported OSes or
  ARCHes, and when a dependency has an unsupported license.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -64,16 +64,19 @@ SHELL := /bin/bash
 
 
 default:
-	cd key/aziot-keys/ && $(CBINDGEN) --config cbindgen.toml --output aziot-keys.h.tmp $(CBINDGEN_VERBOSE)
-	cp key/aziot-keys/cbindgen.prelude.h key/aziot-keys/aziot-keys.h.new
-	< key/aziot-keys/aziot-keys.h.tmp grep -v 'cbindgen_unused_' >> key/aziot-keys/aziot-keys.h.new
-	$(RM) key/aziot-keys/aziot-keys.h.tmp
-	if ! diff -q key/aziot-keys/aziot-keys.h key/aziot-keys/aziot-keys.h.new; then \
-		mv key/aziot-keys/aziot-keys.h.new key/aziot-keys/aziot-keys.h; \
+	# Re-generate aziot-keys.h if necessary
+	set -euo pipefail; \
+	aziot_keys_h_new="$$(mktemp --tmpdir 'aziot-keys.h.new.XXXXXXXXXX')"; \
+	trap "$(RM) '$$aziot_keys_h_new'" EXIT; \
+	cp key/aziot-keys/cbindgen.prelude.h "$$aziot_keys_h_new"; \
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -64,16 +64,19 @@ SHELL := /bin/bash
...
 
 default:
-	cd key/aziot-keys/ && $(CBINDGEN) --config cbindgen.toml --output aziot-keys.h.tmp $(CBINDGEN_VERBOSE)
-	cp key/aziot-keys/cbindgen.prelude.h key/aziot-keys/aziot-keys.h.new
-	< key/aziot-keys/aziot-keys.h.tmp grep -v 'cbindgen_unused_' >> key/aziot-keys/aziot-keys.h.new
-	$(RM) key/aziot-keys/aziot-keys.h.tmp
-	if ! diff -q key/aziot-keys/aziot-keys.h key/aziot-keys/aziot-keys.h.new; then \
-		mv key/aziot-keys/aziot-keys.h.new key/aziot-keys/aziot-keys.h; \
+	# Re-generate aziot-keys.h if necessary
+	set -euo pipefail; \
```
daprilik, line 68: any chance we could use something like https://www.gnu.org/software/make/manual/html_node/One-Shell.html (see [also](https://stackoverflow.com/questions/53541081/embedding-bash-script-into-makefile)) to avoid all the `\` at the end of each line?

daprilik (approved):

ci/install-runtime-deps.sh
```diff
@@ -2,7 +2,11 @@
 
 # This script is meant to be sourced.
 
-case "$CONTAINER_OS" in
+
+OS="$(. /etc/os-release; echo "$ID:$VERSION_ID")"
```
CindyXing, line 6: So the architecture/flavor is an env.; and OS distro:version is from this local file. 
Why not use the same approach for the arch/flavor as well?

Makefile
arsing: The non-oneshell uses outweigh the oneshell uses, so I'd rather not have to remember to prefix every target with `set -e`.

ci/install-runtime-deps.sh
arsing: `ARCH` is used for cross-compiling. It's not limited to the arch of the build machine.

Makefile
daprilik: Sure, that's fine. Just thought I'd bring it up.

massand (approved):
:shipit:

ci/install-build-deps.sh
```diff
@@ -6,22 +6,25 @@
 # modify the contents of /etc/apt. The script is intended to be run inside a container of the corresponding OS, not directly on your machine.
```
CindyXing, line 6: Can we add modify the contents of /etc/apt and /etc/os-release here.

CindyXing (approved):

ci/install-build-deps.sh
arsing: `/etc/os-release` is not modified.

PR #59, opened 2020-10-28
Proposed updates to package naming and init behavior(s) plus minor tweaks

micahl:
Switched from 'iotedge-aziot' to 'aziot-edge'
Noted that the aziot-edged config is still YAML
Proposed behavior for 'iotedge init' with helper options
Created additional section headers and reorganized some content that felt buried after the technical details

I didn't remove all instances of 'migrate', but I'm trying to minimize use of the term which connotes a more permanent movement (i.e. remove the old config after creating the new)  We intentionally don't do that making it more akin to an 'import' than a migration.


docs/packaging.md
```diff
@@ -250,37 +250,59 @@ apt install aziot-identity-service
...
+apt install aziot-edge
+
+iotedge init
+```
 
-## Installation procedure (`iotedge-aziot`)
+After installing the `aziot-edge` package, run `iotedge init` to interactively set up the configuration. It performs the initialization for both the IS+KS+CS components installed by the `aziot-identity-service` package and the MR component installed by the `aziot-edge` package.
 
+## Updating from `iotedge` to `aziot-edge`
 ```sh
```
arsing, line 267: Missing empty line after heading.

docs/packaging.md
```diff
@@ -250,37 +250,59 @@ apt install aziot-identity-service
...
 
-iotedge migrate
+```sh
+iotedge init --import
 ```
 
-The user removes the existing `iotedge` and `libiothsm-std` packages, installs the new package, then runs `iotedge migrate` to migrate the configuration of the old IoT Edge installation to the new one. It is important that the user uninstalls the `iotedge` package before installing the `iotedge-aziot` (or even the `aziot-identity-service` package; we do not want a situation where the services from both packages are running at the same time because they will step over each other trying to provision the device and trying to manage Docker modules. We will enforce this in the packages by having them conflict with each other, so that the distribution's package manager will not allow them both to be installed at the same time either.
+### Automating Upgrades of `iotedge` to `aziot-edge`
+We expect that users will manually run the tool when updating the install on the device. Of course, if it has tested on M devices and the user is confident that it will succeed on the remaining N devices, they can use some custom deployment tooling to automatically perform the update at scale across all their devices.
```
arsing, line 293: Missing empty line after heading.

docs/packaging.md
```diff
@@ -250,37 +250,59 @@ apt install aziot-identity-service
...
+The process of importing the configuration is intentionally designed to be run manually, rather than being done automatically by the new services. It is potentially fallible and could offline the device. 
 
-The precise details of the migration are still being worked out. A high-level view is:
+### Downgrading
+The old configuration from `iotedge` is not removed from its location by any of the above actions; therefore, downgrading from the new package to the old one simply involves uninstalling the `aziot-edge` and `aziot-identity-service` packages, then reinstalling the `iotedge` package.
```
arsing, line 297: Missing empty line after heading.

docs/packaging.md
```diff
@@ -290,8 +312,5 @@ The precise details of the migration are still being worked out. A high-level vi
...
 
-The process of migration is intentionally designed to be run manually by the user, rather than being done automatically by the new services, because it is both fallible and could potentially offline the device. Therefore we expect that users will manually run the tool to update the device. Of course, if the user has tested on M devices and is confident that it will succeed on their remaining N devices, they can use some custom deployment tooling to automatically perform the migration at scale across all their devices.
-
-For the files that are copied to new locations, it is important to note that they are not deleted from their previous locations. Therefore the user can still downgrade from the new package to the old one by uninstalling the `iotedge-aziot` and `aziot-identity-service` packages, then reinstalling the `iotedge` package.
-
+## Open Issues
 It is not yet certain whether Edge Agent, Edge Hub and other modules will be able to decrypt any data when running against the new services that they previously encrypted using the workload API with the old service.
```
arsing, line 316: Missing empty line after heading.

docs/packaging.md
```diff
@@ -250,37 +250,59 @@ apt install aziot-identity-service
...
-The user installs the package, then runs `iotedge init` to set up the configuration of the IS+KS+CS+MR components.
+The user must remove the existing `iotedge` and `libiothsm-std` packages before installing the `aziot-edge` package (or even the `aziot-identity-service` package). We do not want a situation where the services from both packages are running at the same time. They would step over each other trying to provision the device and manage Docker modules. We enforce mutual exclusivity between the packages by having them conflict with each other so that the distribution's package manager does not allow them both to be installed at the same time.
 
+The `iotedge init` automatically detects when the configuration of the old IoT Edge installation is available and prompts whether to create the new configuration based on the old. Agreeing to import the old configuation does not remove it. This allows downgrading from `aziot-edge` to `iotedge`.
 
-## Migration procedure for existing users of `iotedge` to `iotedge-aziot`
+### Initialization Options
 
+The `--force` option can be used to force the initialization sequence.  This rewrites the new configuration values.
 ```sh
```
arsing, line 282: Missing empty line before code block.

arsing (changes requested):

arsing (approved):
The proposal seems fine to me. Rajeev and Cindy should sign off too.

docs/packaging.md
```diff
@@ -1,12 +1,12 @@
...
+The Identity Service (IS), Keys Service (KS) and Certificates Service (CS) have been explicitly designed to not be Azure IoT Edge-specific; they can be used on non-Edge Azure IoT devices also. To this end, we want to cleanly separate these components from existing association with IoT Edge in the form of hosting it in a separate source repository and shipping it as a separate package. Furthermore, the existing IoT Edge daemon (`iotedged`) will be modified to depend on these components for provisioning the device, managing module identities, and managing cryptographic keys and certificates. Its only responsibility will be to act as a Module Runtime (MR).
```
CindyXing, line 3: It would be great to be more specific. How about below:

"Iotedged's only responsibility will be to act as a container Module Runtime. "

docs/packaging.md
```diff
@@ -250,37 +250,59 @@ apt install aziot-identity-service
 aziot init
 ```
 
-The user installs the package, then runs `aziot init` to set up the configuration with minimal information like the device provisioning method.
+After installng the package, run `aziot init` to interactively set up the configuration with minimal information like the device provisioning method.
```
CindyXing, line 253: run 'iotedge init'

docs/packaging.md
```diff
@@ -250,37 +250,59 @@ apt install aziot-identity-service
...
+
+apt install aziot-edge
 
 iotedge init
 ```
 
-The user installs the package, then runs `iotedge init` to set up the configuration of the IS+KS+CS+MR components.
+You must remove the existing `iotedge` and `libiothsm-std` packages before installing the `aziot-edge` package (or even the `aziot-identity-service` package). We do not want a situation where the services from both packages are running at the same time. They would step over each other trying to provision the device and manage Docker modules. We enforce mutual exclusivity between the packages by having them conflict with each other so that the distribution's package manager does not allow them both to be installed at the same time.
 
+The `iotedge init` automatically detects when the configuration of the old IoT Edge installation is available and prompts whether to create the new configuration based on the old. Agreeing to import the old configuation does not remove it. This allows downgrading from `aziot-edge` to `iotedge`.
```
CindyXing, line 278: What do we mean by "this allows downgrading from ... to .."? The old configuration file shouldn't be overwritten in place, right? 

I am not sure if this line is necessary. Or we can share the details in the downgrade step.

docs/packaging.md
```diff
@@ -250,37 +250,59 @@ apt install aziot-identity-service
...
 
-## Migration procedure for existing users of `iotedge` to `iotedge-aziot`
+### Initialization Options
 
+The `--force` option can be used to force the initialization sequence.  This rewrites the new configuration values.
```
CindyXing, line 282: The rewrite will result in a new configuration file(s).

docs/packaging.md
```diff
@@ -102,10 +102,10 @@ Note that the configuration is now in TOML format.
 ```
 /etc/aziot/
 └── edged/
-    └── config.toml
+    └── config.yaml
 ```
 
-Note that the configuration is now in TOML format.
+Note that the configuration is still in YAML format, but reduced in scope.
```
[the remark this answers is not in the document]
micahl: Yes, per a comment someone made when we were reviewing the configuration / migration scenarios.

docs/packaging.md
micahl: How about, "The `iotedged`'s binary name will become `aziot-edged` and its only responsibility will be to act as a Module Runtime for containerized Edge modules."?

docs/packaging.md
micahl: This particular example would be if someone were choosing to not use IoT Edge and planned to only install the identity service for other purposes.

docs/packaging.md
micahl: I'll remove the last sentence as it isn't necessary here.

docs/packaging.md
CindyXing: I thought we plan to enable the globally unified tool in the future. 

If aziot init is still valid, it would be helpful to elaborate it crisp clear here.

docs/packaging.md
CindyXing: Yeah, I saw the downgrade step has details laid out already. Here having the line could be confusing.

docs/packaging.md
micahl: How about "The `--force` option can be used to force the initialization sequence to select new configuration values. This rewrites a new configuration file(s)."

docs/packaging.md
micahl: I made a change to to briefly highlight that `iotedged` is renamed to `aziot-edged` in the prior sentence.  Then in this one I went with basically what you proposed:  `aziot-edged`'s only responsibility will be to act as a Module Runtime (MR) for containerized Edge modules.

docs/packaging.md
micahl: I changed the section name to be "Installation procedure for non-IoT Edge (`aziot-identity-service` only)".  I also moved it to appear after the discussion of IoT Edge installation and included to clarify reiterate the scope of what IS+KS+CS provide.  Although it is repeating parts from the page intro I thought it would be useful to pre-empt questions someone might mentally ask what they'd get if they only used IS+KS+CS.

massand (approved):

CindyXing (approved):

PR #61, opened 2020-10-28
Add API to issue local identities

gordonwang0:
`GET /identities/modules/{module-name}?api-version=2020-09-01&type=local`

The query parameter `type=local` causes IS to return a local identity. When not specified, the module identity (if any) will be returned.

Local identities are issued with ephemeral keys not stored by IS. Therefore, every request to IS for a local identity will provide a new cert and key pair.


identity/aziot-identityd/src/lib.rs
```diff
@@ -702,83 +634,191 @@ impl Api {
...
+                    let (cert, key) = self
+                        .create_identity_cert_if_not_exist_or_expired(
+                            KeyType::Raw,
+                            module_id,
+                            subject.as_str(),
+                            Some(attributes),
+                        )
+                        .await?;
+
+                    aziot_identity_common::Identity::Local(aziot_identity_common::LocalIdSpec {
```
massand, line 677: We need a new spec format with a new type of different auth_type that applies only to LocalIdSpec because you are not always passing the key_handle back now.

identity/aziot-identityd/src/lib.rs
```diff
@@ -825,8 +865,8 @@ fn create_csr(
     subject_name.append_entry_by_nid(openssl::nid::Nid::COMMONNAME, subject)?;
     let subject_name = subject_name.build();
     csr.set_subject_name(&subject_name)?;
-    csr.set_pubkey(public_key)?;
-    csr.sign(private_key, openssl::hash::MessageDigest::sha256())?;
+    csr.set_pubkey(key)?;
```
massand, line 868: The private key shouldn't be set as the public key in the CSR. Please revert this.

gordonwang0: This was something I wasn't sure about, but it since openssl::pkey::Private implements the HasPublic trait, it should be fine to pass the private key to set_pubkey, and that function will generate the public key from the private key. Is that not how this works?

identity/aziot-identityd/src/lib.rs
gordonwang0: What should the spec look like? Currently, it's
```
{
   "type":"local",
   "spec":{
      "attr":"client",
      "auth":{
         "type":"x509",
         "keyHandle":"...",
         "certId":"..."
      }
   }
}
```

I'll rename keyHandle and certId to key and cert, since they'll be raw bytes. Any other changes needed?

massand: Checking the identity workload API. The expiration is also returned there - [https://github.com/Azure/iotedge/blob/iiot/edgelet/edgelet-http-workload/src/server/cert/mod.rs#L33](https://github.com/Azure/iotedge/blob/iiot/edgelet/edgelet-http-workload/src/server/cert/mod.rs#L33)

---
In reply to: [513769412](https://github.com/Azure/iot-identity-service/pull/61#discussion_r513769412) [](ancestors = 513769412)

identity/aziot-identityd/src/lib.rs
massand: It depends on the algorithm and exponent used. For provisioning, identity cert key openssl pkeyrefs are obtained using the key engine and we cannot assume anything about the properties of the key. 

---
In reply to: [513767139](https://github.com/Azure/iot-identity-service/pull/61#discussion_r513767139) [](ancestors = 513767139)

gordonwang0: Got it. Fixed.

identity/aziot-identityd/src/lib.rs
gordonwang0: Changed the spec to add expiration date.
```
{
   "type":"local",
   "spec":{
       "private_key":"...",
       "certificate":"...",
       "expiration":"..."
   }
}
```

identity/aziot-identityd/src/lib.rs
```diff
@@ -702,76 +634,199 @@ impl Api {
...
-        }
+            let result = async {
+                let csr =
+                    create_csr(&subject, &public_key, &private_key, attributes).map_err(|err| {
+                        Error::Internal(InternalError::CreateCertificate(Box::new(err)))
+                    })?;
+
+                let cert = self
+                    .cert_client
+                    .create_cert(&identity_cert, &csr, None)
```
massand, line 788: What's your issuer for these local identities? I thought all the certs would chain to an independent local identities CA certificate.

identity/aziot-identityd/src/lib.rs
```diff
@@ -702,76 +634,199 @@ impl Api {
...
+                    let current_time =
+                        openssl::asn1::Asn1Time::days_from_now(0).map_err(|err| {
+                            Error::Internal(InternalError::CreateCertificate(Box::new(err)))
+                        })?;
+                    let expiration_time = current_time.diff(cert_expiration).map_err(|err| {
+                        Error::Internal(InternalError::CreateCertificate(Box::new(err)))
+                    })?;
+
+                    if expiration_time.days < 1 {
+                        None
```
massand, line 717: We should log a warning that the identity certificates is being renewed to ensure > 24 hour expiry.

identity/aziot-identityd/src/lib.rs
```diff
@@ -702,76 +634,199 @@ impl Api {
...
+                            subject.as_str(),
+                            Some(attributes),
+                        )
+                        .await?;
+
+                    // Parse certificate expiration time.
+                    let expiration = get_cert_expiration(&certificate)?;
+
+                    aziot_identity_common::Identity::Local(aziot_identity_common::LocalIdSpec {
+                        private_key,
```
massand, line 681: We should add the module_id in the response.

identity/aziot-identity-common/src/lib.rs
```diff
@@ -31,6 +32,13 @@ pub struct AzureIoTSpec {
     pub auth: Option<AuthenticationInfo>,
 }
 
+#[derive(Clone, Debug, serde::Deserialize, serde::Serialize)]
+pub struct LocalIdSpec {
```
massand, line 36: missing module_id field. Also, the auth info should be nested for extensibility (since you were looking at supporting other forms of auth?)

identity/aziot-identityd/src/lib.rs
gordonwang0: Yes, the issuer is on an EST server. The CSR is sent to CS, which then makes the EST request to sign.

identity/aziot-identityd/src/lib.rs
gordonwang0: Added. Note that I made it info level since I don't really think cert expiry (which is expected to happen everntually) should be a warning.

identity/aziot-identityd/src/lib.rs
gordonwang0: Added

identity/aziot-identity-common/src/lib.rs
gordonwang0: Fixed

identity/aziot-identityd/src/lib.rs
```diff
@@ -702,76 +634,204 @@ impl Api {
...
+                            expiration,
+                        },
+                    })
+                }
+            };
+
+        Ok(local_identity)
+    }
+
     async fn create_identity_cert_if_not_exist_or_expired(
```
massand, line 694: This method has a lot of tight coupling with the callers in a strange way. So far, there are two types of identities certs (local and device provisioning) and they both always use different types of key types. Is it worth creating two separate create_device_identity_cert (that doesn't need to return anything) and create_module_identity_cert  (that returns the cert pem and key pem) methods? This way the code becomes more readable and makes the caller code consistent with the return type (i.e. dps_509_provisioning logic doesn't need to do anything with the return value as before)

identity/aziot-identityd/src/lib.rs
```diff
@@ -702,76 +634,204 @@ impl Api {
...
+                device_id_cert = Some(cert);
+
+                Ok::<(), Error>(())
+            }
+            .await;
 
-        Ok(())
+            if let Err(err) = result {
+                if let KeyType::Handle(_key_id) = identity_pk {
+                    // TODO: need to delete key from keyd.
```
massand, line 806: Implement the TODO?

identity/aziot-identityd/src/lib.rs

identity/aziot-identityd/src/lib.rs
gordonwang0: Not possible right now -- there's no KS API to delete keys.

identity/aziot-identityd/src/lib.rs
```diff
@@ -702,73 +622,164 @@ impl Api {
...
+                if expiration_time.days < 1 {
+                    log::info!("{} has expired. Renewing certificate", identity_cert);
+
+                    None
+                } else {
+                    Some(pem)
+                }
             }
-            Err(_) => true,
+            Err(_) => None,
```
massand, line 740: We need to check the type of error. If the CS was not accessible, we should not try and delete the preloaded key.

gordonwang0: Yes. Though since we don't actually delete the key right now, I just left it as that.

massand: Please add a note in the TODO comment to ensure that case doesn't get missed when we implement the delete key. Thanks!

massand (approved):

identity/aziot-identityd/src/lib.rs
gordonwang0: Added

PR #62, opened 2020-10-30
Add basic E2E test workflows.

arsing:
This commit adds two GH workflows that run an E2E test script against
all supported OSes in dynamically-created Azure VMs against
dynamically-created IoT Hubs. The test installs the latest
aziot-identity-service package, configures the provisioning method according to
the test parameters, registers a single host process module, and asserts that
identityd is able to complete provisioning and the module is able to get
its identity.

One workflow is a scheduled workflow that runs once a day and is only usable on
the Azure/iot-identity-service repo. The other can be triggered manually
via the github.com UI and is primarily meant for forks, so that devs can
run them against their work branches.

The test script can also be run locally instead of in a GH workflow.

The script uses an OS-appropriate package from the latest successful
packages workflow run for the specified git branch. When running locally,
it can be told to use a locally-built package instead.

Currently the setup has the following limitations:

- Only one provisioning method is tested - manual+symmetric-key device identity.
  Manual+X509 and DPS tests will be added later.

- Only the amd64 architecture is tested. Testing other architectures
  cannot be done in Azure VMs and will require our own hosted GH agents on
  arm32v7 and aarch64 hardware.

- PKCS#11 is not tested. PKCS#11 tests using tpm2-pkcs11 with ibmswtpm2 can
  be run on the Azure VMs, and will be done in the future.

Other changes:

- Fix file extension check when merging configs.

- Clarify service docs that `/run/aziot` also needs to exist and be writable.

- Use systemd helpers in RPM spec to handle enabling and restarting.

### DIFF
--- a/.github/workflows/e2e-tests-manual.yaml
+++ b/.github/workflows/e2e-tests-manual.yaml
@@ -46,5 +46,6 @@ jobs:
         AZURE_USERNAME: "${{ secrets.AZURE_USERNAME }}"
         AZURE_PASSWORD: "${{ secrets.AZURE_PASSWORD }}"
         AZURE_RESOURCE_GROUP_NAME: "${{ secrets.AZURE_RESOURCE_GROUP_NAME }}"
+        AZURE_LOCATION: "${{ secrets.AZURE_LOCATION }}"
 
         GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
--- a/.github/workflows/e2e-tests-manual.yaml
+++ b/.github/workflows/e2e-tests-manual.yaml
@@ -1,0 +1,50 @@
+# Used to run the E2E tests against a branch of your choice from the github.com UI.
+#
+# Note that the workflow file is always used from main, even when running against a different branch.
+#
+# DEVNOTE: Keep in sync with e2e-tests-scheduled.yaml
+
+name: 'e2e-tests-manual'
+
+on:
+  'workflow_dispatch':
+    inputs:
+      branch:
+        default: 'main'
+
+jobs:
+  e2e-tests:
+    runs-on: 'ubuntu-18.04'
+
+    strategy:
+      fail-fast: false
+
+      matrix:
+        os:
+        - 'centos:7'
+        - 'debian:9'
+        - 'debian:10'
+        - 'ubuntu:18.04'
+        - 'ubuntu:20.04'
+        test_name:
+        - 'manual-symmetric-key'
+
+    steps:
+    - uses: 'actions/checkout@v1'
+      with:
+        ref: "${{ github.events.input.branch }}"
+
+    - name: 'Run'
+      run: |
+        ./ci/e2e-tests.sh ${{ matrix.test_name }}
+      env:
+        BRANCH: "${{ github.events.input.branch }}"
+
+        OS: "${{ matrix.os }}"
+
+        AZURE_TENANT_ID: "${{ secrets.AZURE_TENANT_ID }}"
+        AZURE_USERNAME: "${{ secrets.AZURE_USERNAME }}"
+        AZURE_PASSWORD: "${{ secrets.AZURE_PASSWORD }}"
+        AZURE_RESOURCE_GROUP_NAME: "${{ secrets.AZURE_RESOURCE_GROUP_NAME }}"
+
+        GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
--- a/ci/e2e-tests.sh
+++ b/ci/e2e-tests.sh
@@ -149,7 +149,7 @@ echo "Run ID: $run_id" >&2
 
 # Temp directory used as scratch space to store the downloaded package from GitHub
 # and the config files for the package that are scp'd to the test VM.
-working_directory="iot-identity-service-e2e-tests-$run_id.XXXXXXXXXX"
+working_directory="iot-identity-service-e2e-tests-$run_id"
 echo "Working directory: $working_directory" >&2
 mkdir -p "$working_directory"
 cd "$working_directory"
--- a/ci/e2e-tests.sh
+++ b/ci/e2e-tests.sh
@@ -149,7 +149,7 @@ echo "Run ID: $run_id" >&2
 
 # Temp directory used as scratch space to store the downloaded package from GitHub
 # and the config files for the package that are scp'd to the test VM.
-working_directory="iot-identity-service-e2e-tests-$run_id"
+working_directory="iot-identity-service-e2e-tests-${run_id//:/-}"
 echo "Working directory: $working_directory" >&2
 mkdir -p "$working_directory"
 cd "$working_directory"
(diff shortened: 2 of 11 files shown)


ci/e2e-tests.sh
```diff
@@ -0,0 +1,615 @@
...
+
+echo 'Creating allow-ssh rule in NSG...' >&2
+>/dev/null az network nsg rule create \
+    --resource-group "$AZURE_RESOURCE_GROUP_NAME" \
+    --nsg-name "$common_resource_name" \
+    --name 'ssh' \
+    --priority 1000 \
+    --access 'Allow' --direction 'Inbound' --protocol 'Tcp' \
+    --destination-port-ranges '22' \
+    --source-address-prefixes "$self_ip/32"
```
damonbarry, line 391: 👍

damonbarry (approved):
LGTM

.github/workflows/e2e-tests-manual.yaml
```diff
@@ -0,0 +1,50 @@
+# Used to run the E2E tests against a branch of your choice from the github.com UI.
```
gordonwang0, line 1: Is it possible to merge this file and e2e-tests-scheduled by having both `workflow_dispatch` and `cron` under `on`?

ci/e2e-tests.sh
```diff
@@ -0,0 +1,615 @@
+#!/bin/bash
+
+# Usage:
+#
+# e2e-tests.sh <test_name>
+#
+# See /docs/dev/e2e-tests.md for details of some env vars that need to be defined.
```
gordonwang0, line 7: Should this link to the html page? I assume this page will be available with the rest of the docs.

ci/e2e-tests.sh
```diff
@@ -0,0 +1,615 @@
...
+test_name="$1"
+
+
+run_id="$OS:$GITHUB_RUN_ID:$GITHUB_RUN_NUMBER:$test_name"
+echo "Run ID: $run_id" >&2
+
+
+# Temp directory used as scratch space to store the downloaded package from GitHub
+# and the config files for the package that are scp'd to the test VM.
+rm -rf /tmp/iot-identity-service-e2e-tests
```
gordonwang0, line 152: How about `mktemp -d`?

ci/e2e-tests.sh
```diff
@@ -0,0 +1,615 @@
...
+    az network nsg create \
+        --resource-group "$AZURE_RESOURCE_GROUP_NAME" \
+        --name "$common_resource_name" \
+        --tags "$resource_tag" \
+        --query 'id' --output tsv
+)"
+echo 'Created NSG' >&2
+
+echo 'Querying public IP...' >&2
+self_ip="$(curl -L 'https://ipinfo.io/ip')"
```
gordonwang0, line 380: I assume this site only provides IPv4 addresses? Since the nsg rule create won't work if this gives IPv6.

gordonwang0: Just remembered the Azure VMs only have public ipv4 addresses, ignore this.

.github/workflows/e2e-tests-manual.yaml
arsing: The job is different between the two. `-manual` has an input for the branch whereas `-scheduled` has a matrix dimension.

ci/e2e-tests.sh
arsing: It's convenient to have a fixed directory for debugging.

Thinking about it now, it would be nice to be able to run multiple runs at the same time. I'll add the `$run_id` to the directory name.

.github/workflows/e2e-tests-manual.yaml
gordonwang0: Could BRANCH be set with an if? Something like "if github.events.input.branch is set, use it; else use main"?

arsing: I'm not talking about the env var. I'm talking about the matrix dimension.

https://github.com/Azure/iot-identity-service/blob/995f6306b3c45a723af6b215b2c1a67126286db1/.github/workflows/e2e-tests-scheduled.yaml#L23-L24

gordonwang0:
```
branch:
- 'main'
  if github.event_name = ...
```
Haven't tried it, but would that work?

arsing: https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions

gordonwang0: Yeah, I wasn't paying attention to the correct syntax.

Point is: I don't think we should have two almost identical files that require manual updates to both (`DEVNOTE: Keep in sync with ...`) because when making changes to one of them, it's easy to forget to update the other. So combining them into one file would be more maintainable in the future. I'm not familiar with GitHub Actions (haven't used it before) but I'd imagine there's some way to do this.

arsing: Sure. If you find some way to do it, please tell me. I did this after already having checked, but of course I could've missed something.

ci/e2e-tests.sh
arsing: Done.

ci/e2e-tests.sh
arsing: Done.

.github/workflows/e2e-tests-manual.yaml
gordonwang0: I'll take a look, maybe in a few days. I'll sign off for now.

ci/e2e-tests.sh
```diff
@@ -149,9 +149,10 @@ echo "Run ID: $run_id" >&2
 
 # Temp directory used as scratch space to store the downloaded package from GitHub
 # and the config files for the package that are scp'd to the test VM.
-rm -rf /tmp/iot-identity-service-e2e-tests
-mkdir -p /tmp/iot-identity-service-e2e-tests
-cd /tmp/iot-identity-service-e2e-tests
+working_directory="iot-identity-service-e2e-tests-$run_id.XXXXXXXXXX"
```
gordonwang0, line 152: Why append XXXXXXXXXX to the directory name?

gordonwang0 (approved):

ci/e2e-tests.sh
arsing: Ah, accidental leftover from experimenting with `mktemp`.

ci/e2e-tests.sh
massand, line 381: You can use this to determine the public IP address - `az vm list-ip-addresses -g MyResourceGroup -n MyVm`

arsing: This is the IP of the machine running the GH action.

massand (approved):

massand (approved):
:shipit:

gordonwang0 (approved):

CindyXing (approved):
Great to see this. Thanks Arnav for the efforts.

arsing:
I'm going to merge this while we're still sorting out the Azure subscription for this repo. It means the daily scheduled workflow run is going to fail (and send us emails about it), but at least we'll be have the manual action that we can use from our forks with our own subscriptions.

PR #63, opened 2020-11-02
Disable hard warnings + rustfmt during local dev

daprilik:
This PR replaces the `#![deny(warnings, clippy::all, clippy::pedantic)]` directives with `#![warn(clippy::all, clippy::pedantic)]` + passing `-D warning -D clippy::pedantic -D clippy::all` when running in the CI.

The reasoning for this is outlined in https://github.com/rust-unofficial/patterns/blob/master/anti_patterns/deny-warnings.md, but TL;DR, it gives developers the option to leave some warnings in the code while testing new changes, while nonetheless enforcing warning-free code in commits. 

The Makefile has been modified to include a new target `test-release`. This target depends on the `test` target, while also setting the corresponding `-D` flags as well. Additionally, the `rustfmt` check has been moved into the `test-release` target, giving developers the option to only run `rustfmt` once it's time to push a commit.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -26,7 +26,7 @@ contact [opencode@microsoft.com](mailto:opencode@microsoft.com) with any additio
 
     This is enforced by `make test-release`.
 
-    When developing locally, consider using `make test` instead, which does not enforcing rustfmt code formatting.
+    Make sure to run `make test-release` before pushing to your remote so that you run all the checks that CI would run, as `make test` does not enforce rustfmt!
 
 1. Every crate root must enable the major lint groups and warnings-as-errors:
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -20,16 +20,22 @@ contact [opencode@microsoft.com](mailto:opencode@microsoft.com) with any additio
     - Rust: `// Copyright (c) Microsoft. All rights reserved.`
     - C: `/* Copyright (c) Microsoft. All rights reserved. */`
 
-    This is enforced by `make test`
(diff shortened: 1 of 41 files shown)


Makefile
```diff
@@ -151,6 +151,12 @@ target/openapi-schema-validated:
 	touch target/openapi-schema-validated
 
 
+test-release: export RUSTFLAGS += -D warnings
```
arsing, line 154: This sets it for all dependencies, not just the current workspace.

Do you even need this? `CLIPPY_FLAGS` will be enough, no?

daprilik: Oh, you're right, it does set it for all deps.

Sure, I guess it's not really necessary, since clippy catches regular warnings as well. Will fix.

CONTRIBUTING.md
```diff
@@ -20,30 +20,36 @@ contact [opencode@microsoft.com](mailto:opencode@microsoft.com) with any additio
...
     - C: `/* Copyright (c) Microsoft. All rights reserved. */`
 
-    This is enforced by `make test`
+    This is enforced by `make test`.
+
+1. Rust code must be formatted with `rustfmt`
+
+    This is enforced by `make test-release`.
+
+    When developing locally, consider using `make test` instead, which does not enforcing rustfmt code formatting.
```
arsing, line 29: "enforcing" -> "enforce"

Also, probably something like:

```diff
+But make sure to run `make test-release` before you push to your remote so that you run all the checks that CI would run.
```

arsing (approved):

massand (approved):

PR #64, opened 2020-11-03
DPS refactoring

daprilik:
While working on TPM provisioning, I noticed that the DPS provisioning code use some refactoring. 
Some notable changes so far:

- Extracts the DPS registration polling into a separate attestation-method agnostic helper function.
- Refactors the DPS client from a series of standalone methods into a proper `Client` struct.
- Use an enum to encapsulate X509 and Symmetric Key specific parameters, reducing the number of Option unwraps/expects.

Aside from a small modification to the identityd config + a stubbed code path, this PR doesn't actually include any substantial DPS-TPM code, as it requires the `tpmd` infrastructure that only exists on my local branch.

I wanted to open this as a separate PR to get some early feedback on these changes, and to hopefully get some pointers on what I should expect once I add TPM provisioning into the mix (i.e: did I accidentally over-abstract).

### DIFF
--- a/identity/aziot-dps-client-async/src/lib.rs
+++ b/identity/aziot-dps-client-async/src/lib.rs
@@ -15,6 +15,8 @@
 
 use std::sync::Arc;
 
+use aziot_cloud_client_async_common::{get_sas_connector, get_x509_connector};
+
 pub mod model;
 
 pub const DPS_ENCODE_SET: &percent_encoding::AsciiSet =
@@ -142,7 +144,9 @@ impl Client {
 
         let connector = match auth_kind {
             DpsAuthKind::SymmetricKey { sas_key } => {
-                let (connector, token) = self.get_sas_connector(registration_id, &sas_key).await?;
+                let audience = format!("{}/registrations/{}", self.scope_id, registration_id);
+                let (connector, token) =
+                    get_sas_connector(&audience, &sas_key, &self.key_client).await?;
 
                 let authorization_header_value = hyper::header::HeaderValue::from_str(&token)
                     .map_err(|err| std::io::Error::new(std::io::ErrorKind::Other, err))?;
@@ -154,8 +158,14 @@ impl Client {
                 identity_cert,
                 identity_pk,
             } => {
-                self.get_x509_connector(&identity_cert, &identity_pk)
-                    .await?
+                get_x509_connector(
+                    &identity_cert,
+                    &identity_pk,
+                    &self.key_client,
+                    &mut *self.key_engine.lock().await,
+                    &self.cert_client,
+                )
+                .await?
             }
         };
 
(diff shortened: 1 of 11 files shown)


identity/aziot-identityd/src/lib.rs
```diff
@@ -414,99 +457,47 @@ impl Api {
...
+                let dps_client = aziot_dps_client_async::Client::new(
+                    &global_endpoint,
+                    &scope_id,
+                    self.key_client.clone(),
+                    self.key_engine.clone(),
+                    self.cert_client.clone(),
+                );
+
                 let device = match attestation {
+                    settings::DpsAttestationMethod::TPM { registration_id } => {
```
arsing, line 469: What's the point of adding this in this PR instead of your tpmd one?

identity/aziot-dps-client-async/src/lib.rs
```diff
@@ -13,323 +13,339 @@
...
+
+        let token = format!("SharedAccessSignature {}", token);
+
+        let tls_connector = hyper_openssl::HttpsConnector::new()
             .map_err(|err| std::io::Error::new(std::io::ErrorKind::Other, err))?;
-        for cert in device_id_certs {
+        Ok((tls_connector, token))
+    }
+
+    async fn get_x509_connector(
```
arsing, line 289: The `get_{sas,x509}_connector` fns are duplicated in `aziot-hub-client-async`.

Perhaps leave them as free functions, put them in a common crate and have both `aziot-{dps,hub}-client-async` use them.

arsing (changes requested):

identity/aziot-identityd/src/lib.rs
daprilik: I guess it goes back to the age old debate between "one big PR vs. multiple smaller PRs". Having one big PR for the TPM work makes sense, since reviewing/testing it requires having the full E2E flow implemented. This PR doesn't actually add any new functionality, and just refactors existing code, so I thought I'd break it out into it's own smaller PR.

identity/aziot-dps-client-async/src/lib.rs
daprilik: Oh, hey, would you look at that. Neato.
I'm very for reducing duplication when possible, so I'll go ahead and break those out into their own crate. Any suggestion on the crate name? Off the top of my head... `aziot-auth-client-async-common`? A bit of a mouthful, but it's certainly descriptive...

identity/aziot-identityd/src/lib.rs
arsing: If we merge this today and then tomorrow we decide "Let's not merge Daniel's PR because we've decided to stop supporting TPM provisioning", what will have been the point of adding this code?

Adding any code related to TPM provisioning is unrelated to a PR that refactors existing code. In fact it is the opposite of what a refactoring PR should do.

identity/aziot-dps-client-async/src/lib.rs
arsing: `aziot-http-client(-async)(-common)` perhaps. I'd be fine with just `aziot-http-client`

`-async` is optional since there's no `-sync`. (`aziot-cert-client-async` is a counter-point, but that is symmetric with `aziot-key-client-async`. Though I'm not opposed to renaming it to `aziot-cert-client`)

identity/aziot-identityd/src/lib.rs
daprilik: Oh, sorry, I totally didn't realize that this was a comment specifically about the TPM bit.

Yeah, my bad, I don't know what I was thinking including this in the PR (probably lazyness). I'll got ahead and remove it.

identity/aziot-dps-client-async/src/lib.rs
daprilik: While I'm all for removing the `-async` suffxes across the repo, that's best left for a separate PR.

As for naming... what about `aziot-cloud-client-async`? `cloud` encapsulates both the DPS and the Hub, while leaving it open for additional shared code in the future (as opposed to just `auth` stuff)

arsing: >While I'm all for removing the `-async` suffxes across the repo, that's best left for a separate PR.

I didn't say anything about renaming any existing crates in this PR. I'm talking about your new crate, for which `-async` is optional.

>As for naming... what about `aziot-cloud-client-async-common` ? `cloud` encapsulates both the DPS and the Hub, while leaving it open for additional shared code in the future (as opposed to just `auth` stuff)

It's not DPS- or IoT Hub-specific.

daprilik: Sorry, let me rephrase what I meant - I'll leave `-async` in the name to match the other clients in the repo. At some point, it would be nice to normalize all the crate names to be async-first, and renaming the few synchronous clients with the `-sync` suffix.

* * *

Sure, but this crate contains shared code for talking to the cloud, so it seems somewhat appropriate (even in the concrete contents of the crate isn't strictly cloud-specific at the moment).

I don't have super strong feelings about the name, so let me know if this works / what I should rename it to, and I'll go ahead and push an update.

massand (approved):
:shipit:

identity/aziot-dps-client-async/src/lib.rs
```diff
@@ -13,323 +13,233 @@
...
-            .map_err(|err| std::io::Error::new(std::io::ErrorKind::Other, err))?;
-        req.headers_mut()
-            .append(hyper::header::AUTHORIZATION, authorization_header_value);
-        connector
-    } else {
-        get_x509_connector(
-            identity_cert.expect("device identity certificate not found"),
-            identity_pk.expect("device private key not found"),
+impl Client {
+    #[must_use]
```
arsing, line 45: This is to satisfy `clippy::must_use_candidate` ? You can just `allow` it at the crate level, like `CONTRIBUTING.md` says.

arsing (approved):

identity/aziot-dps-client-async/src/lib.rs
daprilik: I won't lie, this `#[must_use]` is just copied/pasted from another Client implementation. It's nothing major, so I'll just go ahead and merge the PR.

PR #66, opened 2020-11-06
Update identity-service.md with local identity API

gordonwang0:
- Adds local identity to the GET module identity API
- Fixes casing to match current API
- Updates section about IS configuration


docs/identity-service.md
```diff
@@ -238,18 +238,22 @@ The returned `auth.keyHandle` value is meant to be used with the [Keys Service](
 
 ### Get IoT module identity information
 
-`GET /identities/modules/{module-id}?api-version=2020-09-01`
+`GET /identities/modules/{module-id}?api-version=2020-09-01[&type={type}]`
```
massand, line 241: I thought the final version of the API would have a required `type` parameter (not optional). So, should we reflect that now or later?

gordonwang0: I'll update it when we implement the change.

docs/identity-service.md
```diff
@@ -328,32 +349,31 @@ IoT Hub module identities will get generated by Identity service (IS) for contai
...
+Generally, for host process modules, IS needs to be configured with a list of host process userid and module names. Based on its module list in `config.toml`, IS will reconcile module identities with IoT Hub on startup. The creation process is shown in the [Provisioning Flow diagram](img/est-ca-provisioning-simple.svg). Note that a device reprovision could also trigger re-creation of module identities.
 
+For installation, the device administrator will initiate a host process package install (typically an elevated admin operation), which will install IS as a dependency. First, the IS package will install a `config.toml` configuration file with a default configuration file with no configured host processes. Then, the host process package install process (using elevated privileges) will add it's respective host process userid into the `aziotid` group. Finally, the host process package install will configure it's host process userid and name in the IS `config.toml` (as shown in IS configuration above).
```
massand, line 379:
```suggestion
For installation, the device administrator will initiate a host process package install (typically an elevated admin operation), which will install IS as a dependency. First, the IS package will install a `config.toml` configuration file with a default configuration file with no configured host processes. Then, the host process package install process (using elevated privileges) will add it's respective host process userid into the `aziotid` group. Finally, the host process package install will configure it's host process userid and name by placing its own `<package>.toml` (with a `[[principal]]` table entry as shown in IS configuration above) in the IS `/etc/aziot/identityd/config.d` override directory.
```

massand (approved):

PR #68, opened 2020-11-08
[e2e] Use device and module identities to fetch their twins to test the auth process.

arsing:
This commit extends the E2E test script to also
use the device identity to fetch the device twin over MQTT, and
use the module identity to fetch the module twin over MQTT.

With this, the full auth process involving the KS is now validated.
When X.509 device identity tests are implemented in the future, this will also
test the auth process involving the aziot_keys openssl engine and the CS.

Other changes:

- Fix some typos in instructions.

- Eagerly test the AZURE_* env vars so that the test fails if they aren't set.
  This is primarily useful when running as a GH workflow and the secrets
  have not been defined in the repo.

- Move working directory back to being under /tmp like it was supposed to be.


ci/iothub-get-twin.sh
```diff
@@ -0,0 +1,297 @@
...
+case "$auth_type" in
+    'sas')
+        if [ -z "$module_id" ]; then
+            # TODO: Bug? For device ID spec, the key handle is actually the key ID.
+            key_id="$(uri_encode "$key_handle")"
+            key_handle="$(
+                curl --unix-socket '/run/aziot/keyd.sock' \
+                    "http://foo/key/$key_id?api-version=2020-09-01" |
+                    jq '.keyHandle' -er
+            )"
```
arsing, line 249: @massand Is this expected?

massand: Bug, I'll fix it in a PR today (and fix the test if this goes in earlier).

massand (approved):
:shipit:

ci/iothub-get-twin.sh
arsing: Thanks.

gordonwang0 (approved):

PR #69, opened 2020-11-09
Add type parameter to IS APIs

gordonwang0:
Adds the required query parameter `type` to IS API calls.

The *Get module identity* call (`GET /identities/modules/{module-id}`) accepts both `type=aziot` and `type=local`. All other APIs only accept `type=aziot`.

### DIFF
--- a/identity/aziot-identityd/src/http/create_or_list_module_identity.rs
+++ b/identity/aziot-identityd/src/http/create_or_list_module_identity.rs
@@ -23,7 +23,7 @@ impl http_common::server::Route for Route {
         }
 
         let id_type: Option<String> = query.iter().find_map(|q| {
-            if q.0.to_lowercase() == "type" {
+            if q.0 == "type" {
                 Some(q.1.to_string())
             } else {
                 None
@@ -51,7 +51,7 @@ impl http_common::server::Route for Route {
         };
 
         //TODO: get uid from UDS
-        let identities = match api.get_identities(auth_id, self.id_type).await {
+        let identities = match api.get_identities(auth_id, self.id_type.as_deref()).await {
             Ok(v) => v,
             Err(err) => return Err(super::to_http_error(&err)),
         };
@@ -81,7 +81,7 @@ impl http_common::server::Route for Route {
 
         //TODO: get uid from UDS
         let identity = match api
-            .create_identity(auth_id, Some(body.id_type), &body.module_id)
+            .create_identity(auth_id, Some(&body.id_type), &body.module_id)
             .await
         {
             Ok(id) => id,
--- a/identity/aziot-identityd/src/http/create_or_list_module_identity.rs
+++ b/identity/aziot-identityd/src/http/create_or_list_module_identity.rs
@@ -2,6 +2,7 @@
 
 pub(super) struct Route {
     api: std::sync::Arc<futures_util::lock::Mutex<crate::Api>>,
+    id_type: Option<String>,
 }
 
 #[async_trait::async_trait]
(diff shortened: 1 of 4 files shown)


identity/aziot-identityd/src/http/create_or_list_module_identity.rs
```diff
@@ -15,14 +16,23 @@ impl http_common::server::Route for Route {
...
         path: &str,
-        _query: &[(std::borrow::Cow<'_, str>, std::borrow::Cow<'_, str>)],
+        query: &[(std::borrow::Cow<'_, str>, std::borrow::Cow<'_, str>)],
     ) -> Option<Self> {
         if path != "/identities/modules" {
             return None;
         }
 
+        let id_type: Option<String> = query.iter().find_map(|q| {
+            if q.0.to_lowercase() == "type" {
```
arsing, line 26: We don't treat api-version case-insensitively, so we shouldn't treat this case-insensitively either.

identity/aziot-identityd/src/lib.rs
```diff
@@ -28,6 +28,27 @@ const DPS_ASSIGNMENT_RETRY_INTERVAL_SECS: u64 = 10;
...
+/// URI query parameter that identifies module identity type.
+const ID_TYPE_AZIOT: &str = "aziot";
+
+/// URI query parameter that identifies local identity type.
+const ID_TYPE_LOCAL: &str = "local";
+
+macro_rules! match_id_type {
+    ($id_type:ident, $(($type:ident, $action:expr)),+) => {
+        if let Some(id_type) = $id_type {
+            match id_type.as_str() {
```
arsing, line 40: If you're just going to use it as a `&str`, then why not take `Option<&str>` instead of `Option<String>` ?

(clippy would've warned you about this but probably didn't because you're doing it inside a macro.)

arsing (changes requested):

identity/aziot-identityd/src/http/create_or_list_module_identity.rs
gordonwang0: Fixed

identity/aziot-identityd/src/lib.rs

identity/aziot-identityd/src/lib.rs
```diff
@@ -28,6 +28,27 @@ const DPS_ASSIGNMENT_RETRY_INTERVAL_SECS: u64 = 10;
 /// This is the number of seconds to wait for DPS to complete assignment to a hub
 const DPS_ASSIGNMENT_TIMEOUT_SECS: u64 = 120;
 
+/// URI query parameter that identifies module identity type.
+const ID_TYPE_AZIOT: &str = "aziot";
+
+/// URI query parameter that identifies local identity type.
+const ID_TYPE_LOCAL: &str = "local";
+
+macro_rules! match_id_type {
```
arsing, line 37:
```rust
($id_type:ident { $( $type:ident => $action:expr ,)+ })
```

will allow you to write:

```rust
match_id_type!(id_type {
    ID_TYPE_AZIOT => self.id_manager.get_module_identity(module_id).await,
    ID_TYPE_LOCAL => self.issue_local_identity(module_id).await,
});
```

and so on. That looks closer to the regular `match` syntax, and allows trailing commas on all arms.

arsing (approved):

identity/aziot-identityd/src/lib.rs
gordonwang0: Fixed, thanks!

massand (approved):
:shipit:

PR #70, opened 2020-11-11
Fix for device identity key handle

massand:
- Device identity symmetric keys were being stored as `key_id`s instead of `key_handle`s on service launch. Now, the `key_id`s are translated into `key_handle`s by calling the Key Service on each provision.  
- Fix for SAS and X509 connector logic to use `key_handle`s
- Removed workaround for `/identities/device` API in e2e script (validated locally)

### DIFF
--- a/ci/iothub-get-twin.sh
+++ b/ci/iothub-get-twin.sh
@@ -239,16 +239,6 @@ fi
 
 case "$auth_type" in
     'sas')
-        if [ -z "$module_id" ]; then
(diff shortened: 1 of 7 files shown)


arsing:
It would be better to continue storing a key as you were before, and create and return a new handle for every get-identity request. In the future handles might expire after some time, and in any case we don't want anyone to assume the API gives the same handle every time.

massand:
@arsing - you are right. This isn't a long term solution. I will send a new PR.

PR #73, opened 2020-11-13
Various improvements related to aziot-keys.

arsing:
aziot-keys.h:

- Update cbindgen to v0.15. This version emits items in an order that
  more closely matches the Rust order, rather than the previous version's
  behavior of emitting them in alphabetical order.

- Rename `AZIOT_KEYS_STATUS` to `AZIOT_KEYS_RC`,
  `AZIOT_KEYS_SUCCESS` to `AZIOT_KEYS_RC_OK`, and
  `AZIOT_KEYS_ERROR_*` to `AZIOT_KEYS_RC_ERR_*`. `RC` more accurately represents
  the type's role as the return code of functions, and the new names for
  the constants match the other constants by being prefixed by the type's name.

- Remove `AZIOT_RC_ERR_FATAL`. It was not being handled "fatally" by the caller,
  and the implementation can just `abort()` anyway. Our implementation only used
  if when locking one of the many `RwLock`s failed, for which `.expect()` can
  be used, just like it's used for integer conversions.

- Document the remaining undocumented API, and fix wrong and outdated docs on
  existing API.

- Add `#![deny(missing_docs)]` to aziot-keys crate to ensure everything remains
  documented from now on.

Makefile:

- Have `make test-release` assert that there are no uncommitted changes in
  aziot-keys.h

- Stop excluding aziot-keys from the `cargo test`.

  5e2070f7c3c8fe03c1231883723622014d34f743 excluded it from the command because
  of a cargo bug that would cause the crate to be rebuilt over and over again.
  While that bug still exists, the removal of the rlib crate type in
  that commit was sufficient to work around it. Building the crate for tests
  does not trigger it.

- Remove ad-hoc handling of extra system include directory for bindgen.
  bindgen reads an env var for the same purpose, so handling it in the Makefile
  is unnecessary. As a bonus, this env var works for both bindgen-as-a-binary
  (as the Makefile uses it) and bindgen-as-a-build-script-library, though
  none of our crates or deps currently use the latter.

- Add `set -e` when running bindgen to ensure the build fails if bindgen fails.

- Consistently use `set -euo pipefail`

Docs:

- Change dev docs to remove references to installing the stable Rust toolchain
  specifically, and to installing the latest versions of bindgen and cbindgen.
  The former was redundant with the rust-toolchain file. The latter was
  both redundant with install-build-deps.sh and wrong, since the script
  installs specific versions instead of the latest.


docs/dev/building.md
```diff
@@ -46,8 +42,10 @@
 
     This will print something like `/usr/lib/gcc/x86_64-linux-gnu/7/include-fixed/limits.h`
 
-    Then invoke `make` with `BINDGEN_EXTRA_INCLUDE_DIR` set to the directory containing the `limits.h`:
+    Then export an env var to tell `bindgen` (and in turn, libclang) about this directory.
```
daprilik, line 45: nice, this seems like a cleaner solution :+1:

key/aziot-keys/src/lib.rs
```diff
@@ -159,69 +159,187 @@ pub struct AZIOT_KEYS_FUNCTION_LIST_2_0_0_0 {
...
-    /// If `preferred_algorithms` is NULL, it should be interpreted the same as if it was `"*"`.
+    /// The implementation is allowed to be unable to generate a key pair regardless of which algorithms are specified; this is true even if
+    /// the wildcard algorithm is specified.
     ///
-    /// The public key is written to `ppublic_key` and the private key to `pprivate_key`.
-    /// For keys generated by openssl in memory, the private and public components of a key live in the same `EVP_PKEY` value.
-    /// However keys loaded from engines do differentiate between the two, so separate `EVP_PKEY` values are required.
-    /// Even if the implementation generates keys in memory using openssl, it must copy the public parameters out of the key into a new `EVP_PKEY`
-    /// and set `ppublic_parameters` to that.
+    /// If `preferred_algorithms` is `NULL`, it should be interpreted the same as if it was `"*"`.
```
daprilik, line 165: Is "should" being used in the [RFC2119](https://tools.ietf.org/html/rfc2119) sense? i.e: is it fine for an implementation to not do that? If not, "MUST" would be more appropriate.

key/aziot-keys/src/lib.rs
```diff
@@ -390,36 +608,57 @@ pub extern "C" fn cbindgen_unused_AZIOT_KEYS_SIGN_DERIVED_PARAMETERS(
     unimplemented!();
```
daprilik, line 608: (unrelated to this particular PR)

What's the point of this function?

daprilik (approved):
One more thing: consider adding `#![deny(missing_docs)]` to key/aziot-keys/src/lib.rs.

LGTM in general though :+1:

key/aziot-keys/src/lib.rs
arsing: https://github.com/Azure/iot-identity-service/blob/408ed4afaf5bb866ab67fa8a87f845f4575984e9/key/aziot-keys/src/lib.rs#L49-L71

key/aziot-keys/src/lib.rs
arsing: Yes, it's a hard requirement. Changed the `should`s to `must`s.

arsing:
>One more thing: consider adding `#![deny(missing_docs)]` to key/aziot-keys/src/lib.rs.

Yes, good idea. Done.

PR #74, opened 2020-11-18
Add local identity to create_identity

gordonwang0:
Adds local identities to the create_identity API.

```
POST /identities/modules?api-version=2020-09-01

{
  "type": "local",
  "moduleId": "module01",
  "localIdOpts": {
    "type": "x509",
    "attributes": "server"
  }
}
```

### DIFF
--- a/identity/aziot-identity-common-http/src/lib.rs
+++ b/identity/aziot-identity-common-http/src/lib.rs
@@ -51,11 +51,19 @@ pub mod get_device_identity {
 
 pub mod create_module_identity {
     #[derive(Clone, Debug, serde::Deserialize, serde::Serialize)]
+    #[serde(rename_all = "camelCase")]
+    pub enum CreateModuleOpts {
+        LocalIdOpts(aziot_identity_common::LocalIdOpts),
+    }
+
+    #[derive(Clone, Debug, serde::Deserialize, serde::Serialize)]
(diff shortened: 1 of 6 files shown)


identity/aziot-identity-common-http/src/lib.rs
```diff
@@ -50,12 +50,20 @@ pub mod get_device_identity {
 }
 
 pub mod create_module_identity {
+    #[derive(Clone, Debug, serde::Deserialize, serde::Serialize)]
+    #[serde(rename_all = "camelCase")]
```
massand, line 54: >rename_all [](start = 12, length = 10)

You can remove rename_all if this attribute is flattened in the Request struct?

gordonwang0: Without the rename, the request will be:
```json
{
    "type": "local",
    "moduleId": "mqtt-server",
    "LocalIdOpts": {
        "type": "x509",
        "attributes": "server"
    }
}
```
instead of
```json
{
    "type": "local",
    "moduleId": "mqtt-server",
    "localIdOpts": {
        "type": "x509",
        "attributes": "server"
    }
}
```

So I used the rename_all to make localIdOpts consistent with the other JSON keys.

massand (approved):

PR #78, opened 2020-11-27
Change aziot-keys to use pkcs11 directly instead of through pkcs11-openssl-engine.

arsing:
The PKCS#11 openssl engine made sense before the final design that separated
KS, IS and MR, since the expectation was that iotedged would use its EVP_PKEYs
directly.

Now that PKCS#11 is only used through aziot-keys, converting from aziot-keys's
API to openssl's engine API and then back to PKCS#11 API is overkill.

The only other use of pkcs11-openssl-engine was for pkcs11-test. Rather than
delete pkcs11-test too, this commit repurposes it to test
the aziot_keys openssl engine. Documentation has been updated accordingly.

Other changes:

- Fix missing phony `test` target. 47e14a7a60026a81173f54dbb23b2974bd759db3
  renamed it to `test-release` instead of appending `test-release` to the list.

- Fix e2e-tests.sh to exit quickly if the `AZURE_*` env vars are not set.
  5d5bf7b110b2d177c27516ec3f2245d586096735 tried to do this by asserting that
  the variables were defined, but that check was insufficient. When the secrets
  aren't defined in GitHub, the variables are defined and empty.

- Consistently use `;;` after `exit` in `case` arms.

- Remove P-384, P-521 and ED25519-related code from the PKCS#11 code. It was
  unused since we only support P-256 through aziot-keys.

- Remove disabled `arm32v7` and `aarch64` stubs from the GH checkin workflow.


daprilik (approved):
A lot of this has gone over my head, but the Rust bits seem good.

openssl-build/src/lib.rs
```diff
@@ -15,10 +15,6 @@ pub fn define_version_number_cfg() {
         if openssl_version >= 0x01_01_00_00_0 {
             println!("cargo:rustc-cfg=ossl110");
         }
-
```
massand, line 18: Any reason not to keep this? I thought all our crates were building fine against both versions of OpenSSL.

massand (approved):
Other than nit question, LGTM :shipit:

openssl-build/src/lib.rs
arsing: Removing this doesn't stop our code from working with 1.1.1. Removing it removes the way to determine whether the code was being compiled against 1.1.0 or 1.1.1. The only reason the code cared about that was that ed25519 is only available with 1.1.1

daprilik (approved):

PR #79, opened 2020-11-28
Use new rust-toolchain format which removes the need to install components manually.

arsing:
Also remove the checks around bindgen and cbindgen existing before
`cargo install`ing them. `cargo install` already acts as a no-op if the crate
is installed with the requested version. It even correctly reinstalls the crate
if the currently installed version doesn't match the requested version (both
for upgrading and downgrading), which the previous code did not handle.

### DIFF
--- a/ci/install-build-deps.sh
+++ b/ci/install-build-deps.sh
@@ -128,19 +128,19 @@ else
     curl -Lo ~/.cargo/bin/rustup 'https://static.rust-lang.org/rustup/dist/x86_64-unknown-linux-gnu/rustup-init'
     chmod +x ~/.cargo/bin/rustup
     hash -r
+    rustup self update # Create the hardlinks for cargo, rustc, etc
 fi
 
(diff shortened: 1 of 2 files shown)


daprilik (approved):

ci/install-build-deps.sh
```diff
@@ -128,19 +128,19 @@ else
     curl -Lo ~/.cargo/bin/rustup 'https://static.rust-lang.org/rustup/dist/x86_64-unknown-linux-gnu/rustup-init'
     chmod +x ~/.cargo/bin/rustup
     hash -r
+    rustup self update # Create the hardlinks for cargo, rustc, etc
```
massand, line 131: You could download `rustup` if it does not exist and then `rustup self update`, instead of checking if it does and repeating the `rustup self update`.

arsing: Done.

massand (approved):

PR #80, opened 2020-11-30
Fix how the TPM service appears in the docs ToC

micahl:
Noticed that in the Contents list for our [docs](https://azure.github.io/iot-identity-service/) the TPM page was listed as just "API".  This change should correct that.


daprilik (approved):
Good catch!

PR #81, opened 2020-12-02
Add proxy support to Identity Service

massand:
Changes:
- Added proxy client support
- Refactored Hub client to reduce redundancy

Verified using local squid proxy (Sas connector works only so far)

TODO:
[ ] Close open issue for setting TLS tunneling on `hyper_openssl` (X509 provisioning fails without setting tunneling connector)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -88,7 +88,7 @@ dependencies = [
  "aziot-keys-common",
  "aziot-tpmd",
  "backtrace",
- "base64",
+ "base64 0.12.3",
  "bytes",
  "derive_more",
  "nix",
@@ -134,7 +134,7 @@ dependencies = [
  "aziot-key-common",
  "aziot-key-common-http",
  "aziot-key-openssl-engine",
- "base64",
+ "base64 0.12.3",
  "foreign-types-shared",
  "futures-util",
  "hex",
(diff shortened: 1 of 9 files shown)


arsing:
This should be done in `http-common` or so, because CS also needs proxy support to connect to the EST server.

Also, you don't need to make an enum around two kinds of `hyper::Client`. An enum around two kinds of connector will be enough.

massand:
> This should be done in `http-common` or so, because CS also needs proxy support to connect to the EST server.

You are right. CS could reach out to an external CA source. We can move this up. However, the question of which Connector would work with ProxyConnector is still open. Any suggestions?

> Also, you don't need to make an enum around two kinds of `hyper::Client`. An enum around two kinds of connector will be enough.

This means a new hyper::Client is built on each `request()` within `MaybeProxyClient`. That should be fine though. I'll make a change.

arsing:
>This means a new hyper::Client is built on each `request()` within `MaybeProxyClient`. That should be fine though. I'll make a change.

I'm saying to not create `MaybeProxyClient` and instead add an `enum MaybeProxyConnector` in `http-common`. [1] Then wherever you were creating a `hyper::Client` you continue to do that, but with a `MaybeProxyConnector` instead of whatever connector you were creating it with.

[1]: You could make it one of the enum variants of the existing `http_common::Connector` but that'll probably be more annoying than it's worth, because proxies only make sense for clients.

PR #82, opened 2020-12-08
Host process authorization for Identity Service

massand:
- Added Unix domain socket peer credential support by switching to low-level connection-based hyper API
- Enabled authentication and authorization based on TOML configuration
- `/identities/identity` now returns module identity when requesting process is configured as `module` type in config.toml
- Changed configuration parsing logic to ensure there is only one principal allowed to act as a "daemon", i.e. identity type is empty in config.toml  

Note that the service still needs to be restarted for new configuration to take effect for now (separate PR will be made for dynamic configuration)

Verified using e2e-tests.sh with locally built package


cert/aziot-certd/Cargo.toml
```diff
@@ -16,6 +16,7 @@ http = "0.2"
 hyper = "0.13"
 hyper-openssl = "0.8"
 lazy_static = "1"
+libc = "0.2"
```
arsing, line 19: Doesn't seem to be used?

aziotd/src/main.rs
```diff
@@ -226,21 +228,50 @@ where
...
-        async move { Ok::<_, std::convert::Infallible>(server.clone()) }
-    }));
-    let () = server
-        .await
-        .map_err(|err| ErrorKind::Service(Box::new(err)))?;
 
-    log::info!("Stopped server.");
-
-    Ok(())
+    match &mut incoming {
```
arsing, line 236: If you keep this as a method of `http_common::Incoming` like:

```rust
async fn serve(&mut self, server: impl hyper::service::Service<...> + Clone + Send + 'static) -> std::io::Result<()> {
    match self { ... }
}
```

... (in place of the `hyper::Accept` impl that this replaces), you can avoid needing to make the `http_common::connector` and `::uid` modules `pub`.

key/aziot-keyd/Cargo.toml
```diff
@@ -12,6 +12,7 @@ futures-util = "0.3"
 http = "0.2"
 hyper = "0.13"
 lazy_static = "1"
+libc = "0.2"
```
arsing, line 15: Doesn't seem to be used?

tpm/aziot-tpmd/Cargo.toml
```diff
@@ -12,6 +12,7 @@ futures-util = "0.3"
 http = "0.2"
 hyper = "0.13"
 lazy_static = "1"
+libc = "0.2"
```
arsing, line 15: Doesn't seem to be used?

aziotd/src/main.rs
```diff
@@ -156,9 +156,11 @@ where
             hyper::Request<hyper::Body>,
             Response = hyper::Response<hyper::Body>,
             Error = std::convert::Infallible,
-        > + Clone
+        >
+        + Clone
         + Send
-        + 'static,
```
arsing, line 161: `Clone + Send + Sync + 'static` in that order, to be consistent.

Also, can you check if you really need the `Sync` bound? I'm surprised you do, given that `server` is essentially being used in the same way as before.

arsing (changes requested):

cert/aziot-certd/Cargo.toml
massand: I need to clean these up. Thanks.

aziotd/src/main.rs
massand: Didn't need it. I need to clean this up.

http-common/src/connector.rs
```diff
@@ -31,6 +31,52 @@ pub enum Incoming {
...
+                let (tcp_stream, _) = listener.accept().await?;
+
+                // TCP is available in test builds only (not production). Assume current user is root.
+                let server = crate::uid::UidService::new(0, server.clone());
+                tokio::task::spawn(async move {
+                    if let Err(http_err) = hyper::server::conn::Http::new()
+                        .serve_connection(tcp_stream, server)
+                        .await
+                    {
+                        eprintln!("Error while serving HTTP connection: {}", http_err);
```
arsing, line 57: `s/eprintln/log::info/`

arsing (changes requested):

ci/e2e-tests.sh
```diff
@@ -654,7 +654,7 @@ ssh -i "$PWD/vm-ssh-key" "aziot@$vm_public_ip" "
     printf 'Device twin: %s\n' \"\$device_twin\" >&2
 
     module_identity=\"\$(
-        curl --unix-socket '/run/aziot/identityd.sock' \\
+        sudo curl --unix-socket '/run/aziot/identityd.sock' \\
```
arsing, line 657: This was only done because `/identities/identity` returned the device identity. Since you've fixed it now, change this to invoke that API and you don't need `sudo`

massand: Good call. Done.

http-common/src/connector.rs
massand: `log::error` right?

arsing: It's not an error from the server's POV. It'll happen if say the client closes their connection in the middle of a request. After all, isn't that why you're logging and swallowing it instead of bubbling it up?

massand: Yes we shouldn't break the accept loop. Logging here is sufficient.

ci/e2e-tests.sh
```diff
@@ -655,7 +655,7 @@ ssh -i "$PWD/vm-ssh-key" "aziot@$vm_public_ip" "
 
     module_identity=\"\$(
         curl --unix-socket '/run/aziot/identityd.sock' \\
-            'http://foo/identities/modules/testmodule?api-version=2020-09-01&type=aziot'
+            'http://foo/identities/identity?api-version=2020-09-01'
```
arsing, line 658: Fix the doc comment in `iothub-get-twin.sh` too please (the example line 8).

arsing (approved):

ci/e2e-tests.sh
massand: Done

PR #83, opened 2020-12-08
Add `aziot check`

daprilik:
Adds a new `aziot check` subcommand to `aziot`, modeled after `iotedge check`.

Should be reviewed in conjunction with the upstream PR at https://github.com/Azure/iotedge/pull/4153

Please note that the `aziot check` implementation is _not_ a straight copy-paste from `iotedge check`, and while the two implementations have a lot in common, `aziot check`'s underlying check framework has been substantially refactored, simplified, and modernized. Notable changes include the use of `async_trait` on the `Check` trait, the use of `serde_erased` to simplify checker serialization, and removing all the Windows console specific text formatting code. 

While the internals of `aziot check` are quite different from `iotedge check`, I've made sure that the user-facing `text` and `json` output formats have remained exactly the same between the two tools.

One notable addition to `aziot check` is the new `json-stream` output format, which will enable `iotedge check` to invoke `iotedge check` as a sub-process. Aside from streamlining check flow (users won't have to invoke both `check` commands separately), the `additional_info` payload will be used to share certain bits of provisioning-specific information with `iotedge` that it wouldn't otherwise have access to (e.g: the `iothub_hostname` configuration option used in the host connectivity checks in `iotedge check` is part of the `identityd` config, which `iotedge` should _not_ be able to access directly).

* * *

This PR should be reviewed along two axes:

1. The `aziot check` subcommand, framework, and implementation quality
2. The actual checks themselves

The checks are divided into two categories:

1. Checks ported over from `iotedge check` (due to certain functionality being moved down the stack into `aziot`)
    - `host_connect_dps_endpoint`
    - `host_local_time`
    - `hostname`
    - `identity_certificate_expiry`
    - `well_formed_configs`
2. New checks specific to the new `aziot` architecture
    - `certs_preloaded`
    - `dameons_running`

While the primary goal of this PR is to get `aziot check` "up and running" in an effort to fix `iotedge check`, if there are any checks that I may have missed that are "critical" to merge as part of this initial PR, please let me know.

* * *

This PR will be un-drafted once the corresponding `iotedge check` PR has been submitted, as there's a non-zero chance the `json-streaming` output format will require changes as part of the integration work.


aziot/src/internal/check/checks/host_local_time.rs
```diff
@@ -0,0 +1,72 @@
...
+    }
+}
+
+impl HostLocalTime {
+    async fn execute_inner(
+        &mut self,
+        shared: &CheckerShared,
+        _cache: &mut CheckerCache,
+    ) -> Result<CheckResult> {
+        // TODO: check against parent time in nested edge scenarios
```
massand, line 32: IS doesn't have any knowledge of nested edge so far. Edge does have awareness of the parent_hostname configuration and hence nested edge.

aziot/src/internal/check/checks/host_local_time.rs
```diff
@@ -0,0 +1,72 @@
...
+
+        fn is_server_unreachable_error(err: &mini_sntp::Error) -> bool {
+            match err {
+                mini_sntp::Error::ResolveNtpPoolHostname(_) => true,
+                mini_sntp::Error::SendClientRequest(err)
+                | mini_sntp::Error::ReceiveServerResponse(err) => {
+                    err.kind() == std::io::ErrorKind::TimedOut || // Windows
```
massand, line 39: Can safely remove Windows support here. Unless this is needed by WSL?

aziot/src/internal/check/checks/mod.rs
```diff
@@ -0,0 +1,96 @@
...
+                identity_certificate_expiry::IdentityCertificateExpiry::default(),
+            ));
+            v.push(Box::new(certs_preloaded::CertsPreloaded::default()));
+            v
+        }),
+        ("Connectivity checks", {
+            let mut v: Vec<Box<dyn Checker>> = Vec::new();
+            v.extend(daemons_running::daemons_running());
+            v.push(Box::new(
+                host_connect_dps_endpoint::HostConnectDpsEndpoint::default(),
```
massand, line 91: Is iot hub connectivity checks in a different PR?

massand (commented):
LGTM so far. Few comments

aziot/src/internal/check/additional_info.rs
```diff
@@ -0,0 +1,192 @@
...
+use std::str;
+
+use byte_unit::{Byte, ByteUnit};
+use serde::Serialize;
+use sysinfo::{DiskExt, SystemExt};
+
+/// Additional info for the JSON output of `aziot check`
+#[derive(Clone, Debug, Serialize)]
+pub struct AdditionalInfo {
+    // TODO: update https://github.com/Azure/azure-iotedge to include aziot version
```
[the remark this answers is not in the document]
daprilik: I'm just assuming that it'd be a good idea to have a version-check test in `aziot` as well (mirroring https://github.com/Azure/iotedge/blob/release/1.0.10/edgelet/iotedge/src/check/checks/iotedged_version.rs)

aziot/src/internal/check/checks/host_local_time.rs
daprilik: Cool, so I can just remove this TODO then?

aziot/src/internal/check/checks/host_local_time.rs
daprilik: https://github.com/hyperium/hyper/issues/626#issuecomment-130754048

Since we aren't targeting windows anymore, sure, it'd be fine to remove this. At the same time, this single line allows `mini_sntp` to run on windows, should the need ever arise. Personally, I'd just leave it in...

aziot/src/internal/check/checks/mod.rs
daprilik: Are you talking about `host_connect_iothub`?

I was going to leave those in `iotedge check`, since fixing them is as easy as passing the `iothub_hostname` up from `aziot check` to `iotedge check` (through the `AdditionalInfo` payload).

That said, reviewing the [check's implementation](https://github.com/Azure/iotedge/blob/release/1.0.10/edgelet/iotedge/src/check/checks/host_connect_iothub.rs), there's no technical reason why it couldn't be pushed down into `aziot check`. Then again, does it make sense to perform AMQP and MQTT connectivity checks as part of the `aziot check`? My gut says no, not really.

Let me know what you think.

massand: >I was going to leave those in iotedge check, since fixing them is as easy as passing the iothub_hostname up from aziot check to iotedge check (through the AdditionalInfo payload).

We cannot assume `iotedge` exists on the same machine as `aziot`. The TCP connectivity check would be the same on both sides, but the protocols are different (HTTPS in `aziot check` vs AMQP/MQTT (+ WS) in `iotedge check`).

aziot/src/internal/check/checks/host_local_time.rs
massand: Sure we can leave it in

aziot/src/internal/check/checks/mod.rs
daprilik: Sounds good. In that case, I'll go ahead and add the HTTPS check into `aziot check`.

arsing: We do need AMQP/MQTT (+ WS) in `aziot check` too.

daprilik: Why?

arsing: The same reason `iotedge check` has them, even though `iotedged` doesn't use them either.

The point of `check` is to not just check the protocols our code uses, but what the user's code might use. That means host modules, which would use those protocols.

daprilik: But doesn't that reinforce the point @massand made? i.e: if the check is there to support a user's code, then shouldn't [some of] the the checks stay in iotedge?

An alternative argument might be that having two nearly identical checks targeting different protocols is a bit wasteful, and that we should just consolidate everything in either `aziot check` or `iotedge check` - likely the former.

arsing: I don't understand the confusion.

Host-connectivity checks are needed to troubleshoot host modules. Host modules are serviced by IS. The presence or absence of `iotedged` is irrelevant. Conclusion: `aziot check` needs host-to-upstream connectivity checks.

Container-connectivity checks are needed to troubleshoot Edge modules. Edge modules are serviced by `iotedged`. Conclusion: `iotedge check` needs container-to-upstream connectivity checks. Since it also delegates to `aziot check`, `iotedge check` does not need host-to-upstream connectivity checks.

daprilik: Oh, shoot, yes, that's right. I forgot that host modules are a thing (or at least, will be more of a thing).

Cool, in that case, I'll migrate the entire check into `aziot check`.

daprilik: Ooooh, now I remember why I wanted to leave it in `iotedge check`: the check doesn't has additional logic to ping the `parent_hostname` in nested-edge scenarios: https://github.com/Azure/iotedge/blob/release/1.2/edgelet/iotedge/src/check/checks/host_connect_upstream.rs#L66-L74

In any case, I have gone ahead and added a iothub connectivity check to `aziot check`, but what should we do about the nested-edge stuff?

aziot/src/internal/check/checks/host_local_time.rs
daprilik: Ahh, well, now I remember why I left this TODO... I accidentally liked to an older version of the check from before nested-edge. The latest version _does_ include some extra logic in nested-edge scenarios:
https://github.com/Azure/iotedge/blob/release/1.2/edgelet/iotedge/src/check/checks/host_local_time.rs#L41-L46

So, what should we do to account for that extra logic?

massand: The nested check seems strange because the root node of the nested edge topology would still need to be in sync with NTP (to connect to IoT Hub). All child nodes down the chain would then be transitively in sync with NTP. IS does not have the concept of `parent_hostname` because the parent hostname is the `iothub_hostname`. So, the NTP check is good enough for both cases in IS and you can remove the parent_hostname time check altogether.

daprilik: Awesome, that's what I like to hear :smile:

aziot/src/internal/check/checks/mod.rs
massand: As I said in the other comment, the `parent_hostname` check is part of the IS already via `aziot check` in the iothub connectivity check. I think if you add a check in `iotedge check` for `parent_hostname` matching the `iothub_hostname` in the IS. BTW, to get the `iothub_hostname`, I would read the `prev_state` file in the IS home folder, instead of the IS `config.toml` because that would be the single source of truth when DPS provisioning is supported for Nested Edge (currently Nested Edge only supports manual provisioning).

daprilik: Ahh, interesting... I like that approach a lot. 

As for how to obtain `iothub_hostname` - I'm currently including it as part of the `aziot check` `additional_info` JSON payload, which gets consumed by `iotedge` when it shells-out to `aziot check`. Let me know if you forsee any issues with that approach, though I think it's a pretty clean way to solving the "single source of truth" problem.

aziot/src/internal/check/checks/host_local_time.rs
arsing: We need both the NTP check and the HTTP Date header check in IS. If IS (and thus the check) doesn't know it's running in a nested scenario, the check might have to try NTP and then fall back to the HTTP Date header method.

@massand I don't understand what you mean by "all child nodes down the chain would then be transitively in sync with NTP."

aziot/aziot-check-common/src/lib.rs
```diff
@@ -0,0 +1,59 @@
...
+#[derive(Debug, Serialize, Deserialize)]
+pub struct CheckOutputSerializable {
+    pub result: CheckResultSerializable,
+    pub additional_info: serde_json::Value,
+}
+
+#[derive(Debug, Serialize, Deserialize)]
+#[serde(tag = "kind")]
+#[serde(rename_all = "snake_case")]
+pub enum CheckOuputSerializableStreaming {
```
massand, line 37: nit: Spelling for Output

massand (approved):
:shipit:

aziot/src/main.rs
```diff
@@ -7,48 +7,43 @@
     clippy::let_unit_value,
     clippy::similar_names,
     clippy::too_many_lines,
-    clippy::type_complexity
+    clippy::type_complexity,
+    clippy::clippy::module_name_repetitions
```
arsing, line 11: `s/clippy::clippy::/clippy::/`

And sort it.

aziot/src/main.rs
```diff
@@ -7,48 +7,43 @@
...
-            source = err.source();
-        }
-
-        writeln!(f, "{:?}", self.1)?;
-
-        Ok(())
+#[tokio::main]
+async fn main() {
+    if let Err(err) = try_main().await {
+        eprintln!("{:?}", err);
```
arsing, line 37: `"{}"`, no?

aziot/src/check.rs
```diff
@@ -0,0 +1,326 @@
...
+    CheckResultsSerializable,
+};
+
+use crate::internal::check::{
+    AdditionalInfo, CheckResult, CheckerCache, CheckerCfg, CheckerShared,
+};
+
+#[derive(StructOpt)]
+#[structopt(about = "Check for common config and deployment issues")]
+pub struct CheckCfg {
```
arsing, line 22: `CheckOptions`, to be consistent with the root `Options`.

Same for `CheckListCfg`

aziot/src/check.rs
```diff
@@ -0,0 +1,326 @@
...
+    #[structopt(
+        long,
+        value_name = "DONT_RUN",
+        value_delimiter = " ",
+        use_delimiter = true
+    )]
+    dont_run: Vec<String>,
+
+    /// Output format. Note that JSON output contains some additional information
+    /// like OS name, OS version, disk space, etc.
```
arsing, line 34: Add `One of "text" or "json".` between the two sentences?

(`"json-stream" omitted intentionally.)

aziot/src/check.rs
```diff
@@ -0,0 +1,326 @@
...
+                    },
+                    line
+                );
+            }
+        };
+    }
+
+    'all_checks: for (section_name, section_checks) in &mut check_data {
+        outputln!(normal, "{}", section_name);
+        outputln!(normal, "{}", "-".repeat(section_name.len()));
```
arsing, line 128: `outputln!(normal, "{:->section_name_len$}", "", section_name_len = section_name.len());`

... will do it without allocating a `String` like `str::repeat` does.

aziot/src/internal/check/checks/mod.rs
```diff
@@ -0,0 +1,98 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+use super::Checker;
+
+mod prelude {
+    pub use anyhow::{anyhow, Context, Error, Result};
+    pub use serde::Serialize;
+    pub use tokio::prelude::*;
```
arsing, line 8: Please no wildcard imports, even from preludes.

aziot/src/internal/check/checks/well_formed_configs.rs
```diff
@@ -0,0 +1,179 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+use super::prelude::*;
```
arsing, line 3: Please no wildcard imports, even from our own preludes.

aziot/src/internal/check/mod.rs
```diff
@@ -0,0 +1,126 @@
...
+}
+
+impl CheckerCache {
+    pub fn new() -> CheckerCache {
+        CheckerCache {
+            cfg: DaemonConfigs::default(),
+        }
+    }
+
+    /// Utility method to call `aziot_certd::get_path()` with the loaded certd config.
```
arsing, line 103: `s/aziot_certd/aziot_certd_config::util::/`

aziot/src/internal/check/mod.rs
```diff
@@ -0,0 +1,126 @@
...
+
+use std::path::PathBuf;
+
+use serde::Serialize;
+use structopt::StructOpt;
+
+mod additional_info;
+mod checks;
+
+pub use additional_info::AdditionalInfo;
```
arsing, line 11: `s/pub/pub(crate)/`

aziot/src/internal/common.rs
```diff
@@ -0,0 +1,111 @@
+// Copyright (c) Microsoft. All rights reserved.
+//! A grab-bag of misc. utilities shared across the various sub-commands.
```
arsing, line 2: Empty line between copyright header and the rest of the file.

identity/aziot-identityd-config/src/lib.rs
```diff
@@ -180,13 +183,9 @@ mod tests {
 
         match s.provisioning.provisioning {
```
arsing, line 184: I guess you could make these two `if !matches!(...) { panic!(...); }` to shorten them even more.

mini-sntp/src/lib.rs
```diff
@@ -0,0 +1,343 @@
...
+        .map_err(Error::SetWriteTimeoutOnSocket)?;
+
+    let mut num_retries_remaining = 3;
+    loop {
+        match query_inner(&socket, addr) {
+            Ok(result) => return Ok(result),
+            Err(err) => {
+                let is_retriable = match &err {
+                    Error::SendClientRequest(err) | Error::ReceiveServerResponse(err) => {
+                        err.kind() == std::io::ErrorKind::TimedOut || // Windows
```
arsing, line 55: Don't need this check any more.

arsing (changes requested):

aziot/src/main.rs
daprilik: Huh, I'm more surprised that `clippy::clippy::foo` not only validates, but has the same behavior as `clippy::foo`. That's weird

aziot/src/check.rs
daprilik: Ah, yep, very true. I carbon-copied this formatting from `iotedge check`, so I wasn't really paying much attention.

I mean, it doesn't _really_ matter in this particular case, since we're not min-maxing for memory use, but sure, I'll change it.

aziot/src/internal/check/checks/mod.rs
daprilik: Sure, I don't mind being more explicit about what tokio types get included here.

aziot/src/internal/check/checks/well_formed_configs.rs
daprilik: Uhh, I'm not sure I agree with you there. The whole point of the prelude is to wildcard import it, otherwise what's the point? If you're already importing types from the prelude manually, you might as well import the types using their fully-qualified paths to begin with.

Given that this prelude is private and can only be used by individual check implementations - all of which follow a rigid and matching module and import structure - I think using a prelude here is an appropriate way to cut down on noise in the implementations.

aziot/src/internal/check/mod.rs
daprilik: Uhh, sure, though it doesn't really matter since there is no top-level `lib.rs`, and all modules are marked private in `main.rs`. But we've had this pub/pub(crate) discussion before, and this is pretty inconsequential, so sure.

identity/aziot-identityd-config/src/lib.rs
daprilik: huh, I'm surprised `clippy` didn't catch that - it's usually pretty good about linting against this pattern.

mini-sntp/src/lib.rs
daprilik: https://github.com/Azure/iot-identity-service/pull/83#discussion_r539954457

aziot/src/internal/check/checks/well_formed_configs.rs
massand: +1 on using `tokio::prelude::*`. The tokio docs suggest manually importing it this way too. On a sidenote, tokio 1.0 no longer has the prelude module I believe - so this will eventually vanish?

aziot/src/internal/check/checks/mod.rs
daprilik: Good point from @massand, it looks like the tokio folks did indeed remove the prelude in 1.0 https://github.com/tokio-rs/tokio/pull/3299

arsing: Two problems with wildcard imports:

1. `pub use foo::*;` behaves weirdly when a `*`-imported ident conflicts with an explicitly defined one. Eg if tokio exported `Foo` and you also had `pub struct Foo`, then your crate would compile fine, but the crate that uses `this_crate::Foo` later gets an error.

1. It makes it impossible for a non-IDE user to know where a particular import is coming from.

daprilik: (I'm going to assume you meant to comment on https://github.com/Azure/iot-identity-service/pull/83#discussion_r552055404)

For point 1, I totally agree with you, hence why I don't like exporting preludes to external users.
For point 2, yeaaah, that's a more compelling argument.

That said, like I said in https://github.com/Azure/iot-identity-service/pull/83#discussion_r552030877, I'm not sure point 2 matters too much in this _particular_ case, since the prelude is a) only visible to submodules, and b) used across a set of submodules that all share a nearly identical import and code structure. IMO this private prelude actually improves the readability of the individual check implementations, since it puts more emphasis on the "meat" of the code, instead of the boilerplate corresponding to importing types like `CheckResult`' from `crate::internal::check` (or alternatively, `super::super::check`).

That said, there is an argument to be made about removing certain external crate re-exports from the prelude, as while it's fairly obvious what the `Serialize` identifier corresponds to, having `Result` correspond to `anyhow::Result` could be a bit confusing at first glance. I personally don't think so, but hey, this entire debate is about personal opinion :wink: 

While this isn't a hill I'm ready to die on, I do think we should keep this little private prelude (or at least some subset of it). What do other folks think about this, @massand and/or @gordonwang0?

aziot/Cargo.toml
```diff
@@ -6,21 +6,39 @@ edition = "2018"
...
+anyhow = "1.0.34"
+async-trait = "0.1.42"
 backtrace = "0.3"
 base64 = "0.12"
+byte-unit = "4.0.9"
+chrono = { version = "0.4", features = ["serde"] }
+colored = "2.0.0"
 derive_more = "0.99.11"
+erased-serde = "0.3.12"
+futures = "0.3"
```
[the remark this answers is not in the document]
arsing: `futures` is a re-export of `futures-core`, `futures-util`, and a bunch of other `futures-*` crates, and some features we don't care about enabled by default. You likely only need to depend on `futures-core` (for `Stream`) and `futures-util` (for `{Try,}{Future,Stream}Ext`), just like the other code in this repo. So I'm asking you to depend on `futures-core`, `futures-util` individually as required instead of the full `futures` crate.

aziot/src/main.rs
arsing: >And sort it.

aziot/src/main.rs
arsing: Does it print `caused by` chains the same way as other crates in the repo (and the checks in verbose mode) do it manually?

mini-sntp/src/lib.rs
arsing: I prefer to remove it. It's out-of-place in a repository that otherwise has no Windows support.

aziot/src/internal/check/checks/mod.rs
arsing: >(I'm going to assume you meant to comment on #83 (comment))

Point 1 only applies here; the other place wasn't a re-export. Point 2 applies to both.

>For point 1, I totally agree with you, hence why I don't like exporting preludes to external users.

The situation I described happens between modules too, not just between crates.

https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=cd3f3d15db79329b0fa6dc6379345588

Notice how the error is reported on the *use* of `Foo` in `m3::bar`, rather than on the `prelude` mod's re-exports that created the ambiguity in the first place. If you comment out the use of the ambiguous `Foo`, you'll see the code compiles fine, ie it's a trap waiting for the first person to start using `Foo`.

>That said, there is an argument to be made about removing certain external crate re-exports from the prelude, as while it's fairly obvious what the `Serialize` identifier corresponds to, having `Result` correspond to `anyhow::Result` could be a bit confusing at first glance.

It's not about confuson but about effort. It requires more effort to determine that `Error` is not imported from anywhere explicitly and thus is presumably coming from `prelude::*`, and then determine that `prelude` defines `Error` as a re-export of `anyhow::Error`. I'd rather it be made obvious that `Error` is `anyhow::Error` without having to jump to another file to get the answer.

Code is read more often than it is written. It helps when every place consistently imports well-known things. Eg I'm used to importing `Serialize` from `serde`, etc, so when searching for where `Serialize` is defined I automatically look for where `use serde::` ought to be in the file.

>IMO this private prelude actually improves the readability of the individual check implementations

Assuming the third-party imports are removed from the prelude, the remaining ones are all from `crate::internal::check`, plus one trait defined inline in the prelude. In the first place preludes don't define anything inline, so this ought to be moved somewhere, probably to `crate::internal::check`. That means all the imports are from `crate::internal::check`. So again, it's counter-productive to have to jump through hoops to figure out that `CheckResult` is not being imported explicitly and must thus be coming from `prelude::*`, and so on.

mini-sntp/src/lib.rs
daprilik: Are you sure? It's just one line, and it's not really a massive "maintenance burden" leaving it in. Plus, while the fix is ostensibly windows specific, there is no reason some other platform couldn't return the same error at some point (highly unlikely as it is).

aziot/src/internal/check/checks/mod.rs
daprilik: Huh, you know what, that's actually a pretty convincing argument you've got there. I think I might reconsider my use of preludes in my own code moving forwards...

So yeah, sure, I'll remove it.

aziot/src/main.rs
daprilik: Ah, yes, I was too quick on the trigger switching from `{:?}` to `{}`. See the corresponding `anyhow` docs: https://docs.rs/anyhow/1.0.37/anyhow/struct.Error.html#display-representations

I'll revert back to `{:?}`

aziot/src/main.rs
daprilik: huh, I kinda assumed `rustfmt` would sort it automatically, but no, it doesn't. I now see why you're annoyed by the fact that there's no trailing comma - spending 5 seconds fixing that build error is time I'll never get back lol

aziot/Cargo.toml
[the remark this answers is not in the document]
daprilik: Ah, yes, sure. Will do.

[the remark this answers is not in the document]
daprilik: Actually, funny enough, it seems that I never actually use `futures` anywhere in my code. I've removed it entirely.

I'll also do a quick sweet through the other deps to see if there's anything else that can be culled.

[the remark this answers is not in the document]
daprilik: huh, looks like we've got quite a few stragglers scattered around the repo. 

Running `cargo udeps` shows the following:

```
unused dependencies:
`aziot-dps-client-async v0.1.0 (/home/danielprilik/Work/iot-identity-service/identity/aziot-dps-client-async)`
└─── dependencies
     ├─── "aziot-key-client"
     └─── "aziot-key-openssl-engine"
`aziot-hub-client-async v0.1.0 (/home/danielprilik/Work/iot-identity-service/identity/aziot-hub-client-async)`
└─── dependencies
     ├─── "aziot-key-client"
     └─── "aziot-key-openssl-engine"
`aziot-identity-common-http v0.1.0 (/home/danielprilik/Work/iot-identity-service/identity/aziot-identity-common-http)`
└─── dependencies
     └─── "aziot-key-common-http"
`http-common v0.1.0 (/home/danielprilik/Work/iot-identity-service/http-common)`
└─── dependencies
     └─── "lazy_static"
```

Can I trim these as part of this PR, or should we leave it 'till later?

aziot/src/main.rs
arsing: But even that is not identical to the way the rest of the code prints cause-chains and backtraces, which was my point. anyhow's cause chain appears to use `Caused by:\n    ` as opposed to the rest of the code's `caused by: `, and its backtrace appears to use `Stack backtrace:` as opposed to `backtrace`'s `backtrace:`

It would be good to be consistent, ie either format anyhow's error manually, or change the rest of the code to match anyhow's style.

mini-sntp/src/lib.rs
arsing: If you want to keep it that's fine, but remove the `// Windows` and `// Unix` comments.

aziot/Cargo.toml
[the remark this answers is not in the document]
arsing: That's a nice tool. TIL

You can remove those now, but it might be nice to run `udeps` as part of `make test-release`, so if you want to do that you can remove these in that PR instead.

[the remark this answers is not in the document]
daprilik: The caveat is that `cargo udeps` require using a nightly Rust compiler, so it wouldn't be a trivial addition to the CI. We could totally add it, but idk if it's worth gunking up the pipelines even more by installing the nightly compiler alongside stable, and then also re-building the project with the nightly compiler just for `cargo udeps`.

In any case, I think i'll just cull these deps in this PR as well.

aziot/src/main.rs
daprilik: guh - I'm not sure if that's something I'd want to spend more time on.

I agree that consistency would be nice, but given that the `aziot` binary is totally distinct from `aziotd`, both in function and in its core error handling abstraction, I think it's fine leaving it as-is for now. If the inconsistency bothers you that much, feel free to tweak `aziotd`'s formatting in a separate PR.

aziot/src/internal/check/checks/mod.rs
daprilik: Yeah, I can't say I'm too keen on your spartan use of `use`.

My personal rule of thumb is that I'll use `use` when the type that's imported is unambiguous in that given context. e.g: I would never do something like `use std::collections::hash_map::IntoIter`, since `IntoIter` is far too ambigous. That said, something like `use std::sync::Arc` is totally fair game, since most of the time, `Arc` can only ever be one thing.

mini-sntp/src/lib.rs
daprilik: Cool, will do :+1:

aziot/src/main.rs
arsing: >feel free to tweak `aziotd`'s formatting in a separate PR.

Yes, that's fine.

aziot/src/internal/check/checks/host_connect_iothub.rs
```diff
@@ -0,0 +1,94 @@
...
+        use aziot_identityd_config::ProvisioningType;
+
+        let iothub_hostname = match &unwrap_or_skip!(&cache.cfg.identityd)
+            .provisioning
+            .provisioning
+        {
+            ProvisioningType::Manual {
+                iothub_hostname, ..
+            } => iothub_hostname,
+            _ => return Ok(CheckResult::Ignored),
```
arsing, line 79: The iotedge version of this check gets the hub name even for DPS provisioning if the device has been provisioned previously. Can you do that here? (Either read the backup file like the iotedge version does, or call out to IS, or perhaps Rajeev has a better idea.)

aziot/src/internal/check/checks/identity_certificate_expiry.rs
```diff
@@ -0,0 +1,81 @@
...
+                self.certificate_info = Some(cert_info.clone());
+                return cert_info.to_check_result();
+            }
+            ProvisioningType::Manual {
+                authentication: ManualAuthMethod::X509 { identity_cert, .. },
+                ..
+            } => {
+                self.provisioning_mode = Some("manual-x509");
+                let cert_info = CertificateValidity::new(
+                    unwrap_or_skip!(cache.cert_path(&identity_cert))?,
```
arsing, line 66: None of the `unwrap_or_skip!(cache.cert_path(cert_id))?`s are going to work if `cert_id` is a dynamically issued cert instead of a preloaded one. In that case `cert_path` returns `Some(Err)` instead of `None`

cert/aziot-certd-config/src/util.rs
```diff
@@ -0,0 +1,65 @@
...
+    if let Some(preloaded_cert) = preloaded_certs.get(cert_id) {
+        let path = get_preloaded_cert_path(preloaded_cert, cert_id)?;
+        return Ok(path);
+    }
+
+    let mut path = homedir_path.to_owned();
+    path.push("certs");
+
+    if !path.exists() {
+        let () = std::fs::create_dir_all(&path)?;
```
arsing, line 19: This should not be done when called from `aziot`. The permissions will be wrong.

aziot/src/internal/check/mod.rs
```diff
@@ -0,0 +1,127 @@
...
+pub(crate) use checks::all_checks;
+
+// NOTE: this struct gets `structopt(flatten)`ed as part of the `aziot check` subcommand.
+#[derive(StructOpt)]
+pub struct CheckerCfg {
+    // TODO: add aziot version info to https://github.com/Azure/azure-iotedge
+    // /// Sets the expected version of the iotedged binary. Defaults to the value
+    // /// contained in <http://aka.ms/latest-iotedge-stable>
+    // expected_iotedged_version: String,
+    //
```
arsing, line 22: This should be an empty line instead of `//` so that it doesn't connect to the next field's comment.

arsing (changes requested):

aziot/src/internal/check/mod.rs
daprilik: Alas, `rustfmt` strikes again, and will automatically remove the empty line. Of the various alternatives, I found adding this extra `//` the least offensive.

cert/aziot-certd-config/src/util.rs
daprilik: Wouldn't it just error due to not having permission to create the directory?

aziot/src/internal/check/checks/identity_certificate_expiry.rs
daprilik: I'm not sure I follow. Unless I'm missing something, the underlying `aziot_certd_config::util::get_path` function just hashes the cert_id if it's not preloaded, so it should still succeed.

aziot/src/internal/check/checks/host_connect_iothub.rs
daprilik: Calling out to IS isn't an option, since this check should work regardless of whether or not the IS is running.

What's this backup file you're referring to?

Also, pinging @massand for input

cert/aziot-certd-config/src/util.rs
daprilik: (if that isn't the case, then I guess I can slap on a `create_dir: bool` parameter and leave it at that)

arsing: `check` runs as root. You wouldn't be able to read the service configs otherwise.

daprilik: ah, very true. okay I feex

aziot/src/internal/check/checks/identity_certificate_expiry.rs
arsing: You're right. `get_path()` and `cache.cert_path()` will suceed. It's `CertificateValidity::new()` that will fail. So the point still stands.

aziot/src/internal/check/mod.rs
arsing: Yeesh.

aziot/src/internal/check/checks/identity_certificate_expiry.rs
daprilik: Wait, why? Wouldn't it just read the cert from disk?
Or am I misunderstanding how dynamically generated certs work?

aziot/src/internal/check/checks/host_connect_iothub.rs
arsing: >What's this backup file you're referring to?

https://github.com/Azure/iot-identity-service/blob/7e6c062988f26299fc4695d0df55a49e48b839cf/identity/aziot-identityd/src/lib.rs#L144-L145

aziot/src/internal/check/checks/identity_certificate_expiry.rs
arsing: The point of being dynamically generated is that the cert *isn't* already on disk. It's issued dynamically by CS when someone invokes its rest API.

arsing: Eg https://github.com/Azure/iot-identity-service/blob/7e6c062988f26299fc4695d0df55a49e48b839cf/identity/aziot-identityd/src/lib.rs#L689

aziot/src/internal/check/checks/host_connect_iothub.rs
daprilik: Oh, interesting...

> The iotedge version of this check gets the hub name even for DPS provisioning if the device has been provisioned previously

Wait a sec, unless I'm looking at the wrong code, _does_ it actually do that?
I'm looking at https://github.com/Azure/iotedge/blob/iiot/edgelet/iotedge/src/check/checks/host_connect_dps_endpoint.rs and/or https://github.com/Azure/iotedge/blob/iiot/edgelet/iotedge/src/check/checks/host_connect_upstream.rs

aziot/src/internal/check/checks/identity_certificate_expiry.rs
daprilik: ...okay, but if that method was previously called, then the cert would exist on disk, right?
Or is a brand new one generated on every run?

aziot/src/internal/check/checks/host_connect_iothub.rs
arsing: Those read `check.iothub_hostname`, not set it. It's set in `well_formed_connection_string.rs`, and the relevant code is https://github.com/Azure/iotedge/blob/iiot/edgelet/iotedge/src/check/checks/well_formed_connection_string.rs#L51-L66

daprilik: Ahhh, I see now. 

Alright, so you're suggesting reading + parsing the `prev_state` backup if `cache.cfg.identityd` isn't available? Yeah, I could do that.

aziot/src/internal/check/checks/identity_certificate_expiry.rs
arsing: >...okay, but if that method was previously called, then the cert would exist on disk, right?

If IS hasn't gotten around to provisioning yet for any reason, the device ID cert hasn't necessarily been created. You shouldn't assume that the user is running `aziot check` after IS has finished provisioning. If that was the assumption, we wouldn't need upstream connectivity checks.

aziot/src/internal/check/checks/host_connect_iothub.rs
arsing: >Alright, so you're suggesting reading + parsing the `prev_state` backup if `cache.cfg.identityd` isn't available? Yeah, I could do that.

No, they are different things. The procedure to use them maps one-to-one with the iotedged config and provisioning backup as in the code I linked.

If IS config doesn't exist at all, there's nothing you can do. Give up and move on to the other checks.

If IS config exists, and says it's manual provisioning, you know what the hub name is. The IS config will exist even if the provisioning backup doesn't exist, so for manual provisioning you only need to look at this one.

If IS config exists, and says it's DPS provisioning, the config doesn't tell you the hub name. In this case the only way to get it automatically is to read the provisioning backup. If the user didn't already override the hostname via `--iothub-hostname`, you fail.

aziot/src/internal/check/checks/identity_certificate_expiry.rs
daprilik: Okay, sure, but that doesn't answer my question about what the check should do if the cert is dynamically issued?

I understand things correctly, the current implementation will return a `CheckResult::Failed(<could not read file>)`, which doesn't seem totally unreasonable. One alternative could be to have a special case that checks if the cert is dynamically issued, and if the file doesn't exist in that case, report some alternative error (i.e: `no dynamically issued identity certificate has been issued`).

Any other ideas?

arsing: Treat it as as success or skipped. (That's what I meant in the first comment when I said it ought to behave like `None`.)

The check is primarily there to catch bad certs that the user created. Even if the dynamic configuration is broken for any reason (say the EST server always issued expired certs), there's no reason to pre-emptively believe a dynamically issued cert will be bad just because it hasn't been created yet. An error here is non-actionable by the user if the cert doesn't exist yet.

arsing: On that note, if the cert file isn't found, you should assert that the cert *is* configured to be dynamically issued according to the CS config, ie its ID is in `.cert_issuance.certs.keys()`.

aziot/src/internal/check/checks/host_connect_iothub.rs
daprilik: Okay, I pushed up a partial fix. The only thing that's missing is that additional logic regarding the `--iothub-hostname` flag, which doesn't exist in `aziot`. Let me know if I should implement it, or if it's fine to leave it out.

I'm signing off for the day, but I'll get on it tomorrow.

aziot/src/internal/check/checks/identity_certificate_expiry.rs
daprilik: Sounds good. I'll push the necessary changes tomorrow morning.

aziot/src/internal/check/checks/mod.rs
```diff
@@ -42,11 +42,11 @@ pub fn all_checks() -> Vec<(&'static str, Vec<Box<dyn Checker>>)> {
         }),
         ("Connectivity checks", {
             let mut v: Vec<Box<dyn Checker>> = Vec::new();
-            v.extend(daemons_running::daemons_running());
-            v.push(Box::new(
-                host_connect_dps_endpoint::HostConnectDpsEndpoint::default(),
-            ));
             v.extend(host_connect_iothub::host_connect_iothub_checks());
+            // v.extend(daemons_running::daemons_running());
```
arsing, line 46: Did you mean to do this change?

aziot/src/internal/check/checks/host_connect_iothub.rs
arsing: That's true. For `iotedge check` the flag came first, which is why we didn't remove it when we added the fallback to the backup file. But nobody remembered to set the flag anyway, even when the message telling them to set it was staring them in the face. So I'd be fine with not carrying it forward into 1.2

Check with Micah and Venkat, though.

aziot/src/internal/check/checks/mod.rs
daprilik: Nope, just fat-fingered it.

aziot/src/internal/check/checks/identity_certificate_expiry.rs
daprilik: Pushed up a fix. Simplified the code in the process, since it turned out my `cert_path` method was only called from this one check anyways.

aziot/src/internal/check/checks/host_connect_iothub.rs
daprilik: Alright, I've re-added the flag. I'll need to update the upstream PR to plumb through the parameter, but it shouldn't be too bad.

arsing (changes requested):

aziot/src/internal/check/checks/host_connect_iothub.rs
```diff
@@ -69,30 +69,49 @@ impl HostConnectIotHub {
...
+                        };
+
+                        if let Some(backup_hostname) = backup_hostname {
+                            backup_hostname
+                        } else {
+                            // the user never manually provisioned, nor have they passed
+                            // the `iothub-hostname` flag.
+                            let reason = "Could not retrieve iothub_hostname from provisioning file.\n\
+                            Please specify the backing IoT Hub name using --iothub-hostname switch if you have that information.\n\
+                            If no hostname is provided, all hub connectivity tests will be skipped.";
```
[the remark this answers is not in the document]
daprilik: Sure, though let it be known this text is a straight copypasta from the old repo

aziot/src/internal/check/checks/identity_certificate_expiry.rs
```diff
@@ -1,6 +1,6 @@
 // Copyright (c) Microsoft. All rights reserved.
```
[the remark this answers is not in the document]
daprilik: Just to clarify before I write the impls, these checks would have basically the same logic as this check (i.e: checking if the path exists, whether they are dynamically/manually issued, etc...), except the cert_ids being checked would be `certd_config.cert_issuance.est?.x509?.{identity,bootstrap_identity}.0` and `certd_config.cert_issuance.local_ca?.cert`?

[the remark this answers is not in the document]
arsing: Yes.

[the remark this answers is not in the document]
daprilik: Alright, I've added the checks.

All three checks are consolidated under `cert_expiry.rs`

aziot/src/internal/check/checks/cert_expiry.rs
```diff
@@ -0,0 +1,241 @@
...
+    if path.exists() {
+        let cert_info = CertificateValidity::new(path, cert_name, cert_id).await?;
+        cert_info
+            .to_check_result()
+            .map(|res| (res, Some(cert_info)))
+    } else if !certd_config.preloaded_certs.contains_key(cert_id)
+        && !certd_config.cert_issuance.certs.contains_key(cert_id)
+    {
+        Err(anyhow!(
+            "{} certificate is not manually specified, nor is it dynamically issued",
```
arsing, line 235: `"{} certificate is neither preloaded nor configured to be dynamically issued, and thus cannot be used."`

arsing (approved):

PR #85, opened 2020-12-16
Split out service config types into separate crates.

arsing:
This is needed to use the config types in `iotedge init --import`. As a bonus,
it also reduces the amount of code `aziot` needs to link to for `aziot init`.

Also change the tokio02 feature of http-common to default to off. Otherwise
it's very easy to write code that relies on it being on but not gate the code
behind a `#[cfg(feature)]`, and it's hard to make sure the feature is off
for good across all dependencies. This change also includes a couple of fixes
for the former.

Other changes:

- Fix missing mention of tpmd in aziot-init success message.

- Fix typos referring to the `dynamic_reprovisioning` config as
  `dynamic_provisioning`.

- Fix some cases of `Result` being unnecessarily referred to by
  its fully-qualified name.


daprilik (approved):
LGTM :+1:

Can't say I'm too excited to rebase #83 over these changes though... ah well :wink:

massand (approved):
:shipit:

PR #86, opened 2020-12-31
Update clippy suppression for new 1.49.0 release.

arsing:
Also fix some documentation that wasn't updated when tpmd was introduced.


daprilik (approved):
I haven't reviewed this PR in too much depth (I mean, look at this thing - it's massive!), but at first glance these changes seems alright. Might want to get another set of eyes to look at it though.

/s

PR #90, opened 2021-01-09
Update Rust dependencies to latest versions

massand:
Bumping major versions of dependencies after the recent release of tokio 1.0


arsing:
Lockfile still has tokio 0.2. You need to update whatever's using it too. (Search for `"tokio 0`)

massand:
> Lockfile still has tokio 0.2. You need to update whatever's using it too. (Search for `"tokio 0`)

Sorry I forgot to call this out in the PR description - I had lots of issues converting the `aziot-key-openssl-engine-shared-test` crate due to changes in some dependency crates in how they used to wrap `openssl` crate types, but now got rid of those wrappers (e.g. `HandshakeError`). We can independently do this (since it is a test only crate) and the last tokio 0.2 dependency will dissappear.

massand:
@arsing - Question about the failing check: have you hit licensing issues with `tracing-futures` crate in the past? It doesn't seem like a new dependency - maybe a new version?

arsing:
Take https://github.com/arsing/iot-identity-service/commits/tokioup . Everything is fixed in that.

arsing:
Rename the `tokio02` feature too please, to `tokio1`.

arsing (approved):

http-common/src/connector.rs
```diff
@@ -398,13 +395,32 @@ impl tokio::io::AsyncWrite for AsyncStream {
...
+    fn is_write_vectored(&self) -> bool {
+        match self {
+            AsyncStream::Tcp(inner) => inner.is_write_vectored(),
+            AsyncStream::Unix(inner) => inner.is_write_vectored(),
+        }
+    }
 }
 
-#[cfg(feature = "tokio02")]
+#[cfg(feature = "tokio1")]
```
arsing, line 418: You forgot to revert this. I'll send out a fix.

PR #92, opened 2021-01-11
Add settings file watcher for dynamic provisioning

massand:
- Add file watcher functionality to detect settings file change and trigger automatic identity reconciliation. Reconciliation is now triggered by 3 methods - `aziot-identityd` startup, `/identities/device/reprovision` API invocation, and settings file change triggered by file watcher
- Add `always_provision_on_startup` support in `[provisioning]` settings and remove `dynamic_provisioning` flag (which is relevant for Edge only)
- Refactor authorization policies to become truly static (i.e. not dependent on configuration) and rework authorization to simpler types of users
- Refactor configuration functions into common crate, which is accessible by `aziot-identityd` and `aziotd` crates
- Fix: Provisioning no longer blocks API server start up in `aziot-identityd`

How verified:
Manually packaged and ran the script as follows:
```
docker run -it --rm -v "$(realpath ~/iot-identity-service):/src" -e 'ARCH=amd64' -e 'PACKAGE_VERSION=1.1.0' -e 'PACKAGE_RELEASE=0' ubuntu:18.04 '/src/ci/package.sh'
./ci/e2e-tests.sh 'manual-symmetric-key'
```

TODO:

- [x] Figure out inclusion of third-party licensing for `inotify` crate (uses ISC which isn't allowed in `contrib\third-party-notices.sh` yet)
- [x] Run automated e2e test (blocked by dist package building due to third-party licensing error)

### DIFF
--- a/aziot/test-files/init/dps-symmetric-key/identityd.toml
+++ b/aziot/test-files/init/dps-symmetric-key/identityd.toml
@@ -2,7 +2,7 @@ hostname = "my-device"
 homedir = "/var/lib/aziot/identityd"
 
 [provisioning]
-dynamic_reprovisioning = true
+always_reprovisioning_on_startup = true
 source = "dps"
 global_endpoint = "https://global.azure-devices-provisioning.net"
 scope_id = "0ab1234C5D6"
--- a/aziot/test-files/init/dps-symmetric-key/identityd.toml
+++ b/aziot/test-files/init/dps-symmetric-key/identityd.toml
@@ -2,7 +2,7 @@ hostname = "my-device"
 homedir = "/var/lib/aziot/identityd"
 
 [provisioning]
-always_reprovisioning_on_startup = true
+always_reprovision_on_startup = true
 source = "dps"
 global_endpoint = "https://global.azure-devices-provisioning.net"
 scope_id = "0ab1234C5D6"
--- a/aziotd/Cargo.toml
+++ b/aziotd/Cargo.toml
@@ -18,4 +18,5 @@ aziot-certd = { path = "../cert/aziot-certd" }
 aziot-identityd = { path = "../identity/aziot-identityd" }
 aziot-keyd = { path = "../key/aziot-keyd" }
 aziot-tpmd = { path = "../tpm/aziot-tpmd" }
+config-common = { path = "../config-common" }
 http-common = { path = "../http-common", features = ["tokio1"] }
--- a/aziotd/Cargo.toml
+++ b/aziotd/Cargo.toml
@@ -12,7 +12,6 @@ hyper = "0.14"
 log = "0.4"
 serde = "1"
 tokio = { version = "1", features = ["macros", "rt-multi-thread"] }
-toml = "0.5"
 
 aziot-certd = { path = "../cert/aziot-certd" }
 aziot-identityd = { path = "../identity/aziot-identityd" }
--- a/aziotd/src/error.rs
+++ b/aziotd/src/error.rs
@@ -6,7 +6,7 @@ pub(crate) struct Error(pub(crate) ErrorKind, pub(crate) backtrace::Backtrace);
 #[derive(Debug)]
 pub(crate) enum ErrorKind {
     GetProcessName(std::borrow::Cow<'static, str>),
-    ReadConfig(Option<std::path::PathBuf>, Box<dyn std::error::Error>),
+    ReadConfig(config_common::error::Error),
     Service(Box<dyn std::error::Error>),
 }
 
@@ -14,10 +14,7 @@ impl std::fmt::Display for ErrorKind {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
             ErrorKind::GetProcessName(message) => write!(f, "could not read argv[0]: {}", message),
-            ErrorKind::ReadConfig(Some(path), _) => {
-                write!(f, "could not read config from {}", path.display())
-            }
-            ErrorKind::ReadConfig(None, _) => f.write_str("could not read config"),
+            ErrorKind::ReadConfig(_) => f.write_str("could not read config"),
             ErrorKind::Service(_) => f.write_str("service encountered an error"),
         }
     }
@@ -28,7 +25,7 @@ impl std::error::Error for ErrorKind {
         #[allow(clippy::match_same_arms)]
         match self {
             ErrorKind::GetProcessName(_) => None,
-            ErrorKind::ReadConfig(_, err) => Some(&**err),
+            ErrorKind::ReadConfig(_) => None,
             ErrorKind::Service(err) => Some(&**err),
         }
     }
--- a/aziotd/src/error.rs
+++ b/aziotd/src/error.rs
@@ -6,7 +6,7 @@ pub(crate) struct Error(pub(crate) ErrorKind, pub(crate) backtrace::Backtrace);
 #[derive(Debug)]
 pub(crate) enum ErrorKind {
     GetProcessName(std::borrow::Cow<'static, str>),
-    ReadConfig(config_common::error::Error),
+    ReadConfig(Box<dyn std::error::Error>),
     Service(Box<dyn std::error::Error>),
 }
 
@@ -25,7 +25,7 @@ impl std::error::Error for ErrorKind {
         #[allow(clippy::match_same_arms)]
         match self {
             ErrorKind::GetProcessName(_) => None,
-            ErrorKind::ReadConfig(_) => None,
+            ErrorKind::ReadConfig(err) => Some(&**err),
             ErrorKind::Service(err) => Some(&**err),
         }
     }
--- a/aziotd/src/main.rs
+++ b/aziotd/src/main.rs
@@ -173,9 +173,13 @@ where
     let config_directory_path: std::path::PathBuf = std::env::var_os(config_directory_env_var)
         .map_or_else(|| config_directory_default.into(), Into::into);
 
-    let config: TConfig = config_common::read_config(config_path.clone(), config_directory_path.clone()).map_err(|err| ErrorKind::ReadConfig(err))?;
+    let config: TConfig =
+        config_common::read_config(config_path.clone(), config_directory_path.clone())
+            .map_err(ErrorKind::ReadConfig)?;
 
-    let (connector, server) = main(config, config_path, config_directory_path).await.map_err(ErrorKind::Service)?;
+    let (connector, server) = main(config, config_path, config_directory_path)
+        .await
+        .map_err(ErrorKind::Service)?;
 
     log::info!("Starting server...");
 
--- a/aziotd/src/main.rs
+++ b/aziotd/src/main.rs
@@ -139,6 +139,19 @@ where
     }
 }
 
+/// Executes the main function of the running service.
+///
+/// The `main` function accepts 3 arguments:
+/// * TOML Configuration file for the running service that is created by reading 
+///   the base configuration TOML file and configuration override directory path that 
+///   contains patch TOML files (that should be merged with the base 
+///   configuration file).
+/// * Path to the base configuration file for the running service, if needed for 
+///   detecting configuration changes at run time.
+/// * Path to the configuration override directory for the running service, if needed for 
+///   detecting configuration changes at run time.
+///
+/// ```
 async fn run<TMain, TConfig, TFuture, TServer>(
     main: TMain,
     config_env_var: &str,
--- a/aziotd/src/main.rs
+++ b/aziotd/src/main.rs
@@ -142,25 +142,29 @@ where
 /// Executes the main function of the running service.
 ///
 /// The `main` function accepts 3 arguments:
-/// * TOML Configuration file for the running service that is created by reading 
-///   the base configuration TOML file and configuration override directory path that 
-///   contains patch TOML files (that should be merged with the base 
+///
+/// * TOML Configuration file for the running service that is created by reading
+///   the base configuration TOML file and configuration override directory path that
+///   contains patch TOML files (that should be merged with the base
 ///   configuration file).
-/// * Path to the base configuration file for the running service, if needed for 
-///   detecting configuration changes at run time.
-/// * Path to the configuration override directory for the running service, if needed for 
+///
+/// * Path to the base configuration file for the running service, if needed for
 ///   detecting configuration changes at run time.
 ///
-/// ```
-async fn run<TMain, TConfig, TFuture, TServer>(
-    main: TMain,
+/// * Path to the configuration override directory for the running service, if needed for
+///   detecting configuration changes at run time.
+async fn run<TConfig, TFuture, TServer>(
+    main: fn(
+        config: TConfig,
+        config_path: std::path::PathBuf,
+        config_directory_path: std::path::PathBuf,
+    ) -> TFuture,
     config_env_var: &str,
     config_file_default: &str,
     config_directory_env_var: &str,
     config_directory_default: &str,
 ) -> Result<(), Error>
 where
-    TMain: FnOnce(TConfig, std::path::PathBuf, std::path::PathBuf) -> TFuture,
     TConfig: serde::de::DeserializeOwned,
     TFuture: std::future::Future<
         Output = Result<(http_common::Connector, TServer), Box<dyn std::error::Error>>,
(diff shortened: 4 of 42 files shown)


aziot/test-files/init/dps-symmetric-key/identityd.toml
```diff
@@ -2,7 +2,7 @@ hostname = "my-device"
 homedir = "/var/lib/aziot/identityd"
 
 [provisioning]
-dynamic_reprovisioning = true
+always_reprovisioning_on_startup = true
```
arsing, line 5: `always_reprovision_on_startup`, not `always_reprovisioning_on_startup`

identity/aziot-identityd/Cargo.toml
```diff
@@ -15,15 +15,18 @@ chrono = "0.4"
 futures-util = "0.3"
 http = "0.2"
 hyper = "0.14"
+futures = "0.3"
```
arsing, line 18: Use only the individual `futures-*` crates that you need instead of the whole `futures` crate. You likely only need `futures-util` for the `{Try,}{Future,Stream}Ext` traits and `futures-core` for `Stream`.

identity/aziot-identityd/Cargo.toml
```diff
@@ -15,15 +15,18 @@ chrono = "0.4"
...
 libc = "0.2"
 log = "0.4"
+notify = "4"
 openssl = "0.10"
 percent-encoding = "2"
 regex = "1"
 serde = "1"
 serde_json = "1.0"
 toml = "0.5"
+tokio = { version = "1", features = ["macros", "rt"] }
```
arsing, line 29: You don't need `macros`

identity/aziot-identityd/src/auth/authorization.rs
```diff
@@ -31,7 +31,7 @@ impl Authorizer for DefaultAuthorizer {
     type Error = Error;
```
arsing, line 31: TODO on line 27 should also be removed?

config-common/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "config-common"
+version = "0.1.0"
+authors = ["Rajeev Massand <massand@gmail.com>"]
```
arsing, line 4: `authors = ["Azure IoT Edge Devs"]`

config-common/Cargo.toml
```diff
@@ -0,0 +1,12 @@
...
+edition = "2018"
+
+# See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
```
arsing, line 7: Remove this comment.

aziotd/Cargo.toml
```diff
@@ -18,4 +18,5 @@ aziot-certd = { path = "../cert/aziot-certd" }
 aziot-identityd = { path = "../identity/aziot-identityd" }
```
arsing, line 18: You don't need the `toml` dep any more, right?

aziotd/src/main.rs
```diff
@@ -147,7 +147,7 @@ async fn run<TMain, TConfig, TFuture, TServer>(
     config_directory_default: &str,
 ) -> Result<(), Error>
 where
-    TMain: FnOnce(TConfig) -> TFuture,
+    TMain: FnOnce(TConfig, std::path::PathBuf, std::path::PathBuf) -> TFuture,
```
arsing, line 150: `TMain: FnOnce(TConfig, &'_ std::path::Path, &'_ std::path::Path) -> TFuture`

Let the `TMain` convert them to owned if it needs to.

(If that doesn't compile because of "'_ not allowed", `TMain: for<'a> FnOnce(TConfig, &'a std::path::Path, &'a std::path::Path) -> TFuture` should.)

identity/aziot-identityd/src/lib.rs
```diff
@@ -47,105 +48,83 @@ macro_rules! match_id_type {
...
+    });
 
-        std::fs::write(prev_device_info_path, device_status)
-            .map_err(error::InternalError::SaveDeviceInfo)?;
-        let () = std::fs::write(prev_settings_path, &settings_serialized)
-            .map_err(error::InternalError::SaveSettings)?;
-    }
+    // Start file change listener that asynchronously updates Identity Service
+    tokio::task::spawn(async move {
+        loop {
```
arsing, line 112:
```rust
while let Some(()) = file_changed_rx.recv().await {
    let new_settings = ...
}
```

identity/aziot-identityd/src/lib.rs
```diff
@@ -47,105 +48,83 @@ macro_rules! match_id_type {
...
+            if event.is_some() {
+                let new_settings =
+                    config_common::read_config(config_path.clone(), config_directory_path.clone())
+                        .unwrap();
+
+                let mut api_ = api_watcher.lock().await;
+                let _ = api_
+                    .update_settings(new_settings, ReprovisionTrigger::ConfigurationFileUpdate)
+                    .await;
+                drop(api_);
```
arsing, line 124: `drop` is unnecessary here.

arsing (changes requested):

identity/aziot-identityd/src/identity.rs
```diff
@@ -407,4 +414,393 @@ impl IdentityManager {
...
+                    log::info!("Hub identity {:?} added", &m.0);
+                }
+
+                let () = std::fs::write(prev_device_info_path, device_status)
+                    .map_err(|err| Error::Internal(InternalError::SaveDeviceInfo(err)))?;
+
+                let () = std::fs::write(prev_settings_path, &settings_serialized)
+                    .map_err(|err| Error::Internal(InternalError::SaveSettings(err)))?;
+            }
+            None => log::warn!("reconcilation skipped since device is not provisioned"),
```
gordonwang0, line 777: I think this should be `log::info`, since there's nothing unusual about skipping Hub reconciliation for an unprovisioned device.

identity/aziot-identityd/src/lib.rs
```diff
@@ -429,192 +396,76 @@ impl Api {
...
-                };
-                self.id_manager.set_device(&device);
-                aziot_identity_common::ProvisioningStatus::Provisioned(device)
-            }
-            config::ProvisioningType::None => {
-                log::info!("Skipping provisioning with IoT Hub.");
+        log::info!("Provisioning starting.");
 
-                aziot_identity_common::ProvisioningStatus::Unprovisioned
+        let _ = match trigger {
```
gordonwang0, line 440: Log messages when settings change (with the reason) would be helpful while debugging.

aziotd/Cargo.toml
daprilik: Relevant utility: https://crates.io/crates/cargo-udeps

aziotd/src/main.rs
daprilik: On a related note, I'm not a huge fan of using common types like `Path` or `PathBuf` as part of a function pointer type, since without parameter names, it makes it hard to tell what types are supposed to represent.

Consider bundling these two paths into a struct with a more descriptive name, or if that's too high-effort, you can just add a doc-comment describing what the parameters represent.

identity/aziot-identityd/Cargo.toml
daprilik: Some food for thought: we already take a dependency on the entire `futures` crate as part of `tokio-openssl` (See our `Cargo.toml` [here](https://github.com/Azure/iot-identity-service/blob/1452c9b2d/Cargo.lock#L2100-L2110)). Moreover, given we have so many async ecosystem crates in our dependency graph, I'd be _very_ surprised if we aren't already implicitly bringing in all of the `futures-*` crates via transitive dependencies already.

If the goal is to avoid bringing in unused dependencies, this change won't really change much. That said, it is definately better to be precise precise about what subset of `futures` functionality we are using, so this fix does have merit.

aziotd/src/error.rs
```diff
@@ -28,7 +25,7 @@ impl std::error::Error for ErrorKind {
         #[allow(clippy::match_same_arms)]
         match self {
             ErrorKind::GetProcessName(_) => None,
-            ErrorKind::ReadConfig(_, err) => Some(&**err),
+            ErrorKind::ReadConfig(_) => None,
```
daprilik, line 28: It would be good to preserve the source of the underlying error message instead of just discarding it.

identity/aziot-identityd/src/configext.rs
```diff
@@ -14,26 +14,46 @@ pub fn load_file(filename: &Path) -> Result<config::Settings, InternalError> {
...
-        let info = match info.as_str() {
-            "unprovisioned" => None,
-            _ => Some(toml::from_str(&info).map_err(InternalError::ParseDeviceInfo)?),
-        };
-
-        Ok(info)
+pub fn prepare_authorized_principals(
+    principal: &[config::Principal],
+) -> (
+    std::collections::BTreeMap<config::Uid, config::Principal>,
```
daprilik, line 20: Could we clean this up a bit by importing `std::collections::{BTreeMap, BTreeSet}` at the top level? The fully qualified paths for those types are used >10 times throughout the function.

identity/aziot-identityd/src/lib.rs
```diff
@@ -47,105 +48,83 @@ macro_rules! match_id_type {
...
-    let api = Api::new(
-        settings,
-        local_modules,
-        authenticator,
-        authorizer,
-        allowed_users,
-    )?;
-    let api = Arc::new(futures_util::lock::Mutex::new(api));
+    let api_watcher = api.clone();
+    let (file_changed_tx, mut file_changed_rx) = tokio::sync::mpsc::channel(10);
```
daprilik, line 88: any significance to the number `10`, or did you just pick a number?
on that note, any reason this can't just be an `unbounded_channel` instead?

identity/aziot-identityd/Cargo.toml
arsing: Yes, I noticed that about `tokio-openssl` when doing the tokioup. It's doing it wrong. I'm planning to send it a PR when I have the time.

aziotd/src/main.rs
arsing: Could also replace `<TMain: FnOnce(...)>(main: TMain)` with `main: fn(config: TConfig, config_path: &Path, config_dir: &Path)`, since `fn` definitions allow parameter names. It's less general but all our callers use a `fn` anyway.

aziotd/Cargo.toml
massand: > While compilation of this tool also works on Rust stable, it needs Rust nightly to actually run.

Alas

daprilik: Yeah, while it wouldn't be feasible to integrate it into the CI (it would really bog things down having to install + recompile with nightly), it is a very useful utility to run _locally_ when you're doing these sorts of refactors.

arsing: CI won't be bogged down. It's just one extra full compile that needs to run in parallel with the 18 other CI tasks that do full compiles (out of 48 total tasks).

The problem is depending on a pinned nightly. We used to do that for cargo-grcov in the iotedge repo and it worked fine initially. Then it got old enough that it could no longer compile our code so it had to be bumped. Then it got old enough that the binaries stopped being hosted on rust-lang.org so rustup couldn't install it any more and had to be bumped. Then...

So we just gave up on the idea.

daprilik: Ah, yes, I forgot that if we were to add it to the CI, it would be as a separate action that isn't tied to the main pipeline (i.e: the big bash file).

I can see how depending on a pinned nightly would be annoying...
I mean, one hack could be to write a shell script that checks that stable version we are using, and then finds the closest nightly release. Just a thought.

arsing: Unpinned nightly is probably worse than pinned nightly re: reliability.

daprilik: What I'm proposing is a quasi-pinned nightly, that gets automatically re-pinned to whatever nightly is closest to the stable version we are using. It'd be somewhere in the middle of unpinned and pinned.

arsing: The comment where I mentioned the case of Azure/iotedge was about how periodically updating the pinned nightly was a pain in the ass that nobody wanted to do, and that was just when we absolutely had to because the alternative was a broken build. Your suggestion involves updating the nightly even more often.

daprilik: Right, but I assume the whole reason it didn't work was because it required _manual_ updates. What I'm saying is that we could write the Action where the prereq script _automatically_ tracks the appropriate nightly version without any manual intevention required.

arsing: Automatically updated nightly is an unpinned nightly.

config-common/Cargo.toml
massand: Fixed

identity/aziot-identityd/src/auth/authorization.rs
massand: Fixed

identity/aziot-identityd/Cargo.toml
massand: Removed

identity/aziot-identityd/Cargo.toml
massand: Removed

aziot/test-files/init/dps-symmetric-key/identityd.toml
massand: Fixed

identity/aziot-identityd/src/lib.rs
massand: Fixed

identity/aziot-identityd/src/lib.rs
massand: Fixed

identity/aziot-identityd/src/lib.rs
massand: IIRC the watcher can only work with a bounded one

identity/aziot-identityd/src/lib.rs
```diff
@@ -47,105 +48,80 @@ macro_rules! match_id_type {
...
-        } else {
-            Ok(auth::AuthId::Unknown)
-        }
-    });
+    let settings_copy = settings.clone();
 
-    let authorizer = Box::new(SettingsAuthorizer {
-        allowed_users: allowed_users.clone(),
-        modules: hub_modules.clone(),
+    tokio::task::spawn(async move {
```
arsing, line 79: `tokio::spawn` (It's a re-export of `tokio::task::spawn` so there's no reason to use the full name.)

identity/aziot-identityd/src/lib.rs
arsing: You've put the `while` inside the `loop`. The `while` was meant to replace the `loop`.

arsing (changes requested):

identity/aziot-identityd/src/lib.rs
arsing: The watcher directly doesn't make any difference, but the point is that only the regular `Sender` has `blocking_send`. The `UnboundedSender` doesn't.

But anyway, I don't see a reason to use `10` either. `1` should be fine - there's only one sender and it's doing a blocking send, so it's fine if it blocks anyway.

identity/aziot-identityd/src/configext.rs
massand: Fixed

identity/aziot-identityd/src/lib.rs
massand: The main thread gets blocked if the message count is set to 1. Added a comment on why this value is being set to 2 for now. And I am certain we will need to tweak with this number at some point to optimize after testing.

identity/aziot-identityd/src/lib.rs
massand: Fixed here and `http-common` crate as well.

identity/aziot-identityd/src/lib.rs
massand: Fixed

identity/aziot-identityd/src/identity.rs
massand: Fixed

identity/aziot-identityd/src/identity.rs
```diff
@@ -407,4 +414,393 @@ impl IdentityManager {
             secondary_key,
         ))
     }
+
+    pub async fn provision_device(
+        &mut self,
+        provisioning: aziot_identityd_config::Provisioning,
```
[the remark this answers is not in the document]
massand: Fixed

identity/aziot-identityd/src/lib.rs
massand: Oops. Fixed.

aziotd/src/main.rs
massand: Thanks for the patch @arsing. Fixed.

aziotd/src/error.rs
massand: Fixed

gordonwang0 (approved):

arsing (approved):

PR #94, opened 2021-01-20
aziot: tweak `check-list` flags to be consistent with `check`

daprilik:
In response to https://github.com/Azure/iotedge/pull/4153#discussion_r555611080

### DIFF
--- a/aziot/src/check_list.rs
+++ b/aziot/src/check_list.rs
@@ -8,15 +8,33 @@ use aziot_check_common::CheckListOutput;
 #[derive(StructOpt, Copy, Clone)]
 #[structopt(about = "List the checks that are run for 'aziot check'")]
 pub struct CheckListOptions {
-    /// Output available checks as a JSON object.
-    #[structopt(short, long)]
-    json: bool,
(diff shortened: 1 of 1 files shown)


aziot/src/check_list.rs
```diff
@@ -8,15 +8,33 @@ use aziot_check_common::CheckListOutput;
...
+            _ => return Err("invalid output format"),
+        })
+    }
 }
 
 pub fn check_list(cfg: CheckListOptions) -> Result<()> {
     let checks = crate::internal::check::all_checks();
 
-    if cfg.json {
+    if matches!(cfg.output, OutputFormat::Json) {
```
arsing, line 37: Make this a `match` so that it's exhaustive.

daprilik: b-b-but think of the extra layer of nesting!

arsing: I do. Exhaustiveness is nice.

daprilik: fixed.

arsing (approved):

PR #97, opened 2021-01-25
[aziot check] Skip tpmd connectivity check if it's not being used

daprilik:
One small point of discussion: should this connectivity check be run if it isn't possible to determine the provisioning mode? If not, then line 114 needs to be changed from `true` to `false`.

### DIFF
--- a/aziot/src/internal/check/checks/daemons_running.rs
+++ b/aziot/src/internal/check/checks/daemons_running.rs
@@ -84,8 +84,41 @@ impl Checker for DaemonRunningTpmd {
         }
     }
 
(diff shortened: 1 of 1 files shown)


huguesBouvier (approved):
LGTM

arsing:
It doesn't matter. Based on today's discussion with Varun and Micah we're going to change this to use `systemctl status` on the socket units instead. (That also has the advantage of not waking up the service like the current impl does.)

arsing (approved):

arsing:
@huguesBouvier Please let the bot or the PR author merge the PR in the future.

PR #98, opened 2021-01-28
Decorating error message for nested configuration

huguesBouvier:
Adding to the error message for nested

### DIFF
--- a/aziot/src/internal/common.rs
+++ b/aziot/src/internal/common.rs
@@ -94,7 +94,7 @@ pub async fn resolve_and_tls_handshake(endpoint: hyper::Uri, hostname_display: &
     let _ = hyper_openssl::HttpsConnector::new()
         .with_context(|| {
             anyhow!(
-                "Could not connect to {} : could not create TLS connector",
+                "Could not connect to {} : could not create TLS connector. In nested configuration make sure root certificate and parent root cerificate are identical",
                 hostname_display,
             )
         })?


aziot/src/internal/common.rs
```diff
@@ -94,7 +94,7 @@ pub async fn resolve_and_tls_handshake(endpoint: hyper::Uri, hostname_display: &
     let _ = hyper_openssl::HttpsConnector::new()
         .with_context(|| {
             anyhow!(
-                "Could not connect to {} : could not create TLS connector",
+                "Could not connect to {} : could not create TLS connector. In nested configuration make sure root certificate and parent root cerificate are identical",
```
arsing, line 97:
```
If this is a nested Edge configuration, make sure the root certificate and the parent root certificate are identical.
```

(`cerificate` -> `certificate`, punctuation, grammar.)

---

Also, this message is going to be confusing for someone who a) isn't running IoT Edge in a nested Edge configuration, or b) isn't running IoT Edge at all. Is there a way to know for sure whether the user is running in a nested Edge configuration? Maybe based on the IoT Hub hostname in IS config? ( @daprilik , @massand , @gordonwang0 )

arsing (changes requested):

aziot/src/internal/common.rs
arsing: Also also, is this "(parent) root certificate" error supposed to be here, such that it's reported for both DPS and Hub checks? Or is it only supposed to be for hub checks? If the latter, it should be done in the caller of this function in `host_connect_iothub.rs`, not here.

massand: This PR comes a tad early. The parent vs service hostname split has not happened yet in IS, which is required for nested topology awareness. That would be a separate check and this error message should not be here.

huguesBouvier: "Also, this message is going to be confusing for someone who a) isn't running IoT Edge in a nested Edge configuration" I agree with you. However, unless I am wrong,  the only way to know we are in nested configuration is the parent_hostname which is not accessible from here.

huguesBouvier: "If the latter, it should be done in the caller of this function in host_connect_iothub.rs, not here." Agreed

huguesBouvier: "This PR comes a tad early. The parent vs service hostname split has not happened yet in IS, which is required for nested topology awareness. That would be a separate check and this error message should not be here." -> Gotcha! That will be much better. That's what I was originally trying to do. Please keep me posted, I'll remake that PR once I can

huguesBouvier (commented):
Closing for now. Will reopen once service vs parenthostname split has happened

huguesBouvier (commented):
Re-opening this PR. There is an idea I want to try:
If iotHub name end with .azure-devices.net we know we high certainty the device is not nested.

PR #99, opened 2021-02-01
Add checks for Y2038 and Y2050 certificate issues

daprilik:


arsing:
But why? This was a problem with libiothsm-std. Are you sure it still exists with certd's code too?

daprilik:
Indeed it seems you're right. I just tried making an identity certificate that expires in ~20 years, and it seems to work just fine.

Thankfully, it didn't take too long to write this check, so no harm no foul.

That said, maybe it makes sense to leave the Y2038 check in place, but downgrade it to a warning? While our code seems to handle this case correctly, the user might be playing with fire using these types of certs on a system with a 32-bit time_t. Just an idea - I don't have enough hands on experience with the issue to determine whether or not that'd be worthwhile.

As for the Y2050 check, it seems like it should be fine to remove it entirely, right?

arsing:
>While our code seems to handle this case correctly, the user might be playing with fire using these types of certs on a system with a 32-bit time_t.

I don't think it would be useful. The problem is not with our code, the warning wouldn't be actionable by the user (ie there's nothing they can do to silence the warning), and most importantly the factor that caused the issue (parsing X.509 times with `time_t` instead of a better type) is not common enough in other software that we should pre-emptively big-brother against it.

daprilik:
Sounds reasonable. In that case, I'm closing this PR.

kpm-at-hfi:
A Y2038 warning would have been extremely useful for me actually implementing... I wasn't aware there were goblins lurking there.

PR #102, opened 2021-02-05
[aziotctl check] ignore (instead of skip) tpmd check

daprilik:
Skipped implies that the test didn't run due to a failure, which isn't the case here.

### DIFF
--- a/aziotctl/src/internal/check/checks/daemons_running.rs
+++ b/aziotctl/src/internal/check/checks/daemons_running.rs
@@ -117,7 +117,7 @@ impl Checker for DaemonRunningTpmd {
         };
 
         if !using_tpmd {
-            return CheckResult::Skipped;
+            return CheckResult::Ignored;
         }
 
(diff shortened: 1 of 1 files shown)


arsing (approved):

aziotctl/src/internal/check/checks/daemons_running.rs
```diff
@@ -117,7 +117,7 @@ impl Checker for DaemonRunningTpmd {
         };
 
         if !using_tpmd {
-            return CheckResult::Skipped;
+            return CheckResult::Ignored;
```
massand, line 120: Not related to this PR, but I am curious why the TPM service status is being checked if the provisioning information is missing in either current or previous configuration. Should we bother checking tpmd status if `aziot init` was not run?

daprilik: See discussion in https://github.com/Azure/iot-identity-service/pull/97

PR #104, opened 2021-02-09
Allow runtime reconfiguration of KS and CS

gordonwang0:
- Moves Identity Service config file watcher code to config-common.
- Shares watcher code with CS and KS so that they can update their configs at runtime. This will be needed to support runtime authorization policy updates.

I updated CS to also update cert_issuance and preloaded_certs at runtime. This will allow reconfiguration of EST and other cert issuance methods. I didn't update KS to do anything because any updates to its preloaded keys can just be done by importing the key.

### DIFF
--- a/cert/aziot-certd/src/lib.rs
+++ b/cert/aziot-certd/src/lib.rs
@@ -143,8 +143,8 @@ impl UpdateConfig for Api {
     ) -> Result<(), Self::Error> {
         log::info!("Updating config due to {:?}.", trigger);
 
-        // Don't allow changes to homedir path while daemon is running. Only update
-        // cert issuance method and preloaded certs.
+        // Don't allow changes to homedir path or endpoints while daemon is running.
+        // Only update other fields.
         self.cert_issuance = new_config.cert_issuance;
         self.preloaded_certs = new_config.preloaded_certs;
 
--- a/cert/aziot-certd/src/lib.rs
+++ b/cert/aziot-certd/src/lib.rs
@@ -24,7 +24,7 @@ use aziot_certd_config::{
 };
 
 use async_trait::async_trait;
-use config_common::watcher::{UpdateConfig, ReprovisionTrigger};
+use config_common::watcher::{ReprovisionTrigger, UpdateConfig};
 
 pub async fn main(
     config: Config,
--- a/cert/aziot-certd/src/lib.rs
+++ b/cert/aziot-certd/src/lib.rs
@@ -23,10 +23,13 @@ use aziot_certd_config::{
     EstAuthX509, LocalCa, PreloadedCert,
 };
 
+use async_trait::async_trait;
+use config_common::watcher::{UpdateConfig, ReprovisionTrigger};
+
 pub async fn main(
     config: Config,
-    _: std::path::PathBuf,
-    _: std::path::PathBuf,
+    config_path: std::path::PathBuf,
+    config_directory_path: std::path::PathBuf,
 ) -> Result<(http_common::Connector, http::Service), Box<dyn std::error::Error>> {
     let Config {
         homedir_path,
@@ -62,6 +65,8 @@ pub async fn main(
         }
     };
     let api = std::sync::Arc::new(futures_util::lock::Mutex::new(api));
+
+    config_common::watcher::start_watcher(config_path, config_directory_path, api.clone());
 
     let service = http::Service { api };
 
@@ -123,6 +128,28 @@ impl Api {
             Err(ref err) if err.kind() == std::io::ErrorKind::NotFound => Ok(()),
             Err(err) => Err(Error::Internal(InternalError::DeleteFile(err))),
         }
+    }
+}
+
+#[async_trait]
+impl UpdateConfig for Api {
+    type Config = Config;
+    type Error = Error;
+
+    async fn update_config(
+        &mut self,
+        new_config: Self::Config,
+        trigger: ReprovisionTrigger,
+    ) -> Result<(), Self::Error> {
+        log::info!("Updating config due to {:?}.", trigger);
+
+        // Don't allow changes to homedir path while daemon is running. Only update
+        // cert issuance method and preloaded certs.
+        self.cert_issuance = new_config.cert_issuance;
+        self.preloaded_certs = new_config.preloaded_certs;
+
+        log::info!("Config update finished.");
+        Ok(())
     }
 }
 
--- a/cert/aziot-certd/src/lib.rs
+++ b/cert/aziot-certd/src/lib.rs
@@ -11,6 +11,8 @@
     clippy::too_many_lines
 )]
 
+use async_trait::async_trait;
+
 mod error;
 use error::{Error, InternalError};
 
@@ -23,8 +25,7 @@ use aziot_certd_config::{
     EstAuthX509, LocalCa, PreloadedCert,
 };
 
-use async_trait::async_trait;
-use config_common::watcher::{ReprovisionTrigger, UpdateConfig};
+use config_common::watcher::UpdateConfig;
 
 pub async fn main(
     config: Config,
@@ -136,17 +137,19 @@ impl UpdateConfig for Api {
     type Config = Config;
     type Error = Error;
 
-    async fn update_config(
-        &mut self,
-        new_config: Self::Config,
-        trigger: ReprovisionTrigger,
-    ) -> Result<(), Self::Error> {
-        log::info!("Updating config due to {:?}.", trigger);
+    async fn update_config(&mut self, new_config: Self::Config) -> Result<(), Self::Error> {
+        log::info!("Detected change in config files. Updating config.");
 
         // Don't allow changes to homedir path or endpoints while daemon is running.
         // Only update other fields.
-        self.cert_issuance = new_config.cert_issuance;
-        self.preloaded_certs = new_config.preloaded_certs;
+        let Config {
+            homedir_path: _,
+            cert_issuance,
+            preloaded_certs,
+            endpoints: _,
+        } = new_config;
+        self.cert_issuance = cert_issuance;
+        self.preloaded_certs = preloaded_certs;
 
         log::info!("Config update finished.");
         Ok(())
--- a/config-common/Cargo.toml
+++ b/config-common/Cargo.toml
@@ -11,5 +11,5 @@ backtrace = "0.3"
 futures-util = "0.3"
 notify = "4"
 serde = "1"
-tokio = { version = "1", features = ["rt"] }
+tokio = { version = "1", features = ["rt", "sync"] }
 toml = "0.5"
--- a/config-common/Cargo.toml
+++ b/config-common/Cargo.toml
@@ -6,6 +6,10 @@ edition = "2018"
 
 
 [dependencies]
+async-trait = "0.1"
 backtrace = "0.3"
+futures-util = "0.3"
+notify = "4"
 serde = "1"
+tokio = { version = "1", features = ["rt"] }
 toml = "0.5"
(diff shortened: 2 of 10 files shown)


config-common/src/watcher.rs
```diff
@@ -0,0 +1,81 @@
...
+use async_trait::async_trait;
+use notify::Watcher;
+use std::path::PathBuf;
+
+type Api<TConfig, TError> = std::sync::Arc<
+    futures_util::lock::Mutex<dyn UpdateConfig<Config = TConfig, Error = TError> + Send>,
+>;
+
+#[derive(Debug)]
+pub enum ReprovisionTrigger {
```
arsing, line 12: 1. Only IS considers these to be "reprovision triggers". The other services have no concept of provisioning.

1. `Api` and `Startup` are only used by IS. `Api` in particular only makes sense for IS.

1. Most importantly, the file watcher is the only caller of `update_config` except for one case where it's called in IS startup.

I suggest keeping this type IS-specific and removing it from `update_config`. For IS, you can have another function that does the config update and have it take a `ReprovisionTrigger`, and call it from IS's `update_config` impl with `ReprovisionTrigger::ConfigFileUpdate`.

key/aziot-keyd/src/lib.rs
```diff
@@ -18,10 +18,13 @@ mod http;
 
 use aziot_keyd_config::{Config, Endpoints};
 
+use async_trait::async_trait;
```
arsing, line 21: Same issue here re: order of imports.

config-common/src/watcher.rs
```diff
@@ -0,0 +1,81 @@
...
+    type Error: std::error::Error;
+
+    async fn update_config(
+        &mut self,
+        new_config: Self::Config,
+        trigger: ReprovisionTrigger,
+    ) -> Result<(), Self::Error>;
+}
+
+pub fn start_watcher<TConfig: 'static, TError: 'static>(
```
arsing, line 30: Lifetime bounds should go along with the other bounds, on lines 35 and 36.

config-common/Cargo.toml
```diff
@@ -6,6 +6,10 @@ edition = "2018"
 
 
 [dependencies]
+async-trait = "0.1"
 backtrace = "0.3"
+futures-util = "0.3"
+notify = "4"
 serde = "1"
+tokio = { version = "1", features = ["rt"] }
```
arsing, line 14: Also need the `"sync"` feature for `tokio::sync::mpsc`

config-common/src/watcher.rs
```diff
@@ -0,0 +1,81 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+use async_trait::async_trait;
```
arsing, line 3: Same issue here re: order of imports.

config-common/src/watcher.rs
```diff
@@ -0,0 +1,81 @@
...
+use notify::Watcher;
+use std::path::PathBuf;
+
+type Api<TConfig, TError> = std::sync::Arc<
+    futures_util::lock::Mutex<dyn UpdateConfig<Config = TConfig, Error = TError> + Send>,
```
arsing, line 8: Why does this need to be a trait object? The user of this, `start_watcher` doesn't need it to be one. `start_watcher` is already generic anyway, on `TConfig` and `TError`.

```rust
fn start_watcher<TApi>(...)
where
    TApi: UpdateConfig,
    <TApi as UpdateConfig>::Config: serde::de::DeserializeOwned + Send,
    <TApi as UpdateConfig>::Error: std::error::Error,
{
```

arsing (changes requested):

cert/aziot-certd/src/lib.rs
```diff
@@ -23,10 +23,13 @@ use aziot_certd_config::{
     EstAuthX509, LocalCa, PreloadedCert,
 };
 
+use async_trait::async_trait;
```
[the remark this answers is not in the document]
daprilik: As it happens, `rustfmt` recently (?) gained unstable support for enforcing this automatically:

https://github.com/rust-lang/rustfmt/blob/master/Configurations.md#group_imports

As with most things rustfmt related, it isn't _exactly_ what we'd like (it doesn't handle the crate.io vs. workspace crate grouping), but if it ever hits stable it might be worth enabling.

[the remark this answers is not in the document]
daprilik: In-fact, I decided to open an issue on the `rustfmt` repo that references this exact issue:

https://github.com/rust-lang/rustfmt/issues/4693

config-common/Cargo.toml
daprilik: wait, then how did this compile in the first place?

config-common/src/watcher.rs
daprilik: Looks like `UpdateConfig`'s trait definition already includes the `serde::de::DeserializeOwned` and `std::error::Error` bounds, so I'm pretty sure you won't need those two extra bounds on `start_watcher`.

Oh, but `UpdateConfig::::Config` doesn't have a `Send` bound in the trait definition...
IMO it's probably fine to add a `Send` bound to the trait definition in order to simplify `start_watcher`'s trait bounds. After all, I can't imagine a situation where TConfig _wouldn't_ be `Send`.

config-common/src/watcher.rs
```diff
@@ -0,0 +1,81 @@
...
+            let (file_watcher_tx, file_watcher_rx) = std::sync::mpsc::channel();
+
+            // Create a watcher object, delivering debounced events.
+            let mut file_watcher =
+                notify::watcher(file_watcher_tx, std::time::Duration::from_secs(10)).unwrap();
+
+            // Add configuration paths to be watched.
+            file_watcher
+                .watch(config_directory_path, notify::RecursiveMode::Recursive)
+                .unwrap();
```
daprilik, line 58: Consider changing these `unwrap`s to `expect`s with an appropriate error message.

identity/aziot-identityd/src/lib.rs
```diff
@@ -24,8 +24,9 @@ pub mod error;
 mod http;
 pub mod identity;
 
+use async_trait::async_trait;
+use config_common::watcher::{ReprovisionTrigger, UpdateConfig};
```
daprilik, line 28: nit: import order / grouping

cert/aziot-certd/src/lib.rs
```diff
@@ -126,6 +131,28 @@ impl Api {
...
+    async fn update_config(
+        &mut self,
+        new_config: Self::Config,
+        trigger: ReprovisionTrigger,
+    ) -> Result<(), Self::Error> {
+        log::info!("Updating config due to {:?}.", trigger);
+
+        // Don't allow changes to homedir path or endpoints while daemon is running.
+        // Only update other fields.
+        self.cert_issuance = new_config.cert_issuance;
```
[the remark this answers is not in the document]
gordonwang0: Fixed

cert/aziot-certd/src/lib.rs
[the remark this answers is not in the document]
gordonwang0: Moved async_trait to the top

key/aziot-keyd/src/lib.rs
```diff
@@ -444,6 +449,25 @@ impl Api {
...
+    async fn update_config(
+        &mut self,
+        _new_config: Self::Config,
+        trigger: ReprovisionTrigger,
+    ) -> Result<(), Self::Error> {
+        log::info!("Updating config due to {:?}.", trigger);
+
+        // TODO: implement update to authorization.
+
+        log::info!("Config update finished.");
```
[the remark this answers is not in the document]
gordonwang0: See the todo, I haven't implemented the update yet. Commented out logs for now.

config-common/src/watcher.rs
gordonwang0: Fixed

key/aziot-keyd/src/lib.rs
gordonwang0: Fixed

config-common/src/watcher.rs
gordonwang0: Fixed

config-common/Cargo.toml
gordonwang0: I was able to compile without "sync"

config-common/src/watcher.rs
gordonwang0: Fixed

config-common/src/watcher.rs
gordonwang0: Fixed

config-common/src/watcher.rs
gordonwang0: Fixed

identity/aziot-identityd/src/lib.rs
gordonwang0: Fixed

config-common/Cargo.toml
arsing: Features get unioned across the workspaces, and something else is enabling it.

I didn't say it wouldn't compile. I said it's needed by the code. Hence it needs to be listed.

key/aziot-keyd/src/lib.rs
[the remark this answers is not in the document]
arsing: If you're not doing it as part of this PR, just delete the entire implementation and the call to `start_watcher` in `main`

config-common/src/watcher.rs
arsing: It hasn't been fixed? All I see is you've inlined the typedef into the function definition. It's still a trait object.

config-common/src/watcher.rs
```diff
@@ -55,10 +43,10 @@ pub fn start_watcher<TConfig: 'static, TError: 'static>(
             // Add configuration paths to be watched.
             file_watcher
```
arsing, line 44: This needs a check for whether the `config_directory_path` exists. It may not. Eg `read_config` treats `std::io::ErrorKind::NotFound` as success when discovering config files.

config-common/src/watcher.rs
```diff
@@ -55,10 +43,10 @@ pub fn start_watcher<TConfig: 'static, TError: 'static>(
...
                 .watch(config_directory_path, notify::RecursiveMode::Recursive)
```
arsing, line 45: This should be `NonRecursive`.

https://docs.rs/notify/4.0.15/notify/trait.Watcher.html#tymethod.watch

>If the `path` is a directory, `recursive_mode` will be evaluated. If `recursive_mode` is `RecursiveMode::Recursive` events will be delivered for all files in that tree. Otherwise only the directory and its immediate children will be watched.

We only look for config files in the immediate children.

arsing (changes requested):

key/aziot-keyd/src/lib.rs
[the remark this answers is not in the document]
gordonwang0: Done

config-common/Cargo.toml
gordonwang0: Fixed

config-common/src/watcher.rs
gordonwang0: Sorry, misread what you wrote the first time.  Fixed.

config-common/src/watcher.rs
gordonwang0: Fixed.

config-common/src/watcher.rs
gordonwang0: Fixed.

arsing (approved):

PR #105, opened 2021-02-09
System commands

lfitchett:
Add a 'system' command to aziotctl with sub-commands: restart, logs, status, and set-log-level.

It is architected in such a way to make re-use in iotedge simple. All functionality is in the aztiot_common folder, and a list of services is passed in.


aziotctl/aziotctl-common/src/lib.rs
```diff
@@ -57,3 +67,24 @@ pub struct CheckerMetaSerializable {
 
 /// Keys are section names
 pub type CheckListOutput = BTreeMap<String, Vec<CheckerMetaSerializable>>;
+
+pub struct ServiceDefinition {
+    pub service: &'static str,
+    pub sockets: &'static [&'static str],
+}
+
+// Note, the ordering is important, since the first service is considered the root and will be started.
```
arsing, line 76: Append "by the `start` subcommand" to this comment perhaps, so it's clearer.

aziotctl/aziotctl-common/src/set_log_level.rs
```diff
@@ -0,0 +1,68 @@
...
+            "debug" => LogLevel::Debug,
+            _ => return Err("invalid log level"),
+        })
+    }
+}
+
+impl fmt::Display for LogLevel {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            LogLevel::Normal => write!(f, "warn"),
```
lfitchett, line 64: I wasn't sure what to put here. Is warn the default or "normal" log level? Would it be better to just delete the log file to set the "normal" value?

aziotctl/aziotctl-common/src/restart.rs
```diff
@@ -0,0 +1,59 @@
...
+    for ServiceDefinition {
+        service,
+        sockets: _,
+    } in services
+    {
+        print!("Stopping {}...", service);
+        let result = Command::new("systemctl")
+            .args(&["stop", service])
+            .output()
+            .unwrap();
```
arsing, line 17: Bubble up the errors using `.with_context(|| format!("could not stop {}", ...))` instead of `.unwrap()` (`use anyhow::Context;` . See the rest of the aziotctl code for examples.)

Same for the `status.success()` check.

aziotctl/aziotctl-common/src/restart.rs
```diff
@@ -0,0 +1,59 @@
...
+        service: _,
+        sockets,
+    } in services
+    {
+        for socket in sockets.iter() {
+            start(socket);
+        }
+    }
+
+    // start the first service. This service should be the one that will use the other services
```
arsing, line 40:
```diff
-// start the first service. This service should be the one that will use the other services
+// Start the first service. This is the primary service that should be enabled and started.
+// Other services will be started automatically by the primary service via socket activation as necessary.
```

aziotctl/aziotctl-common/src/set_log_level.rs
```diff
@@ -0,0 +1,68 @@
...
+use std::fs::File;
+use std::io;
+use std::io::prelude::*;
+use std::process::Command;
+use std::str::FromStr;
+
+use crate::ServiceDefinition;
+
+#[allow(clippy::module_name_repetitions)]
+/// # Errors
```
arsing, line 11: Don't bother with the `# Errors` docs. Just `#![allow]` the clippy lint that told you to add them. (See `CONTRIBUTING.md` for the list of clippy lints it's okay to allow in general, and why this lint ought to be allowed in particular.)

aziotctl/aziotctl-common/src/set_log_level.rs
```diff
@@ -0,0 +1,68 @@
...
+    for ServiceDefinition {
+        service,
+        sockets: _,
+    } in services
+    {
+        write_log(service, level)?;
+    }
+    Command::new("systemctl").arg("daemon-reload").output()?;
+
+    println!("Set log level to {} for all services. Run the `system restart` command for the changes to take effect.");
```
arsing: "Run the `system restart` command" will make the user think they need to run `system restart` rather than `aziotctl system restart` I assume you made this generic because it'll have to work for `iotedge` too? Jet get `argv[0]` for that.

```rust
let program_name =
    std::env::args_os().next()
    .and_then(|arg| arg.into_string().ok())
    .unwrap_or_else(|| "<unknown>".to_owned());
```

Same for the others like the mention of "`system logs`" in the `system status` code below.

aziotctl/aziotctl-common/src/set_log_level.rs
```diff
@@ -0,0 +1,68 @@
...
+    Ok(())
+}
+
+fn write_log(service: &str, level: LogLevel) -> Result<(), io::Error> {
```
arsing, line 29: `write_log_level` (`write_log` sounds like a logging function.)

aziotctl/aziotctl-common/src/set_log_level.rs
```diff
@@ -0,0 +1,68 @@
...
+    let filename = format!("/etc/systemd/system/{}.d/log-level.conf", service);
```
arsing, line 30: You need to create the `/etc/systemd/system/{}.d` directory. It won't exist by default.

aziotctl/aziotctl-common/src/set_log_level.rs
```diff
@@ -0,0 +1,68 @@
...
+
+    let mut file = File::create(filename)?;
+    file.write_all(contents.as_bytes())?;
+
+    Ok(())
+}
+
+#[derive(Clone, Copy, Debug)]
+pub enum LogLevel {
+    Debug,
```
arsing, line 45: There's no `Normal = "normal"`. It's `Info = "info"`.

Also, add `Error = "error"`, `Off = "off"`, `Trace = "trace"` and `Warn = "warn"` for completeness.

(Full list is https://docs.rs/log/0.4.14/src/log/lib.rs.html#362 )

aziotctl/aziotctl-common/src/set_log_level.rs
arsing, line 64: `write!(f, "...")` can be `f.write_str("...")` which saves on the `format_args!` machinery.

aziotctl/aziotctl-common/src/set_log_level.rs
```diff
@@ -0,0 +1,68 @@
+use std::fmt;
+use std::fs::File;
+use std::io;
+use std::io::prelude::*;
+use std::process::Command;
+use std::str::FromStr;
+
+use crate::ServiceDefinition;
+
+#[allow(clippy::module_name_repetitions)]
```
arsing, line 10: Disallow this globally too, just like all the other crates in this repo.

aziotctl/aziotctl-common/src/status.rs
```diff
@@ -0,0 +1,125 @@
...
+        .output()
+        .unwrap();
+
+    let output = String::from_utf8_lossy(&result.stdout);
+
+    match output.trim() {
+        "active" => Status::Active,
+        "failed" => Status::Failed,
+        "inactive" => Status::Inactive,
+        &_ => {
```
arsing, line 62: `&_` should be just `_`

aziotctl/aziotctl-common/src/status.rs
```diff
@@ -0,0 +1,125 @@
...
+    if results.iter().any(|s| !s.ok()) {
+        for result in &results {
+            println!("{}: {}", result.service_name, result.service_status);
+            for socket in &result.sockets {
+                println!("{}: {}", socket.socket_name, socket.socket_status);
+            }
+            println!();
+        }
+
+        for result in &results {
```
arsing, line 33: You can consume `results` here instead of borrowing it.

arsing, line 33: Is printing the logs here part of Micah's spec?

aziotctl/aziotctl-common/src/status.rs
```diff
@@ -0,0 +1,125 @@
...
+        .unwrap();
+
+    let output = String::from_utf8_lossy(&result.stdout);
+
+    match output.trim() {
+        "active" => Status::Active,
+        "failed" => Status::Failed,
+        "inactive" => Status::Inactive,
+        &_ => {
+            println!("\nError calling `systemctl is-active {}`.", process);
```
arsing, line 63: There are a whole bunch of states, so I'd a) rather not treat an unexpected state as an "error calling systemctl is-active" (calling it went fine, and probably emitted valid output), and b) rather not emit a line for this that breaks up the rest of the output.

The easiest way is to make `Status::Failed` be `Status::Failed(String)` and make it hold the original string returned by `systemctl is-active`. So you silently treat it as "failed", but still have the original string to show in the status output.

aziotctl/aziotctl-common/src/system_logs.rs
```diff
@@ -0,0 +1,17 @@
+use std::process::Command;
+
+#[allow(clippy::module_name_repetitions)]
+pub fn get_system_logs(processes: &[&str], mut additional_args: &[&str]) {
+    let processes = processes.iter().flat_map(|p| vec!["-u", p]);
```
arsing, line 5: `&["-u", p]` instead of allocating a new `Vec` for each iterator item. Plus a `.copied()` at the end to convert the `Iterator<Item = &&str>` to `Iterator<Item = &str>` as `Command::args` wants.

aziotctl/aziotctl-common/src/system_logs.rs
```diff
@@ -0,0 +1,17 @@
+use std::process::Command;
+
+#[allow(clippy::module_name_repetitions)]
+pub fn get_system_logs(processes: &[&str], mut additional_args: &[&str]) {
```
arsing, line 4: `additional_args` should be `&[&OsStr]` in case the user gave non-utf-8 args (and it would be used with `Command::args_os` instead of `Command::args`).

aziotctl/src/main.rs
```diff
@@ -48,4 +50,6 @@ enum Options {
 
     /// List the checks that are run for 'aziotctl check'
     CheckList(check_list::CheckListOptions),
+    ///Use system helper commands
```
arsing, line 53: Missing empty line between the enum variants.

aziotctl/src/system.rs
```diff
@@ -0,0 +1,62 @@
...
+#[derive(StructOpt)]
+#[structopt(about = "Get logs for the services")]
+pub struct LogsOptions {
+    /// Extra args to be passed to journalctl
+    #[structopt(last = true)]
+    args: Vec<String>,
+}
+
+#[derive(StructOpt)]
+#[structopt(about = "Set the log level")]
```
arsing, line 33: "Set the log level of the services."

aziotctl/src/system.rs
```diff
@@ -0,0 +1,62 @@
...
+pub struct LogLevelOptions {
+    #[structopt(value_name = "normal | debug")]
```
arsing, line 35: `value_name = r#"One of "normal" or "debug"."#`

(See `aziotctl check`'s `--output` parameter.)

arsing (changes requested):

aziotctl/aziotctl-common/src/restart.rs
daprilik: This is in `aziotctl-common`, which doesn't currently depend on `anyhow`.

I'm all for adding it as a dependency, since `anyhow` is pretty slick, and plays nicely with `std::error::Error`. Alternatively, we could just have a simple module-specific error enum.

In either case, `unwrap` 🙅

aziotctl/aziotctl-common/src/restart.rs
```diff
@@ -0,0 +1,59 @@
...
+            .args(&["stop", service])
+            .output()
+            .unwrap();
+
+        if result.status.success() {
+            println!("Stopped!");
+        } else {
+            println!("\nError stopping {}", service);
+            io::stdout().write_all(&result.stdout).unwrap();
+            io::stderr().write_all(&result.stderr).unwrap();
```
daprilik, line 24: It's fine to leave these unwraps, given that `println!` does pretty much the same thing under the hood anyways.

aziotctl/aziotctl-common/src/restart.rs
```diff
@@ -0,0 +1,59 @@
...
+        print!("Stopping {}...", service);
+        let result = Command::new("systemctl")
+            .args(&["stop", service])
+            .output()
+            .unwrap();
+
+        if result.status.success() {
+            println!("Stopped!");
+        } else {
+            println!("\nError stopping {}", service);
```
daprilik, line 22: suggestions: switch this to `eprintln!`, and write out the `result.stdout` and `result.stderr` to `stdout`. You might also want to add a `eprintln!` between lines 23 and 24 that breaks-up the process's stdout and stderr.

aziotctl/aziotctl-common/src/restart.rs
```diff
@@ -0,0 +1,59 @@
...
+    print!("Starting {}...", name);
+    let result = Command::new("systemctl")
+        .args(&["start", name])
+        .output()
+        .unwrap();
+
+    if result.status.success() {
+        println!("Started!");
+    } else {
+        println!("\nError starting {}", name);
```
daprilik, line 54: ditto the comments from `restart`

aziotctl/aziotctl-common/src/restart.rs
```diff
@@ -0,0 +1,59 @@
...
+use std::process::Command;
+
+use crate::ServiceDefinition;
+
+pub fn restart(services: &[&ServiceDefinition]) {
+    // stop all services
+    for ServiceDefinition {
+        service,
+        sockets: _,
+    } in services
```
daprilik, line 11: This multi-line destructure inside a for loop looks a little... weird?

It might be a bit more verbose, but something like this seems easier to grok at a glance:

```rust
for service in services.iter().map(|s| s.service)
```

You used this in a whole bunch of places, so if you decide to change it, make sure to fix the other ones as well.

aziotctl/aziotctl-common/src/set_log_level.rs
```diff
@@ -0,0 +1,68 @@
...
+    for ServiceDefinition {
+        service,
+        sockets: _,
+    } in services
+    {
+        write_log(service, level)?;
+    }
+    Command::new("systemctl").arg("daemon-reload").output()?;
+
+    println!("Set log level to {} for all services. Run the `system restart` command for the changes to take effect.", level);
```
daprilik, line 25: question: what `system restart` command is this referring to?

aziotctl/aziotctl-common/src/set_log_level.rs
```diff
@@ -0,0 +1,68 @@
...
+fn write_log(service: &str, level: LogLevel) -> Result<(), io::Error> {
+    let filename = format!("/etc/systemd/system/{}.d/log-level.conf", service);
+    let contents = format!(
+        "[Service]
+Environment=AZIOT_LOG={}",
+        level
+    );
+
+    let mut file = File::create(filename)?;
+    file.write_all(contents.as_bytes())?;
```
daprilik, line 38: https://doc.rust-lang.org/std/fs/fn.write.html

aziotctl/aziotctl-common/src/set_log_level.rs
daprilik: ehhhh, I'd just stick with `write!(f, "...")`. This isn't on a hot path, and we don't use `f.write_str` anywhere else.

(This point is moot if we switch over the using `log::Level` directly)

aziotctl/aziotctl-common/src/set_log_level.rs
daprilik: Actually, it might be better to add `log` as a dependency, and use the `log::Level` type directly.

It already implements `FromStr` and `fmt::Display`, and is guaranteed to stay in-sync with the values used by the daemons.

aziotctl/aziotctl-common/src/status.rs
```diff
@@ -0,0 +1,126 @@
...
+                print_logs(result.service_name);
+            }
+            for socket in &result.sockets {
+                if !socket.ok() {
+                    print_logs(socket.socket_name);
+                }
+            }
+        }
+
+        println!("\n\nFor more detailed logs, use the `system logs` command. If the logs do not contain enough information, consider setting debug logs using `system set-log-level`.");
```
daprilik, line 44: nit: I'd swap out the `\n`s for calls to `println!()`.
Super inconsequential change, but IMO it makes it easier to grok what the output will look like.

aziotctl/aziotctl-common/src/status.rs
```diff
@@ -0,0 +1,126 @@
...
+
+    match output.trim() {
+        "active" => Status::Active,
+        "failed" => Status::Failed,
+        "inactive" => Status::Inactive,
+        &_ => {
+            println!("\nError calling `systemctl is-active {}`.", process);
+            println!(
+                "{}\n{}",
+                String::from_utf8_lossy(&result.stdout),
```
daprilik, line 66: I would keep things consistent with the code you had in `aziotctl/aziotctl-common/src/restart.rs` (i.e: writing the raw bytes out over stderr/stdout)

aziotctl/src/main.rs
daprilik: and missing a space between `///` and `Use`

aziotctl/src/system.rs
daprilik: this will have to be changed if you end up using `log::Level` directly.

aziotctl/src/system.rs
```diff
@@ -0,0 +1,62 @@
...
+        }
+        SystemOptions::Status(_) => {
+            get_status(SERVICE_DEFINITIONS);
+            Ok(())
+        }
+        SystemOptions::Logs(opts) => {
+            logs(&opts);
+            Ok(())
+        }
+        SystemOptions::SetLogLevel(opts) => Ok(set_log_level(SERVICE_DEFINITIONS, opts.log_level)?),
```
daprilik, line 53: Simplify `Ok(foo()?)` to just `foo()`. The types will line up once you switch over to using `anyhow::Result` in `set_log_level`.

daprilik (changes requested):
Mostly minor nits, but one big thing you're missing is support for `tpmd`. 

You'll need to add some logic that checks the current provisioning configuration, and conditionally includes tpmd in the set of services to check. Unfortunately, this means that your current `pub const SERVICE_DEFINITIONS` approach will need to be changed, as the set of services is not constant at compile time, and will change depending on the user's configuration.

Check out the `daemons_running` check for an example of how to read the provisioning configuration + decide whether or not `tpmd` should be used.

aziotctl/aziotctl-common/src/status.rs
lfitchett: yep

aziotctl/aziotctl-common/src/system_logs.rs
lfitchett: I used a vec because using a slice gives:
```
error[E0515]: cannot return reference to temporary value
 --> aziotctl/aziotctl-common/src/system_logs.rs:7:51
  |
7 |     let processes = processes.iter().flat_map(|p| &["-u", p]);
  |                                                   ^---------
  |                                                   ||
  |                                                   |temporary value created here
  |                                                   returns a reference to data owned by the current function

error: aborting due to previous error
```

aziotctl/aziotctl-common/src/restart.rs
lfitchett: its really just the auto-formatter that makes it look weird. I agree its weird split on multiple lines. single line it reads fine. Since we always use the auto-formatter, I'll switch to map.

```
for ServiceDefinition {service, sockets: _,} in services {

}
```

aziotctl/aziotctl-common/src/set_log_level.rs
lfitchett: added the name of the current tool to description. I left it out, since this will be used by both aziotctl and iotedge. Now I jsut print argv[0]

aziotctl/aziotctl-common/src/lib.rs
```diff
@@ -2,11 +2,31 @@
...
+mod set_log_level;
+mod status;
+mod system_logs;
+
+pub use restart::restart;
+pub use set_log_level::set_log_level;
+pub use status::get_status;
+pub use system_logs::get_system_logs;
+
+pub use log::Level as LogLevel;
```
daprilik, line 28: ehhhh, I'd avoid introducing this typedef, since it's re-exporting a foreign type under a different name. 

If you want to "hide" the fact that the underlying type of `LogLevel` is a `log::Level`, then you could introduce a `struct LogLevel(log::Level)` type, but it would add a bunch of boilerplate and is probably overkill.

aziotctl/aziotctl-common/src/lib.rs
```diff
@@ -57,3 +77,36 @@ pub struct CheckerMetaSerializable {
...
+        service: "aziot-certd.service",
+        sockets: &["aziot-certd.socket"],
+    },
+    &ServiceDefinition {
+        service: "aziot-tpmd.service",
+        sockets: &["aziot-tpmd.socket"],
+    },
+];
+
+#[allow(clippy::must_use_candidate)]
```
daprilik, line 106: https://github.com/Azure/iot-identity-service/blob/main/CONTRIBUTING.md

See 3.vii

aziotctl/aziotctl-common/src/restart.rs
```diff
@@ -0,0 +1,55 @@
+use std::io::{self, Write};
+use std::process::Command;
+
+use anyhow::{Context, Result};
+
+use crate::ServiceDefinition;
+
+#[allow(clippy::missing_errors_doc)]
```
daprilik, line 8: https://github.com/Azure/iot-identity-service/blob/main/CONTRIBUTING.md

See 3.vi

ditto for the other instances

arsing, line 8: Like I said last time:

>Just `#![allow]` the clippy lint that told you to add them. (See `CONTRIBUTING.md` for the list of clippy lints it's okay to allow in general, and why this lint ought to be allowed in particular.)

The `!` in `#!` was to point that you should allow it globally, in lib.rs

aziotctl/aziotctl-common/src/restart.rs
```diff
@@ -0,0 +1,55 @@
...
+            .output()
+            .context("Failed to call systemctl stop")?;
+
+        if result.status.success() {
+            println!("Stopped!");
+        } else {
+            eprintln!("\nError stopping {}\n", service);
+            io::stdout().write_all(&result.stdout)?;
+            io::stderr().write_all(&result.stderr)?;
+            println!();
```
daprilik, line 24: so, uh, did you just not do https://github.com/Azure/iot-identity-service/pull/105#discussion_r573032167 then?

(ditto for the other instances)

aziotctl/aziotctl-common/src/set_log_level.rs
```diff
@@ -0,0 +1,57 @@
...
+// }
+
+// impl fmt::Display for LogLevel {
+//     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+//         match self {
+//             LogLevel::Normal => write!(f, "info"),
+//             LogLevel::Debug => write!(f, "debug"),
+//         }
+//     }
+// }
```
daprilik, line 57: remove these

aziotctl/aziotctl-common/src/set_log_level.rs
```diff
@@ -0,0 +1,57 @@
...
+}
+
+fn write_log_level_file(service: &str, level: log::Level) -> Result<()> {
+    let directory = format!("/etc/systemd/system/{}.d", service);
+    fs::create_dir_all(&directory)?;
+
+    let filename = format!("{}/log-level.conf", directory);
+    let contents = format!(
+        "[Service]
+Environment=AZIOT_LOG={}",
```
daprilik, line 27: nit: just use `\n` instead

daprilik (changes requested):

aziotctl/aziotctl-common/src/set_log_level.rs
```diff
@@ -0,0 +1,57 @@
...
+    );
+
+    let mut file = fs::File::create(filename)
+        .with_context(|| format!("Failed to create log-level.conf file for {}", service))?;
+    file.write_all(contents.as_bytes())?;
+
+    Ok(())
+}
+
+// impl FromStr for LogLevel {
```
arsing, line 38: Remove this commented-out code.

aziotctl/aziotctl-common/src/set_log_level.rs
```diff
@@ -0,0 +1,57 @@
...
+use anyhow::{Context, Result};
+
+use crate::{program_name, ServiceDefinition};
+
+#[allow(clippy::missing_errors_doc)]
+pub fn set_log_level(services: &[&ServiceDefinition], level: log::Level) -> Result<()> {
+    for service in services.iter().map(|s| s.service) {
+        write_log_level_file(service, level)?;
+    }
+    Command::new("systemctl").arg("daemon-reload").output()?;
```
arsing, line 14: Add a `.context("could not run systemctl daemon-reload")` before the `?`

aziotctl/aziotctl-common/src/set_log_level.rs
```diff
@@ -0,0 +1,57 @@
...
+use std::process::Command;
+
+use anyhow::{Context, Result};
+
+use crate::{program_name, ServiceDefinition};
+
+#[allow(clippy::missing_errors_doc)]
+pub fn set_log_level(services: &[&ServiceDefinition], level: log::Level) -> Result<()> {
+    for service in services.iter().map(|s| s.service) {
+        write_log_level_file(service, level)?;
```
arsing, line 12: Add a `.with_context(|| format!("could not write log level service override for {}", service))` before the `?`

aziotctl/aziotctl-common/src/set_log_level.rs
```diff
@@ -0,0 +1,57 @@
...
+
+    let filename = format!("{}/log-level.conf", directory);
+    let contents = format!(
+        "[Service]
+Environment=AZIOT_LOG={}",
+        level
+    );
+
+    let mut file = fs::File::create(filename)
+        .with_context(|| format!("Failed to create log-level.conf file for {}", service))?;
```
arsing, line 32: This one context doesn't really help when all the other `Result`s are `?`'d without a context. In any case, if you make the change in the caller of this function as I suggested, using a context within this function becomes unnecessary.

aziotctl/aziotctl-common/src/set_log_level.rs
arsing: @daprilik

>and we don't use `f.write_str` anywhere else.

On what do you base this assertion?

aziotctl/aziotctl-common/src/system_logs.rs
arsing: Ah, right, the slice is still a temporary even if the borrows are longer-lived.

Then do `std::iter::once("-u").chain(std::iter::once(p))` instead. And that won't need a `.copied()` at the end.

aziotctl/aziotctl-common/src/system_logs.rs
```diff
@@ -0,0 +1,24 @@
+use std::ffi::OsString;
+use std::process::Command;
+
+use anyhow::{Context, Result};
+
+#[allow(clippy::missing_errors_doc)]
+pub fn get_system_logs(processes: &[&str], additional_args: &[OsString]) -> Result<()> {
```
arsing, line 7: `additional_args: &[&OsStr]`

For the caller, `options.args.iter().map(AsRef::as_ref)`

arsing (changes requested):

aziotctl/aziotctl-common/src/system_logs.rs
daprilik: I'll also throw out another alternative: `Some("-u").into_iter().chain(Some(p).into_iter())`. I find `std::iter::once` kinda redundant, since `Option` is always in-scope, and implements `IntoIter`.

 `Some("-u").into_iter().chain(Some(p).into_iter())`.

aziotctl/aziotctl-common/src/restart.rs
lfitchett: I did change it to eprintln. Did you mean for me to print the result to stderr ass well?

daprilik: I had meant to include the entire block in my comment.

Basically, what I was proposing was to do something like this:

```rust
eprintln!("\nError stopping {}\n", service);
eprintln!("stdout:")
eprintln!("=======")
io::stdout().write_all(&result.stdout)?;
eprintln!("stderr:")
eprintln!("=======")
io::stdout().write_all(&result.stderr)?;
eprintln!();
```

daprilik: actually, since you've already got an empty `eprintln!` at the end there, might as well split `eprintln!("\nError stopping {}\n", service);` into `eprintln!(); eprintln!("Error stopping {}", service); eprintln!();`

aziotctl/aziotctl-common/src/system_logs.rs
lfitchett: Honestly, I think the vec is much more readable, and I'm not too concerned about 4 extra allocations of a few bytes. This is a command line tool that is run rarely, I don't think it needs to be optimized.

daprilik: Yeah, I'm with Lee on this one. It's fun to optimize these sorts of things, but in this case it doesn't really matter, and `vec!` is far more readable.

aziotctl/src/system.rs
```diff
@@ -0,0 +1,55 @@
...
+    /// Extra args to be passed to journalctl
+    #[structopt(last = true)]
+    args: Vec<OsString>,
+}
+
+#[derive(StructOpt)]
+#[structopt(about = "Set the log level of the services")]
+pub struct LogLevelOptions {
+    #[structopt(value_name = r#"One of "trace", "debug", "info", "warn", or "error""#)]
+    log_level: Level,
```
daprilik, line 38: just use `log::Level` directly

aziotctl/aziotctl-common/src/restart.rs
```diff
@@ -0,0 +1,60 @@
...
+    if result.status.success() {
+        println!("Started!");
+    } else {
+        print_command_error(&result);
+    }
+
+    Ok(())
+}
+
+fn print_command_error(result: &std::process::Output) {
```
daprilik, line 52: Did you leave out the "Error stopping {}" line on purpose?

lfitchett: yep, the stdout ends up printing something similar, so it felt superfluous. for example:
"Failed to start aziot-tpmd.socket: Access denied
See system logs and 'systemctl status aziot-tpmd.socket' for details."

aziotctl/src/system.rs
lfitchett: log isn't in this crate's toml, its getting Level from common's re-export

arsing: If you're going to re-export it from `aziotctl-common`, continue calling it `LogLevel`.

What Daniel is saying is to not re-export it at all and instead have `aziotctl` use `log::Level` via its own `log` dependency.

aziotctl/aziotctl-common/src/restart.rs
daprilik: I would still include something along the lines of `systemd exited with non-zero status code`, as we shouldn't exclusively rely on downstream error text.

aziotctl/src/system.rs
daprilik: Oh, I'm surprised we don't already use `log` into `aziotctl` (though now that I think about it, why would we?). In any case, `log` is already part of `aziotctl`'s dependency graph, so you might as well add it to its `Cargo.toml` and use it directly.

daprilik (approved):
🚢 it

arsing (approved):

PR #106, opened 2021-02-09
Add authorization to CS

gordonwang0:
- Requires a principal specified in config.toml to modify certificates.
- Retrieving certificates is still available to everyone.
- Root has universal access.

Example authorization principal:
```
[[principal]]
uid = 1000
certs = ["aziot-edged-ca", "test*"]
```

### DIFF
--- a/aziotctl/src/init.rs
+++ b/aziotctl/src/init.rs
@@ -423,7 +423,7 @@ fn run_inner(stdin: &mut impl Reader) -> Result<RunOutput> {
             homedir_path: AZIOT_CERTD_HOMEDIR_PATH.into(),
             cert_issuance: Default::default(),
             preloaded_certs: Default::default(),
-            principals: Default::default(),
+            principal: Default::default(),
             endpoints: Default::default(),
         };
 
--- a/aziotctl/src/init.rs
+++ b/aziotctl/src/init.rs
@@ -423,6 +423,7 @@ fn run_inner(stdin: &mut impl Reader) -> Result<RunOutput> {
             homedir_path: AZIOT_CERTD_HOMEDIR_PATH.into(),
             cert_issuance: Default::default(),
             preloaded_certs: Default::default(),
+            principals: Default::default(),
             endpoints: Default::default(),
         };
 
--- a/aziotctl/src/init.rs
+++ b/aziotctl/src/init.rs
@@ -423,8 +423,8 @@ fn run_inner(stdin: &mut impl Reader) -> Result<RunOutput> {
             homedir_path: AZIOT_CERTD_HOMEDIR_PATH.into(),
             cert_issuance: Default::default(),
             preloaded_certs: Default::default(),
-            principal: Default::default(),
             endpoints: Default::default(),
+            principal: Default::default(),
         };
 
         match device_id_source {
--- a/cert/aziot-certd-config/src/lib.rs
+++ b/cert/aziot-certd-config/src/lib.rs
@@ -9,6 +9,8 @@
     clippy::missing_errors_doc
 )]
 
+use libc::uid_t as Uid;
+
 pub mod util;
 
 #[derive(Debug, PartialEq, serde::Deserialize, serde::Serialize)]
@@ -30,6 +32,16 @@ pub struct Config {
     #[serde(default, skip_serializing)]
     #[cfg_attr(not(debug_assertions), serde(skip_deserializing))]
     pub endpoints: Endpoints,
+
+    /// Authorized Unix users and the corresponding certificates.
+    ///
+    /// A Unix user with the given UID is granted write access to the certificates
+    /// specified. Wildcards may be used for certificate names.
+    ///
+    /// This authorization only affects write access. Read access for certificates is
+    /// granted to all users.
+    #[serde(default)]
+    pub principals: Vec<Principal>,
 }
 
 /// Configuration of how new certificates should be issued.
@@ -303,6 +315,16 @@ impl Default for Endpoints {
             },
         }
     }
+}
+
+/// Map of a Unix UID to certificates with write access.
+#[derive(Debug, PartialEq, serde::Deserialize, serde::Serialize)]
+pub struct Principal {
+    /// Unix UID.
+    uid: Uid,
+
+    /// Certificates for which the given UID has write access.
+    certs: Vec<String>,
 }
 
 #[cfg(test)]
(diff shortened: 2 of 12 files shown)


cert/aziot-certd-config/src/lib.rs
```diff
@@ -30,6 +30,16 @@ pub struct Config {
     #[serde(default, skip_serializing)]
     #[cfg_attr(not(debug_assertions), serde(skip_deserializing))]
     pub endpoints: Endpoints,
+
```
arsing, line 33: For consistency, move this above the `endpoints` field. (`endpoints` is always the last field.)

cert/aziot-certd-config/src/lib.rs
```diff
@@ -305,6 +315,16 @@ impl Default for Endpoints {
     }
 }
 
+/// Map of a Unix UID to certificates with write access.
+#[derive(Debug, PartialEq, serde::Deserialize, serde::Serialize)]
+pub struct Principal {
+    /// Unix UID.
+    pub uid: libc::uid_t,
+
+    /// Certificates for which the given UID has write access. Wildcards may be used.
```
arsing, line 324: `s/Certificates/Certificate IDs/`

cert/aziot-certd/src/error.rs
```diff
@@ -22,6 +23,9 @@ impl std::fmt::Display for Error {
             Error::InvalidParameter(name, _) => {
                 write!(f, "parameter {:?} has an invalid value", name)
             }
+            Error::Unauthorized(user, id) => {
+                write!(f, "User {} not authorized to modify cert {}.", user, id)
```
arsing, line 27: `"user {} is not authorized to modify the cert {}"`

(lower-case u, no period, grammar)

cert/aziot-certd/src/http/create.rs
```diff
@@ -16,14 +17,17 @@ impl http_common::server::Route for Route {
...
         path: &str,
         _query: &[(std::borrow::Cow<'_, str>, std::borrow::Cow<'_, str>)],
-        _extensions: &http::Extensions,
+        extensions: &http::Extensions,
     ) -> Option<Self> {
         if path != "/certificates" {
             return None;
         }
 
+        let uid = extensions.get::<libc::uid_t>().cloned()?;
```
arsing, line 26: `.copied()` instead of `.cloned()`

cert/aziot-certd/src/lib.rs
```diff
@@ -784,3 +820,12 @@ fn get_cert_inner(
     let bytes = load_inner(&path)?;
     Ok(bytes)
 }
+
+fn principal_to_map(
+    principal: Vec<Principal>,
+) -> std::collections::BTreeMap<libc::uid_t, Vec<String>> {
```
arsing, line 826: Make this a `BTreeMap<libc::uid_t, Vec<wildmatch::WildMatch>` so that you don't have to parse the `String` into a `WildMatch` again and again for every request.

cert/aziot-certd/src/lib.rs
```diff
@@ -130,6 +146,24 @@ impl Api {
...
+
+    fn authorize(&self, user: libc::uid_t, id: &str) -> bool {
+        // Root user is always authorized.
+        if user == 0 {
+            return true;
+        }
+
+        // Authorize user based on stored principals config.
+        if let Some(certs) = self.principals.get(&user) {
+            for cert in certs {
```
arsing, line 158: (After making the change to `principals_to_map` from my comment there.)

`return certs.any(|cert| cert.is_match(id));`

arsing (changes requested):

cert/aziot-certd-config/src/lib.rs
gordonwang0: Fixed

cert/aziot-certd-config/src/lib.rs
gordonwang0: Fixed

cert/aziot-certd/src/error.rs
gordonwang0: Fixed

cert/aziot-certd/src/http/create.rs
gordonwang0: Fixed

cert/aziot-certd/src/lib.rs
gordonwang0: Fixed

cert/aziot-certd/src/lib.rs
gordonwang0: Fixed

aziotctl/src/init.rs
```diff
@@ -423,6 +423,7 @@ fn run_inner(stdin: &mut impl Reader) -> Result<RunOutput> {
             homedir_path: AZIOT_CERTD_HOMEDIR_PATH.into(),
             cert_issuance: Default::default(),
             preloaded_certs: Default::default(),
+            principal: Default::default(),
```
arsing, line 426: This will need to give IS access to the device ID cert in case `device_id_source` is `Some()`.

Add a `if device_id_source.is_some() { certd_config.principal.push(...); }` at the end of this block, right before it returns `certd_config` (line 636).

gordonwang0: Does IS need to modify the device ID cert? Reading the cert is authorized for everyone.

arsing: I thought dynamically issued certs (requesting a cert that doesn't exist but has an issuance defined, so it's created as part of getting it) would require authorization equivalent to creating them, but I see now you only check the auth for import and delete.

I guess it's fine.

gordonwang0: Do you know how to fix the error:
```
could not serialize aziot-certd config

Caused by:
    values must be emitted before tables', aziotctl/src/init.rs:1161:46
```

I tried a few things that didn't work. To repro: `cargo test -p aziotctl`

arsing: https://news.ycombinator.com/item?id=24853386

Example of a workaround for the same reason: https://github.com/Azure/iot-identity-service/blob/8c471e53a8c77276f0af7e4f0d0e5eed06d57eed/identity/aziot-identityd-config/src/lib.rs#L90-L96

gordonwang0: Should be fixed now. Thanks.

arsing: Makes sense. The serializer can't resume the struct after a struct-array member's `[[member]]` section.

arsing (approved):

massand (approved):

PR #107, opened 2021-02-10
[e2e tests] tag DPS after creation

daprilik:
I realized that `az iot dps create` is also impacted by https://github.com/Azure/azure-cli/issues/13497

Two observations I had while fixing this:

- Why does `az iot dps create` silently accept invalid CLI flags?? I passed `--tags`, and instead of failing, it just silently succeeded? :roll_eyes: 
- For some reason, `az resource tag` _must_ come after `az iot dps linked-hub create`, or else the tag won't "stick"

### DIFF
--- a/ci/e2e-tests.sh
+++ b/ci/e2e-tests.sh
@@ -338,13 +338,13 @@ trap "
     case "$test_name" in
         dps-*)
             echo 'Creating DPS...' >&2
-            dps_scope_id="$(
+            dps_result="$(
                 az iot dps create \
                     --resource-group "$AZURE_RESOURCE_GROUP_NAME" \
-                    --name "$common_resource_name" \
-                    --tags "$resource_tag" \
-                    --query 'properties.idScope' --output tsv
+                    --name "$common_resource_name"
             )"
+            dps_scope_id="$(<<< "$dps_result" jq '.properties.idScope' -r)"
+            dps_resource_id="$(<<< "$dps_result" jq '.id' -r)"
 
             az iot dps linked-hub create \
                 --resource-group "$AZURE_RESOURCE_GROUP_NAME" --dps-name "$common_resource_name" \
@@ -358,6 +358,14 @@ trap "
                         --resource-group "$AZURE_RESOURCE_GROUP_NAME" --name "$common_resource_name" \
                         --query 'location' --output tsv
                 )"
+
+            # `az iot dps create` doesn't have `--tags`, so tag it manually.
+            #
+            # Ref: https://github.com/Azure/azure-cli/issues/13497
+            >/dev/null az resource tag \
+                --ids "$dps_resource_id" \
+                --tags "$resource_tag"
+
             echo 'Created DPS' >&2
             ;;
     esac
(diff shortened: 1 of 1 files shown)


arsing:
I don't follow.

```sh
$ az group create --name daprilik

$ az iot dps create --resource-group daprilik --name arsing-daprilik --tags 'foo=bar'

$ az resource show --resource-group daprilik --name arsing-daprilik --resource-type 'Microsoft.Devices/ProvisioningServices' --query tags

{
  "foo": "bar"
}
```

daprilik:
Huh, very weird. I tested that exact thing on my dev box, and it didn't work. Then again, I did immediately follow it up with `az iot dps linked-hub create`, so it could be that the tags _were_ added as part of the initial create, and then later stripped.

In either case, the current e2e tests in-tree will not delete the DPS after the test is done (for whatever reason)

arsing:
lol, indeed.

```sh
$ az iot hub create --resource-group daprilik --name arsing-daprilik --sku S1 --unit 1 --partition-count 2

$ az resource show --resource-group daprilik --name arsing-daprilik --resource-type 'Microsoft.Devices/ProvisioningServices' --query tags

{
  "foo": "bar"
}

$ az iot dps linked-hub create --resource-group daprilik --dps-name arsing-daprilik --connection-string "$(az iot hub connection-string show --resource-group daprilik --hub-name arsing-daprilik --query 'connectionString' --output tsv)"

$ az resource show --resource-group daprilik --name arsing-daprilik --resource-type 'Microsoft.Devices/ProvisioningServices' --query tags

{}
```

daprilik:
Is this an upstream bug? If so, we might want to open an issue about this (somewhere, not sure where)

arsing (approved):

ci/e2e-tests.sh
```diff
@@ -358,6 +358,14 @@ trap "
                         --resource-group "$AZURE_RESOURCE_GROUP_NAME" --name "$common_resource_name" \
                         --query 'location' --output tsv
                 )"
+
+            # `az iot dps create` doesn't have `--tags`, so tag it manually.
```
arsing, line 362: Just fix this comment to be more accurate.

daprilik: I'll replace it with a comment explaining that the tags must be added after the call to `linked-hub create`

ci/e2e-tests.sh
```diff
@@ -358,6 +358,13 @@ trap "
                         --resource-group "$AZURE_RESOURCE_GROUP_NAME" --name "$common_resource_name" \
                         --query 'location' --output tsv
                 )"
+
+            # For some reason, `az resource tag` must come after
```
arsing, line 362: Well, it's more like:

```
# `az iot dps linked-hub create` deletes the tags on the DPS that were set by `az iot dps create`
# for some unknown reason, so we need to tag it again.
```

Also, please restore the `--tags` parameter of the initial `az iot dps create` so that the DPS is tagged in case the script fails at the `linked-hub create` step.

daprilik: dangit, my timing \*shakes fist\*

Should I open a hotfix PR for this, or can I lump it in with my upcoming dps-symkey PR?

arsing: Latter is fine. No big deal. It's not like e2e tests are running automatically anyway.

PR #109, opened 2021-02-11
Fix build for new Rust stable 1.50.0

arsing:
Other changes:

- Fix mutex locking code in tpmd that didn't check for lock failures due to
  poisoning.

- Update docs example of X.509 CA provisioning to pass in the cert subject
  to the openssl commands instead of being prompted interactively.


gordonwang0 (approved):

identity/aziot-identityd/src/identity.rs
```diff
@@ -670,13 +670,14 @@ impl IdentityManager {
...
                 })?;
 
-                let _ = self
-                    .cert_client
-                    .create_cert(&identity_cert, &csr, None)
-                    .await
-                    .map_err(|err| {
-                        Error::Internal(InternalError::CreateCertificate(Box::new(err)))
-                    })?;
+                drop(
```
daprilik, line 673: Oh cool, a new `pedantic` clippy lint that I'm not a huge fan of.

See https://rust-lang.github.io/rust-clippy/master/#let_underscore_drop and the original PR https://github.com/rust-lang/rust-clippy/pull/6305

Personally, I feel like this lint is a bit _too_ aggressive. If I'm using `let _ =` on a value, I'm doing so because I don't care about it _at all_, and therefore don't care when/where it gets dropped. It seems like a truly pedantic generalization of the much more useful `let_underscore_lock` lint.

I'd personally vote to suppress this lint, but folks think we should keep it, so be it.

arsing: Ugh, I didn't notice that rustfmt had expanded it to more lines. I'd initially typed `drop(self.client_client(` and `})?);` which was palatable.

I agree it's worth suppressing it.

arsing: Done.

daprilik (approved):

arsing:
I need time to investigate the CentOS package failure; it's not caused by this PR so it has to be some upstream change. To unblock PRs for now for rustup fixes I'm going to temporarily disable CentOS 7 packaging.

PR #111, opened 2021-02-11
Make `registration_id` optional for dps x509

daprilik:
After talking to @massand, it looks like there are actually several subtle requirements around this change.

I'm opening this PR as a draft, as I believe that seeing the implementation of the feature can help drive the discussion around how we should be handle various edge-cases.


identity/aziot-identityd-config/src/lib.rs
```diff
@@ -75,7 +75,7 @@ pub enum DpsAttestationMethod {
         symmetric_key: String,
     },
     X509 {
-        registration_id: String,
+        registration_id: Option<String>,
```
arsing, line 78: I still suggest removing it entirely and deriving it entirely from the cert CN.

identity/aziot-identityd/src/identity.rs
```diff
@@ -522,12 +522,27 @@ impl IdentityManager {
...
+                                &identity_cert,
+                                IdentitySubject::RegistrationId(registration_id.as_deref()),
+                            )
+                            .await?;
+
+                        let cert_cn = cert
+                            .subject_name()
+                            .entries_by_nid(openssl::nid::Nid::COMMONNAME)
+                            .next()
+                            .expect("CN is guaranteed to be present")
```
arsing, line 537: It isn't. Make it an `InternalError::ParseCertCN` too.

identity/aziot-identityd-config/src/lib.rs
massand: If we are going with rotatable certs and we need to preserve backwards-compat with the previous IoT Edge edgelet behavior, then we need to follow the same prioritization - first, we look for registration_id in the IS config.toml, then we look for CN in certificate. If we need to issue/renew a new identity, the CN from registration_id should applied in the CSR for the new certificate. If that is missing, then CN from certificate should be reused to generate the CSR.

arsing: IoT Hub rejects the request if registration ID does not match the cert CN. That's what triggered this change in the first place. So anyone who has an existing config.yaml with the `registration_id` explicitly specified is already setting it the same as the cert CN.

arsing: Ugh, but if the cert doesn't already exist and IS need to creates a CSR, IS needs to know the CN first.

So we need a property just for the initial CSR. It should be called something other than `registration_id` to make that clear.

Also, in case the cert exists, IS should ignore this property. That is, prioritization of this property vs cert CN should be opposite of what you suggest.

daprilik: Just pushed up a pass at this prioritization logic.

daprilik: Also, I agree that we aught to rename `registration_id` to something like that better reflects its role as a CN fallback. I have no strong opinions on the name, so bikeshed away :D

daprilik:
I'm holding off on fixing the failing test cases until we decide whether or not we are changing the name of `registration_id`

identity/aziot-identityd-config/src/lib.rs
massand: I was thinking along the same lines about IS ignoring the registration_id property. However, there is a disconnect with documented APIs. 

DPS doesn't require CN in the X.509 cert, which is why registration_id from config.toml should be used if CN is not found and should be prioritized first. Also, the [DPS REST API docs](https://docs.microsoft.com/en-us/rest/api/iot-dps/runtimeregistration/registerdevice) indicate registration_id is mandatory for all types of auth. Which is why registration_id was never optional. 

Unfortunately, the DPS REST APIs don't document the X.509 cert CN-matching-the-registration-id requirement, which means it can change anytime. I rather just add a descriptive error for the user about the mismatch, if the DPS service fails to auth with mismatched CN in X.509 cert. And preserve the previous priority order of config.toml registration_id before the certificate CN.

arsing: >DPS doesn't require CN in the X.509 cert

Daniel, can you test this?

>Also, the DPS REST API docs indicate registration_id is mandatory for all types of auth. Which is why registration_id was never optional.

We're not talking about registration ID in the HTTP request. We're talking about registration ID in the IS config file, ie what value is eventually used in the HTTP request.

>I rather just add a descriptive error for the user about the mismatch, if the DPS service fails to auth with mismatched CN in X.509 cert.

I assume we print DPS's response in the logs. That's sufficient. We shouldn't hard-code their validation into our code.

>And preserve the previous priority order of config.toml registration_id before the certificate CN.

Assuming that the existing cert happens to not have a CN, we can use the registration ID. This is just one more case on top of the existing one (no existing cert). I still don't see why if the cert has a CN we should prefer the registration ID.

daprilik: > Daniel, can you test this?

I'll give it a try tomorrow morning.

> I assume we print DPS's response in the logs.

Well, only when `AZIOT_LOGS=debug`, and even then, the response back from DPS is a very uninformative and terse JSON blob that just says the operation is "unauthorized", which doesn't tell you much. I would be in favor of detecting this error case and special casing it (if it's still applicable after this PR is merged)

arsing: >Well, only when `AZIOT_LOGS=debug`

We ought to parse [the response](https://docs.microsoft.com/en-us/rest/api/iot-dps/runtimeregistration/registerdevice#provisioningserviceerrordetails) and print the message (not the whole JSON, which can continue being debug-only, just the message).

>I would be in favor of detecting this error case and special casing it

Well, special-casing it means "if DPS said unauthorized *and* if registration_id != cert.subject.cn then print 'It might be because registration_id and cert.subject.cn don't match'". That "might" is what I'm not a fan of.

arsing: Thinking about it more since then, I agree that Rajeev's suggestion is better in the long run. Optional registration ID, registration ID preferred over cert CN.

Do make the change to surface the DPS error response as I said in my previous comment.

arsing:
Since manual X.509 has the same issue wrt device ID and cert CN, should that be optional too?

identity/aziot-identityd-config/src/lib.rs
daprilik: > DPS doesn't require CN in the X.509 cert

So, I'm not really sure how exactly we can test this.

It's not like I can just register a device using a cert with an empty CN, since the [register device API](https://docs.microsoft.com/en-us/rest/api/iot-dps/runtimeregistration/registerdevice) requires a `registrationId` string that matches the CN (or else it'll just return `unauthorized`), yet the `registrationId` cannot be empty.

identity/aziot-identityd/src/identity.rs
```diff
@@ -522,12 +523,20 @@ impl IdentityManager {
...
+                                &identity_cert,
+                                IdentitySubject::RegistrationId(
+                                    registration_id.as_deref().map(Into::into),
+                                ),
+                            )
+                            .await?;
+                        let cert_cn = extract_cn(&cert)?;
+
+                        if cert_cn.is_empty() && registration_id.map_or(false, |s| !s.is_empty()) {
+                            // ???
```
daprilik, line 538: @massand I think I recall you mentioning that this should be special cased?

identity/aziot-identityd-config/src/lib.rs
arsing: The ask is to test the API with a valid registration ID in the URL but no CN in the client cert.

daprilik: > Do make the change to surface the DPS error response as I said in my previous comment.

As it turns out, we are [already doing that](https://github.com/Azure/iot-identity-service/blob/d0cbd913e90dd46ef45c15d8be3f4f2d1ae9b4b2/identity/aziot-dps-client-async/src/lib.rs#L331-L339), but it looks like some changes merged in #104 broke the error-handling logic during provisioning, namely, these lines: https://github.com/Azure/iot-identity-service/pull/104/files#diff-d29aaf0847f8c3c0d8726007b4cc7f68eaf1d3f9892243837c4f0939dffc7275R80-R82. 

I've pushed up a commit that fixes this concrete issue, but upon re-reviewing #104, I also noticed https://github.com/Azure/iot-identity-service/blob/d0cbd913e90dd46ef45c15d8be3f4f2d1ae9b4b2/config-common/src/watcher.rs#L73, which will result in runtime re-configuration errors being silently swallowed. 

@gordonwang0, can you double-check that I'm reading this right. If I am, you may want to open a dedicated PR to fix these issues.

daprilik: > The ask is to test the API with a valid registration ID in the URL but no CN in the client cert.

Okay, sure, I just did that. As I had expected, it fails with the same uninformative `unauthorized` error.

massand: Thanks for checking. 
There doesn't seem to be a way to create an individual enrollment with an identity certificate containing empty CN (I just tried). This leads me to believe [the docs are out of date](https://docs.microsoft.com/en-us/azure/iot-edge/how-to-auto-provision-x509-certs?view=iotedge-2018-06#configure-the-device-with-provisioning-information). It is best to confirm with DPS team about this behavior, instead of reverse-engineering it. 

If CN is always required and it should always match registration_id in the DPS REST API payload, then we should sending the registration_id completely in X.509 case. We can rename the setting to `device_id_cert_subject_name` as a required field for x509, and we can document why it is required in the config.toml.

identity/aziot-identityd/src/identity.rs
massand: Judging by DPS's behavior of rejecting empty CN certs, the first check is sufficient to error this out and fail the X.509 attestation. Initially, I thought registration_id would be used in the JSON payload when CN is missing. However, we should rename `registration_id` to `device_id_cert_subject_name` and request a cert.

daprilik:
@arsing and @massand, I've un-drafted the PR since it's getting pretty close to done. Can you give it a once-over and let me know what (if any) things are left to do?

daprilik:
After lots of discussion and some back-and-forth with the DPS folks, we've decided to keep the behavior as-is and close this PR. 


Instead, to help mitigate the current issue around having a mismatched `registration_id` and certificate CN, we will add a new `aziotctl check` that warns if a mismatch is detected.

PR #116, opened 2021-02-16
Implement new super-config and `aziotctl config apply`

arsing:
The super-config is a new config that is an amalgam of
all the individual services' config files, with some tweaks to simplify them.
The new config lives at `/etc/aziot/config.toml` by default.

The user edits this file as they want, then runs `aziotctl config apply`
to "apply" the changes to the individual services. They way this works is that
`aziotctl config apply` writes the files
`/etc/aziot/{key,cert,identity,tpm}d/config.d/00-super.toml` with contents
derived from the super config.

Other changes:

- Rename `aziotctl init` to `aziotctl config wizard`

- Fix `aziotctl config wizard` to correctly add KS and CS principals for
  dynamically-issued certs.

- Allow services to start up without base `config.toml`, since
  a sufficient config may be available via the patches under `config.d`.
  This is indeed what will happen when the user runs `aziotctl config apply`
  on a clean install.


config-common/src/lib.rs
```diff
@@ -4,11 +4,15 @@
...
     clippy::default_trait_access,
+    clippy::large_enum_variant,
     clippy::missing_errors_doc,
     clippy::module_name_repetitions
 )]
 
 pub mod error;
+
+pub mod super_config;
+
```
daprilik, line 15: any reason for this whitespace?

config-common/src/lib.rs
```diff
@@ -20,10 +24,18 @@ pub fn read_config<TConfig>(
...
+        Ok(contents) => toml::from_slice(&contents)
+            .map_err(|err| ErrorKind::ReadConfig(Some(config_path.to_owned()), Box::new(err)))?,
+
+        Err(err) if err.kind() == std::io::ErrorKind::NotFound => {
+            toml::Value::Table(Default::default())
+        }
+
+        Err(err) => {
+            return Err(ErrorKind::ReadConfig(Some(config_path.to_owned()), Box::new(err)).into())
+        }
```
daprilik, line 37: nit: not a _huge_ fan of the whitespace here, but whatever. After all, the braces do a good enough job padding the various match arms, so why add more?

daprilik (commented):
Sending out some low-hanging comments before diving into the meat of the PR

CONTRIBUTING.md
```diff
@@ -57,6 +57,8 @@ contact [opencode@microsoft.com](mailto:opencode@microsoft.com) with any additio
 
         Note that when the doc comment that raised the lint is a crate-level inner doc comment, there is unfortunately no way to suppress the lint for that doc comment other than by adding a crate-level `#![allow]` attribute for it, which also disables the lint for every other doc comment in the crate.
 
+    1. `clippy::large_enum_variant`: This lint fires for any enum that has at least one "small" variant and at least one "big" variant, and suggests boxing the "big" variant so that the enum doesn't waste space in case it's the "small" variant. However this makes `match`ing such enums harder as `box` patterns are not stable.
```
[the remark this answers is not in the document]
arsing: Done.

aziotctl/src/check.rs
```diff
@@ -19,7 +19,7 @@ use crate::internal::check::{
 
 #[derive(StructOpt)]
 #[structopt(about = "Check for common config and deployment issues")]
-pub struct CheckOptions {
+pub struct Options {
```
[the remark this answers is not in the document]
arsing: Reverted for check and check-list.

config-common/src/lib.rs
arsing: I was doing a refactor and forgot to finish it. Making `mod error` not pub and instead making the `Error` a pub re-export. Done now.

config-common/src/lib.rs
arsing: Well, *I* didn't add the braces, so... (Notice also the missing semi-colon on line 36.)

Fixed now.

aziotctl/src/main.rs
```diff
@@ -14,18 +14,17 @@
 use anyhow::Result;
 use structopt::StructOpt;
 
-mod internal;
-
 mod check;
 mod check_list;
-mod init;
+mod config;
+mod internal;
```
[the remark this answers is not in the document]
arsing: Done.

aziotctl/src/config/wizard.rs
```diff
@@ -121,37 +100,37 @@ pub(crate) fn run() -> Result<()> {
     if let Some(preloaded_device_id_pk_bytes) = preloaded_device_id_pk_bytes {
         println!("Note: Symmetric key will be written to /var/secrets/aziot/keyd/device-id");
 
-        create_dir_all("/var/secrets/aziot/keyd", &aziotks_user, 0o0700)?;
-        write_file(
+        super::create_dir_all("/var/secrets/aziot/keyd", &aziotks_user, 0o0700)?;
```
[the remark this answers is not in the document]
arsing: There's a lot of them, and it doesn't really make sense to list them out an extra time.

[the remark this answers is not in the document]
daprilik: Yes, I agree, there are a lot of `super::`s, and it doesn't make sense to keep listing them out all the time :stuck_out_tongue:

aziotctl/src/config/mod.rs
```diff
@@ -0,0 +1,126 @@
...
+    for sections in connection_string.split(';') {
+        let mut parts = sections.split('=');
+        match parts
+            .next()
+            .expect("str::split always returns at least one part")
+        {
+            HOSTNAME_KEY => iothub_hostname = parts.next(),
+            DEVICEID_KEY => device_id = parts.next(),
+            SHAREDACCESSKEY_KEY => symmetric_key = parts.next(),
+            _ => (), // Ignore extraneous component in the connection string
```
daprilik, line 107: Is this the right behavior here? Shouldn't we at-least trigger some warning?

config-common/src/super_config.rs
```diff
@@ -0,0 +1,232 @@
...
+#[derive(Debug, serde::Deserialize, serde::Serialize)]
+pub struct Config {
+    pub hostname: Option<String>,
+
+    pub provisioning: Provisioning,
+
+    pub localid: Option<aziot_identityd_config::LocalId>,
+
+    #[serde(default)]
+    pub aziot_keys: std::collections::BTreeMap<String, String>,
```
daprilik, line 31: pls just import `BTreeMap` once instead of using the FQT everywhere.

massand:
> 		for d in /var/lib/aziot/certd /var/lib/aziot/identityd /var/lib/aziot/keyd /var/lib/aziot/tpmd; do

nit: Abbreviate the path range here, just like the change above. 

---
Refers to: contrib/debian/postinst:46 in 8c76940. [](commit_id = 8c76940d30df10e53faaa07717963d030f7f8867, deletion_comment = False)

config-common/src/super_config.rs
```diff
@@ -0,0 +1,232 @@
...
+/// - Wherever cert IDs were used to link certs between services, the key is referenced directly.
+///   For example, instead of IS provisioning referencing a cert by its cert ID and CS associating a cert ID with a cert URI or an issuance method,
+///   the IS provisioning references the cert URI or issuance method directly.
+///
+/// - A consequence of the above is that the EST endpoint spec only allows one of identity or bootstrap identity X.509 auth to be specified, not both.
+///   (If bootstrap identity was specified, the regular identity would be IDs.)
+///
+/// Unfortunately it's not easy to do this without duplicating all the individual services' config types.
+/// Inner types are reused when possible, ie when they don't need modification to work with the super-config.
+#[derive(Debug, serde::Deserialize, serde::Serialize)]
```
daprilik, line 22: similarly, for consistency's sake, we've been using `use serde::{Deserialize, Serialize};` everywhere else in `aziotctl`

config-common/src/super_config.rs
```diff
@@ -0,0 +1,232 @@
...
+///
+/// - Wherever cert IDs were used to link certs between services, the key is referenced directly.
+///   For example, instead of IS provisioning referencing a cert by its cert ID and CS associating a cert ID with a cert URI or an issuance method,
+///   the IS provisioning references the cert URI or issuance method directly.
+///
+/// - A consequence of the above is that the EST endpoint spec only allows one of identity or bootstrap identity X.509 auth to be specified, not both.
+///   (If bootstrap identity was specified, the regular identity would be IDs.)
+///
+/// Unfortunately it's not easy to do this without duplicating all the individual services' config types.
+/// Inner types are reused when possible, ie when they don't need modification to work with the super-config.
```
daprilik, line 21: What if we use g e n e r i c s to to avoid some of the more blatant duplication.

On one hand, it's a bit overkill, but on the other hand, it will guarantee that things are kept in-sync if/when configuration types start getting changed.

aziotctl/src/config/mod.rs
arsing: Yes, it's the right behavior. It's architected like a query string and sometimes has other parameters for client SDKs that we don't care about.

config-common/src/super_config.rs
arsing: Done.

config-common/src/super_config.rs
arsing: Done.

config-common/src/super_config.rs
arsing: The guarantee already exists, because the translation code in `apply` will fail if either the super-config or the individual service config is changed. (That's why I ensure all the code uses destructuring instead of field access, doesn't use `..` in pats, etc.)

I did think of generics, but some of the types like the `Est` tree are sufficiently different that it may not work. I'll check again.

arsing:
On Daniel's suggestion I split the commit into two parts, one to refactor existing `aziotctl init` to `aziotctl config wizard` (entirely a restructuring of existing code), and the other to add `aziotctl config apply` (entirely new code), to make the diffs cleaner. I recommend reviewing them separately.

aziotctl/src/config/apply.rs
```diff
@@ -0,0 +1,635 @@
...
+
+//! This subcommand takes the specific super-config file, converts it into the individual services' config files,
+//! writes those files, and restarts the services.
+
+use anyhow::{anyhow, Context};
+
+use config_common::super_config;
+
+#[derive(structopt::StructOpt)]
+pub(crate) struct Options {
```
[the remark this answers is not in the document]
daprilik: Well, in that case, I guess you might as well switch the other commands to use `Options` directly as well? My comment from earlier still applies, but as always, consistency takes precedent.

gordonwang0 (approved):

aziotctl/aziotctl-common/test-files/apply/dps-tpm/keyd.toml
```diff
@@ -4,5 +4,5 @@ homedir_path = "/var/lib/aziot/keyd"
 [preloaded_keys]
 
 [[principal]]
-uid = 1001
+uid = 5557
```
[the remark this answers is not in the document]
arsing: Ugh, it's wrong. It should be 5556. I'll fix it.

But the reason for the change is to be consistent with the files in iotedge that were using 5555+, and to reduce any chance of colliding with the default 1000-ish users. I "standardized" on `aziotcs` = 5555, `aziotid` = 5556, `aziotks` = 5557, `iotedge` = 5558

aziotctl/src/config/apply.rs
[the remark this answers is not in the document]
arsing: Done.

aziotctl/aziotctl-common/test-files/apply/dps-tpm/keyd.toml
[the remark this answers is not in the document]
arsing: Fixed.

gordonwang0 (approved):

massand (approved):

PR #118, opened 2021-02-18
Abort IS if a runtime config update fails

gordonwang0:
Identity Service config updates may require reprovisioning with IoT Hub and updating Hub identities. If reprovisioning fails, IS cannot fall back on the old config because the old config may no longer be valid (e.g. the user created a new IoT Hub and deleted the old one). So, IS will abort and the user will need to fix the config and restart IS.

### DIFF
--- a/identity/aziot-identityd/src/lib.rs
+++ b/identity/aziot-identityd/src/lib.rs
@@ -370,8 +370,10 @@ impl Api {
 
         let _ = match trigger {
             ReprovisionTrigger::ConfigurationFileUpdate => {
+                // For now, skip reprovisioning if there's a valid backup. This means config file
+                // updates will only reconcile identities.
                 self.id_manager
-                    .provision_device(self.settings.provisioning.clone(), false)
+                    .provision_device(self.settings.provisioning.clone(), true)
                     .await?
             }
             ReprovisionTrigger::Api => {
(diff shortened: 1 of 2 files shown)


identity/aziot-identityd/src/lib.rs
```diff
@@ -371,7 +371,7 @@ impl Api {
         let _ = match trigger {
             ReprovisionTrigger::ConfigurationFileUpdate => {
                 self.id_manager
-                    .provision_device(self.settings.provisioning.clone(), true)
+                    .provision_device(self.settings.provisioning.clone(), false)
```
arsing, line 374: In the case of DPS provisioning, this will lead to repeated re-provisioning for every file update, right? That is what we were talking about avoiding in the meeting. We need to check the backup to see if the provisioning info has changed or not.

I'll add you to the thread with Varun.

gordonwang0: Per that thread, I think we'll need to discuss further what the final behavior should be. For now, I'd like to fix the bug where config file update errors were getting ignored. Could you take another look at this PR just for that?

arsing (approved):

identity/aziot-identityd/src/lib.rs
```diff
@@ -512,8 +514,14 @@ impl UpdateConfig for Api {
     type Error = Error;
 
     async fn update_config(&mut self, new_config: config::Settings) -> Result<(), Self::Error> {
+        // Abort if update fails; i.e. if reconciling identities fails. This may happen if there's
+        // an error in the config or IoT Hub. The user will have to fix the config or IoT Hub and
+        // restart IS.
         self.update_config_inner(new_config, ReprovisionTrigger::ConfigurationFileUpdate)
             .await
+            .expect("failed to reprovision");
```
massand, line 522: Does this fail even if there is network connectivity failure? The same IS shutdown doesn't happen during normal identity operations. Hence, the question

gordonwang0: Yes, this would fail if Hub can't be reached to create/delete identities.

Normal identity operations are able to return an error via the HTTP response. Reloading the config can't do that because it's not an API call. It's probably better to abort so that it's obvious something is wrong instead of continuing here.

massand: I thought the list of reasons for IS failure was scoped to provisioning, not reconcilation. I would stay consistent with the identity API behavior and not kill the process, if Hub calls fail.

gordonwang0: Got it. I've updated the behavior so IS doesn't abort on reconciliation failure.

massand (approved):

PR #121, opened 2021-02-24
Fix for dynamic registration

massand:
The order of creating the reconciliation thread before the notify listener thread fixed the issue of the first missed update.
Verified by manually testing on Ubuntu 18.04.

### DIFF
--- a/config-common/src/watcher.rs
+++ b/config-common/src/watcher.rs
@@ -24,12 +24,33 @@ pub fn start_watcher<TApi>(
     // since the message is resent to file change receiver using a blocking send.
     // When the number of messages is set to 1, then main thread appears to block.
     let (file_changed_tx, mut file_changed_rx) = tokio::sync::mpsc::channel(2);
+    
+    let config_path_clone = config_path.clone();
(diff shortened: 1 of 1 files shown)


config-common/src/watcher.rs
```diff
@@ -25,11 +25,34 @@ pub fn start_watcher<TApi>(
     // When the number of messages is set to 1, then main thread appears to block.
     let (file_changed_tx, mut file_changed_rx) = tokio::sync::mpsc::channel(2);
 
+    let config_path_clone = config_path.clone();
+    let config_directory_path_clone = config_directory_path.clone();
+
+    // Start file change listener that asynchronously updates service config.
+    tokio::spawn(async move {
```
gordonwang0, line 32: Should this initial config file read and update always complete before starting the thread that watches the config files? If yes, I think it's better to just do this read+update in the main thread instead of spawning another thread to handle it.

massand: `async move` returns a future. Both threads are spawned instantly on startup. 

Updating the wording to reflect current state. Thanks.

gordonwang0 (approved):

PR #123, opened 2021-02-24
[aziotctl check] Add registration_id == identity cert common name check

daprilik:
Following up on the discussion from #111.

This new check will check if `registration_id` matches the identity certificate's common name when using x509 provisioning.

* * *

A couple points of discussion:
- Should this be a warning or an error? Technically DPS could relax this requirement at any time, so making it a warning seems more appropriate
- On that note, I feel like the warning message should include a line along the lines of "This requirement may change at some point in the future".
- Should we be including the certificate CN as part of the itoedge check output (both human readable and JSON)? In other words, could this be considered "sensitive" information?


daprilik:
It looks like DPS might have beat us to the punch, and may have already loosened this restriction.
We can validate if that's the case, and close out this PR if need be.

PR #126, opened 2021-02-25
Fix preloaded-certs-validity check to correctly handle lists where not every element is another preloaded cert.

arsing:


aziotctl/src/internal/check/checks/certs_preloaded.rs
```diff
@@ -36,41 +39,65 @@ impl CertsPreloaded {
     ) -> Result<CheckResult> {
         let preloaded_certs = &unwrap_or_skip!(&cache.cfg.certd).preloaded_certs;
 
-        // TODO?: support returning multiple check results from a single check
```
daprilik, line 39: You might want to leave this TODO comment, since this is something the check would benefit from if it were supported.

daprilik (approved):
Yep, in hindsight, it makes sense that `Ids` can point to other `Ids`. Not sure why I only did a shallow traversal in the first place.

gordonwang0 (approved):

aziotctl/src/internal/check/checks/certs_preloaded.rs
arsing: Done.

PR #127, opened 2021-02-25
Fixes from bugbash

arsing:
- Change config.toml.template from `r--` to `rw-`

  This allows the user's editor to write it in-place instead of overwriting it.

- Enable logging for `hyper` via `env_logger`.

- Disable HTTP/2 via ALPN.

  This is a temporary workaround until we figure out why IS can't talk HTTP/2
  to Edge Hub even after enabling the `http2` feature of `hyper`.

- Parse DPS global endpoint as a `url::Url` so that invalid values
  are rejected at config parse time.

- Allow duplicate principals and merge them at runtime.

### DIFF
--- a/contrib/centos/aziot-identity-service.spec
+++ b/contrib/centos/aziot-identity-service.spec
@@ -174,7 +174,7 @@ fi
 %attr(400, aziottpm, aziottpm) %{_sysconfdir}/aziot/tpmd/config.toml.default
 %attr(700, aziottpm, aziottpm) %dir %{_sysconfdir}/aziot/tpmd/config.d
 
-%attr(700, root, root) %{_sysconfdir}/aziot/config.toml.template
+%attr(600, root, root) %{_sysconfdir}/aziot/config.toml.template
 
 # Home directories
 %attr(-, aziotcs, aziotcs) %dir /var/lib/aziot/certd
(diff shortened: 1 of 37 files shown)


daprilik (approved):

contrib/centos/aziot-identity-service.spec
```diff
@@ -174,7 +174,7 @@ fi
 %attr(400, aziottpm, aziottpm) %{_sysconfdir}/aziot/tpmd/config.toml.default
 %attr(700, aziottpm, aziottpm) %dir %{_sysconfdir}/aziot/tpmd/config.d
 
-%attr(400, root, root) %{_sysconfdir}/aziot/config.toml.template
+%attr(700, root, root) %{_sysconfdir}/aziot/config.toml.template
```
daprilik, line 177: Wait, shouldn't this be 600?

gordonwang0:
Regarding the duplicate principals:

I think it's better to allow `iotedge config apply` to generate a principals list with duplicate uids and modify the `principal_to_map` function (in keyd/certd lib.rs) to dedup.

gordonwang0 (changes requested):
See comment above

arsing:
True, that would make patching easier too.

contrib/centos/aziot-identity-service.spec
arsing: Ah yes, I fixed it in iotedge but forgot to fix it here. Fixed now.

gordonwang0 (approved):

PR #138, opened 2021-03-02
Update docs on packaging to match latest.

micahl:
Some of the content appears to be out-of-sync with the current reality on the device when the packages are installed. This PR is an attempt to bring that up-to-date.

### DIFF
--- a/docs/packaging.md
+++ b/docs/packaging.md
@@ -1,12 +1,12 @@
 # Packaging
 
-The Identity Service (IS), Keys Service (KS) and Certificates Service (CS) have been explicitly designed to not be Azure IoT Edge-specific; they can be used on non-Edge Azure IoT devices also. To this end, we want to cleanly separate these components from existing association with IoT Edge in the form of hosting it in a separate source repository and shipping it as a separate package. Furthermore, the existing IoT Edge daemon will be renamed (`iotedged` -> `aziot-edged`) and modified to depend on these new components for provisioning the device, managing module identities, and managing cryptographic keys and certificates. `aziot-edged`'s only responsibility will be to act as a Module Runtime (MR) for containerized Edge modules.
+The Identity Service (IS), Keys Service (KS), Certificates Service (CS), and TPM Service (TPMS) have been designed to function as stand-alone components and can be used on Linux-based Azure IoT devices as well as Azure IoT Edge devices. To this end, we host these components in a separate source repository and ship them as a separate package. Furthermore, the existing IoT Edge daemon has been renamed (`iotedged` -> `aziot-edged`) and modified to depend on these new components for provisioning the device, managing module identities, and managing cryptographic keys and certificates. `aziot-edged`'s responsibility is to act as a Module Runtime (MR) for containerized Edge modules.
 
-Because of these large-scale changes, we plan to make the current `iotedge` Linux package into a long-term servicing (LTS) release. The new components (IS/KS/CS), along with the refactoring for an `aziot-edged` with smaller responsibilities, will be shipped in two new lines of packages:
-
-- `aziot-identity-service`: This package contains the IS, KS and CS components.
+Because of these large-scale changes, we've made the current `iotedge` Linux package into a long-term servicing (LTS) release. The new components (IS/KS/CS/TPMS), along with the refactoring for an `aziot-edged` with smaller responsibilities, is shipped in two new lines of packages:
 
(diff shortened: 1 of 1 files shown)


docs/packaging.md
```diff
@@ -1,13 +1,13 @@
...
 
-Because of these large-scale changes, we plan to make the current `iotedge` Linux package into a long-term servicing (LTS) release. The new components (IS/KS/CS), along with the refactoring for an `aziot-edged` with smaller responsibilities, will be shipped in two new lines of packages:
-
-- `aziot-identity-service`: This package contains the IS, KS and CS components.
+Because of these large-scale changes, we've made the current `iotedge` Linux package into a long-term servicing (LTS) release. The new components (IS/KS/CS/TPMS), along with the refactoring for an `aziot-edged` with smaller responsibilities, is shipped in two new lines of packages:
 
 - `aziot-edge`: This package contains the MR component needed to deploy dockerized Edge modules. It will have a dependency on the `aziot-identity-service` package, and on the Moby runtime package.
 
+- `aziot-identity-service`: This package contains the IS, KS and CS components.
```
arsing, line 9: Revert this reordering.

docs/packaging.md
```diff
@@ -87,25 +89,30 @@ The `iotedge` CLI tool is used to interact with the `aziot-edged` service.
...
+
 </td>
 <td>
 
 ```
 /etc/aziot/
+├── config.toml.edge.template
 └── edged/
-    └── config.yaml
+    └── config.toml
```
arsing, line 112: `config.toml.default`

docs/packaging.md
```diff
@@ -87,25 +89,30 @@ The `iotedge` CLI tool is used to interact with the `aziot-edged` service.
...
 ```
 
-Note that the configuration is still in YAML format, but reduced in scope.
+The `config.toml.edge.template` is a template for a super `config.toml` that additionally includes configuration values specific to IoT Edge. The role of that super `config.toml` replaces the `config.yaml` that was previously used. It will be common to use a copy of the template as the super `/etc/aziot/config.toml`. The `iotedge config apply` wraps `aziotctl config apply` in addition to generating the individual `config.toml` for the `aziot-edged` service.
```
arsing, line 115: This paragraph belongs in the iotedge repo. We should only talk about `config.toml.template` here.

docs/packaging.md
```diff
@@ -247,12 +258,10 @@ Note that the configuration is still in YAML format, but reduced in scope.
 ```sh
 apt install aziot-edge
 
-iotedge init
+iotedge config
```
arsing, line 261: All this should be moved to the iotedge repo too. We should only talk about `aziotctl config ...` here.

docs/packaging.md
```diff
@@ -305,7 +313,10 @@ The IS+KS+CS+TPMS components can still be installed as a standalone package on d
 ```sh
 apt install aziot-identity-service
 
-aziotctl config wizard
-```
+sudo cp /etc/aziot/config.toml.template /etc/aziot/config.toml
 
-After installing the package, run `aziotctl config wizard` to interactively set up the configuration with minimal information like the device provisioning method.
+# set device provisioning information
+sudo nano /etc/aziot/config.toml
```
arsing, line 319: `nano -w` so that it doesn't break long lines on saving.

arsing (changes requested):

docs/packaging.md
micahl: Why not move this entire document to the iotedge repo?

arsing: The users of the a-i-s package need documentation for the contents of the a-i-s package.

micahl: If you're agreed, I'll put a copy of this doc in the iotedge repo with most of the content that is currently here.  And then this copy I'll strip of anything that isn't only for this package which would mean it becomes most of the first paragraph with the middle column of the table (only), and then the very last section about installing a-i-s.  Sound good?

arsing: re: what will be left in this file, yes. The `aziotctl config wizard` stuff is also wrong and should talk about `aziotctl config apply`, but I'll fix that as part of my PR to remove `wizard`.

re: putting a copy of this in the iotedge repo, whether you want to do that or link to this one when talking about the common parts (to reduce duplication) is up to you.

arsing: I went ahead and included the fixes you made here in https://github.com/Azure/iot-identity-service/pull/145 along with getting rid of `aziotctl config wizard`. Please review that. For this PR, you can close it and reopen against the iotedge repo.

PR #140, opened 2021-03-02
Also initialize an env_logger::Logger in libaziot-keys.

arsing:
It's compiled as a cdylib so it doesn't share the global logger of
the host process (keyd).

Ref: #120


daprilik (approved):
Nice catch!

gordonwang0 (approved):

PR #142, opened 2021-03-03
Add top-level license information for dependencies

massand:


arsing (changes requested):
As I told you before, this file is different for each target.

Have you verified the file is the same for all targets? Otherwise there's no one single file that you can check in.

*If* it so happens that the file is the same for all targets, then this needs a) a test to ensure that all targets generate the same file, and b) that the checked-in file is always up-to-date.

massand:
@arsing Good point. I see all our supported targets are generating the same NOTICE file, but that could diverge in the future. It is a bit of a pain that `cargo metadata` doesn't provide a range of platforms to be passed into `--filter-platform`, which would make our lives easier. But, it doesn't sound like a super long process to merge all dependencies from all our supported targets into one. 

One approach to achieve the license precheck is to split the third-party-notices script into two - one that builds the dependencies list per target (like current script) and another that merges dependencies for all supported targets. Both these scripts can then pass the dependencies list into another common script that pulls together the license files.

arsing:
Yes, that'll work, but you only need two scripts, not three. Make `get-dependencies.sh <TARGET>` that emits multiple lines of `"\(.id)|\(.license)|\(.manifest_path)|\(.repository)"`, and `third-party-notices.sh` that reads those lines from stdin and writes the TPN to stdout. Then the TPN for the package build is generated by `get-dependencies.sh $target | third-party-notices.sh > TPN` and the file you want to check in is generated by `(get-dependencies.sh target1 && get-dependencies.sh target2 && get-dependencies.sh target3) | sort | uniq | third-party-notices.sh > TPN`. Do this in `make test-release`

And like I said, copy the existing `git status --porcelain` command in `make test-release` to also do the same "uncommitted modifications" check for the checked-in TPN file.

PR #143, opened 2021-03-03
Fix UB when accessing CK_FUNCTION_LIST of unsupported PKCS#11 implementations.

arsing:


daprilik (commented):
Wouldn't this still be UB since the code _after_ the version check is still using the same pointer from _before_ the version check?

i.e: I forget whether or not the compiler preserves pointer "origin" between conversions.

I think the more correct thing to do would be to change the return type of `C_GetFunctionList` to something `*const pkcs11_sys::CK_VERSION` directly, and then do `let function_list = <version pointer> as _` after validating the version.

That said, I'm not 100% certain about this, so if you're confident with it, I can approve.

arsing:
The only thing that Miri complains about is accessing out of the bounds of the allocation. So it complains about the `*function_list` in `(*function_list).version`, because it interprets that as accessing the entire range of `CK_FUNCTION_LIST`, even though in the end the `*function_list` was just a gateway to getting its `.version`

It's fine with `*(function_list as *const CK_VERSION)` because the derefence still occurs within the bounds of the (potentially) smaller allocation.

It doesn't actually care if you do [this](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=2c9454c4c53ef8697938f14099f40648) for example, even though the dereference yields garbage in the `minor` field, because the garbage is not read. But [as soon as you read the garbage,](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=e898dc3dd93006b348de892b2f2aa363) it complains again.

daprilik (approved):
Ahhh, interesting...

If it were me, I'd rewrite it just to be extra safe / explicit, but sure, this ought to be fine as well.

PR #144, opened 2021-03-03
Change PACKAGE_VERSION to 1.2.0~rc4

gordonwang0:
For consistency with the aziot-edge package, change the package version and revision to `1.2.0~rc4` and `1`, respectively.

### DIFF
--- a/.github/workflows/packages.yaml
+++ b/.github/workflows/packages.yaml
@@ -45,8 +45,8 @@ jobs:
           '/src/ci/package.sh'
       env:
         ARCH: "${{ matrix.arch }}"
-        PACKAGE_VERSION: '1.2.0'
-        PACKAGE_RELEASE: 'rc4'
+        PACKAGE_VERSION: '1.2.0~rc4'
+        PACKAGE_RELEASE: '1'
     - name: 'Generate artifact properties'
       id: 'generate-artifact-properties'
       run: |


arsing (changes requested):
This is wrong. Revision and release are separated by `-`, not `~`.

Makefile
```diff
@@ -245,7 +245,7 @@ dist:
 deb: contrib/debian/*
 deb: dist
 	# Rename dist tarball to .orig.tar.gz as the source format version "3.0 (quilt)" requires
-	mv /tmp/aziot-identity-service-$(PACKAGE_VERSION).tar.gz /tmp/aziot-identity-service_$(PACKAGE_VERSION).orig.tar.gz
+	mv /tmp/aziot-identity-service-$(PACKAGE_VERSION).tar.gz /tmp/aziot-identity-service_$(PACKAGE_VERSION)~$(PACKAGE_RELEASE).orig.tar.gz
```
arsing, line 248: This is also wrong. The source tarball does not contain the release identififer either.

arsing (changes requested):

arsing:
>For consistency with the aziot-edge package, change the package version/revision separator from - to ~.

But the aziot-edge package also uses `-`. I'm very confused what this PR is trying to achive.

Makefile
gordonwang0: Fixed. The rpm tarball below does contain the revision though, should I change that so that it doesn't or leave as is?

arsing:
I saw the email thread related to this and have responded there. Please update your PR with the fix as I described there - there is no changes needed for anything except the default in packages.yaml

Makefile
arsing: It's a matter of conventions. Debian requires the release ident to not be there (IIRC). rpm doesn't care either way because you tell it what the tarball name is, and the convention is to include the release ident.

arsing (approved):

PR #145, opened 2021-03-04
Remove `aziotctl config wizard`

arsing:
We determined such a prompt-based wizard is too hard to use. For now,
we're shipping the super-config approach used with `aziotctl config apply`,
and are looking to add more focused wizards like `aziotctl config mp` and
`aziotctl config dps` for manual and DPS provisioning respectively
in the future.

As part of removing the mentions of `aziotctl config wizard`
from `packaging.md`, this commit also removes the IoT Edge-specific text from
that document. This text will be moved to a matching document in
the iotedge repo and be maintained there.

Co-authored-by: Micah Lewis <1009123+micahl@users.noreply.github.com>


docs/packaging.md
```diff
@@ -1,155 +1,86 @@
...
 
-Because of these large-scale changes, we plan to make the current `iotedge` Linux package into a long-term servicing (LTS) release. The new components (IS/KS/CS), along with the refactoring for an `aziot-edged` with smaller responsibilities, will be shipped in two new lines of packages:
+On Azure IoT Edge devices, these services form the backbone of the Azure IoT Edge service - the Azure IoT Edge service relies on the Identity Service to provision the Edge device identity and Edge module identities in Azure, to provide the CA certificate used for module server certificates like Edge Hub's, and so on.
 
-- `aziot-identity-service`: This package contains the IS, KS and CS components.
+These components ship in the `aziot-identity-service` package. This document describes the contents of this package and how to install and configure it.
```
micahl, line 7: In the respective 'Packaging' doc I'm adding to the iotedge repo I cross-link back to this one.  Suggest doing something similar here.

docs/packaging.md
```diff
@@ -158,154 +89,74 @@ Note that the configuration is still in YAML format, but reduced in scope.
...
 
-Alternatively, a non-interactive initialization that attempts to import the old configuration from `iotedge` can be done using the `import` subcommand.
 
-```sh
-iotedge init import
-```
+## Package dependencies
 
-- Device provisioning method is parsed from `iotedge/config.yaml` and translated into the provisioning information in `identityd/config.toml`, `keyd/config.toml`, `certd/config.toml`, `tpmd/config.toml` and `edged/config.yaml`. For example, in case of manual-symmetric-key provisioning, the SAS key will be imported as a preloaded key in `keyd/config.toml`, and `identityd/config.toml` will be updated to use manual provisioning with a reference to the key ID.
+The binaries and libraries shipped in this package only depend on OpenSSL.
```
micahl, line 134: Might be worth calling out which version of openssl.

docs/packaging.md
```diff
@@ -1,155 +1,86 @@
...
 
-```
-/usr/
-├── bin/
-│   └── iotedge
-└── libexec/
-    └── aziot/
-        └── aziot-edged
-```
+`/usr/lib/libaziot_keys.so` is a library used by the KS for pluggable cryptographic key storage. It exposes a simple C API and is designed to be replacable by the user if they need to. However the implementation shipped by default is already quite versatile, since it supports both filesystem keys as well as all HSMs with PKCS#11 libraries, so we don't expect users to need to replace it.
```
micahl, line 34: nit: typo "replacable"

arsing: Fixed.

docs/packaging.md
arsing: Done.

docs/packaging.md
arsing: What's the URL?

arsing: nvm, saw your other PR.

micahl (approved):

daprilik (approved):
⚰️

daprilik:
@arsing just noticed the branch name :mage:

PR #146, opened 2021-03-04
Misc. documentation fixes

### LINKED ISSUE #131: Missing `aziot-tpmd` for key 'endpoints` when trying to launch identity service
The [doc](https://azure.github.io/iot-identity-service/dev/running/aziot-identityd.html) doesn't mention this field. What should be configured there?

daprilik:
- Add references to tpmd in the `identityd` docs regarding the `[endpoints]` table (closes #131)
- Remove the comment `# Optional for X.509 attestation` on `registration_id` when using X509 in `identityd` docs (closes #123)
- Update the "Code Organization" sections of tpmd and keyd
- Update tpmd docs to include a note specifying that it is not externally accessible


docs/dev/running/aziot-identityd.md
```diff
@@ -12,10 +12,6 @@ homedir = "/var/lib/aziot/identityd"
 
 [provisioning]
 
-[endpoints]
```
arsing, line 15: Every service documents them because these are dev docs. So if you're going to remove this here you need to remove it from all the others as well.

But these are dev docs, and the expected reader is us devs. I suggest restoring it for now, with the entry added for tpmd to fix that GH issue. Then we need to figure out a way so that non-devs don't find these docs under the dev/ directory. Hide them from the default sidebar unless you're under dev/ or something, perhaps.

arsing (changes requested):

docs/dev/running/aziot-identityd.md
daprilik: Fair enough. In that case, I'll just re-add `[endpoints]` to the example config + add a `tpmd` entry as well.

arsing (approved):
Update the PR before you merge.

PR #147, opened 2021-03-04
Add postrm to aziot-identity-service

gordonwang0:
Adds a script so that apt purge fully removes all files used by aziot-identity-service and removes the services' users.

This only applies to Debian, as CentOS (rpm) doesn't have a purge command.


contrib/debian/postrm
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
```
arsing, line 1: How did you generate this file?

contrib/debian/postrm
```diff
@@ -0,0 +1,50 @@
...
+			for service in "${services[@]}"; do
+				rm -rf "/etc/aziot/${service}"
+			done
+
+			if [ -z "$(ls -A /etc/aziot)" ]; then
+				rm -rf /etc/aziot
+			fi
+		fi
+
+		directories=("/run/aziot" "/var/secrets/aziot")
```
arsing, line 27: `/var/secrets/aziot` doesn't belong to us. It contains user files.

contrib/debian/postrm
gordonwang0: I took the [postrm file in edgelet](https://github.com/Azure/iotedge/blob/master/edgelet/contrib/debian/postrm) and modified that.

contrib/debian/postrm
gordonwang0: Fixed.

contrib/debian/postrm
arsing: Generate it properly with `dh_make` please. iotedge uses an old package format, but more importantly your version is a) using the wrong shebang, and b) missing the `DEBHELPER` comment that the builder uses to substitute additional commands.

gordonwang0: I regenerated the file with `dh_make`. Note that I kept the shebang `#!/bin/bash` because this script uses arrays.

arsing: Okay, I assume you made it bash because you want to use an array for deleting the users. Just delete them one at a time instead.

gordonwang0: Fixed. Is there a reason that the preinst is using bash in its shebang?

arsing: Huh.

Just to be sure, `dh_make` generated the template to use `sh` right? If so, I guess I changed it, but I'm not sure why I did that. It shouldn't have needed to be changed.

gordonwang0: Yeah, the template (preinst.ex) has `sh`.

arsing: Yeah, I just verified myself too. Can you change that to `sh` too please?

contrib/debian/postrm
```diff
@@ -0,0 +1,63 @@
...
+
+		# Delete directories used by aziot-identity-service.
+		# Home directories are not deleted here; they will be deleted with their user.
+		if [ -d /etc/aziot ]; then
+			rm -rf /etc/aziot/identityd
+			rm -rf /etc/aziot/certd
+			rm -rf /etc/aziot/keyd
+			rm -rf /etc/aziot/tpmd
+
+			if [ -z "$(ls -A /etc/aziot)" ]; then
```
arsing, line 33: What would be left that we wouldn't want to delete? We already deleted any user customizations for the individual services.

contrib/debian/postrm
```diff
@@ -0,0 +1,63 @@
...
+				rm -rf /etc/aziot
+			fi
+		fi
+
+		rm -rf /run/aziot
+
+		# Delete aziot-identity-service users.
+		/usr/sbin/userdel aziotid
```
arsing, line 41: Delete the groups too, then?

contrib/debian/postrm
gordonwang0: Fixed

contrib/debian/postrm
gordonwang0: Yes, you're right. Fixed.

contrib/debian/postrm
gordonwang0: On my machine, this command deletes the groups too, since `USERGROUPS_ENAB` is set in `/etc/login.defs`. This seems to be the default behavior.

arsing: Oh, TIL.

arsing (approved):

PR #148, opened 2021-03-04
Cherry-pick packaging changes

gordonwang0:


gordonwang0:
Looking at the history between release/1.2 and master, it seems like it's ok to import all the changes from master to release/1.2. I can do a git reset of release/1.2 to master.

arsing:
Did we already release rc4? So that we don't have to worry about adding more untested changes to it.

If yes, then I'm fine with `reset`ing it.

daprilik:
(I too am in favor of simply resetting the branch)

gordonwang0:
Aside from these packaging changes, resetting would bring in the some doc fixes and 347a1e6f2b61e50978721b8d9c25bde0af61f684. So I think it's fine to reset. We'll need a new rc4 package anyways because we need the package versioning update here.

PR #150, opened 2021-03-05
Add proxy support for EST client

massand:
This change introduces setting the proxy information in `certd` through an environment variable (similar to `identityd`), since the EST client in `certd` service could use the proxy to connect to an EST server.


arsing:
(CI failure can be fixed by changing that parameter to `Option<(&openssl::pkey::PKey<openssl::pkey::Private>, &[u8])>` and changing all the `match`es from `match &identity { ... }` to `match identity { ... }` )

PR #151, opened 2021-03-05
[tpmd] Pass through AZIOT_LOG level to C code

daprilik:


tpm/aziot-tpm-rs/src/tpm.rs
```diff
@@ -89,7 +90,14 @@ impl Tpm {
         // ensure that `aziot_tpm_init` is only called once
         INIT_C_LIB.call_once(|| unsafe {
             INIT_RESULT = {
-                let result = aziot_tpm_init() as isize;
+                let log_level = match log::max_level() {
+                    l if l <= log::Level::Error => LOG_LVL_ERROR,
```
arsing, line 94: For the first arm, `<=` will cause `LevelFilter::Off` to be converted to `LOG_LVL_ERROR`, whereas it ought to be converted to `LOG_LVL_INFO` by the `_` case.

For each of the other arms, `<=` is redundant with the arms before them. Eg the only way `l if l <= log::Level::Debug` will be hit is if `l == log::level::Debug`, because `l < log::Level::Debug` was already handled by the previous arms.

So I think all the `<=` should be `==`

daprilik: Yep, fair enough. In hindsight, `==` seems to be a lot easier and less bug-prone, so I'll just switch 'em out.

arsing (approved):

PR #153, opened 2021-03-09
Fix various issues found while testing with tpm2-pkcs11.

arsing:
- Document Debian arm32v7 location of self-compiled `tpm2-pkcs11 .so`

- Configure tpm2-pkcs11 with `--enable-debug=info`. The default is
  `--enable-debug=yes`, which cases any `CKR_FUNCTION_NOT_SUPPORTED`-returning
  unimplemented function to call `assert(0)` and `abort()`. We want to handle
  this error ourselves.

- Also treat `CKR_FUNCTION_NOT_SUPPORTED` from `C_GenerateKey` as
  a fallible error, so that creating a symmetric key can fall back to
  the filesystem.

- Fix typo in "C_GenerateKey failed with" error message.

### DIFF
--- a/docs/dev/running/pkcs11/index.md
+++ b/docs/dev/running/pkcs11/index.md
@@ -45,6 +45,7 @@ Follow the steps corresponding to your hardware:
         ```sh
         # Path of the PKCS#11 library
         export PKCS11_LIB_PATH='/usr/local/lib/libtpm2_pkcs11.so'
+        # export PKCS11_LIB_PATH='/usr/lib/arm-linux-gnueabihf/pkcs11/libtpm2_pkcs11.so'
 
         # Variables to identify and log in to the PKCS#11 token
         TOKEN='Key pairs'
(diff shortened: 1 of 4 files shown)


docs/dev/running/pkcs11/index.md
```diff
@@ -45,6 +45,7 @@ Follow the steps corresponding to your hardware:
         ```sh
         # Path of the PKCS#11 library
         export PKCS11_LIB_PATH='/usr/local/lib/libtpm2_pkcs11.so'
+        # export PKCS11_LIB_PATH='/usr/lib/arm-linux-gnueabihf/pkcs11/libtpm2_pkcs11.so'
```
massand, line 48: Is this intentionally left here for reference for tpm2-pkcs11 library only? Are any other libraries following a similar base path `/usr/lib/arm...`?

massand (approved):

docs/dev/running/pkcs11/index.md
arsing: Yeah it's just an example, similar to the one on line 23. It isn't meant to be comprehensive or anything; in the end the user runs `make install` / `apt install` / whatever so the user knows where the library is.

PR #154, opened 2021-03-09
Enable `hyper::Client` to talk to HTTP/2 servers.

arsing:
- Enable the `http2` feature of the `hyper` crate.

- Fix `MaybeProxyStream::connected()` to correctly forward to the inner stream's
  `connected()`. In particular, the inner `hyper_openssl::MaybeHttpsStream`
  notices that the server negotiated HTTP/2 via ALPN and creates a `Connected`
  with that bit set.

  Before this change, the old code was ignoring the `Connected` created by
  `hyper_openssl::MaybeHttpsStream` and always creating a new one.
  Thus the inner `hyper::Client` would attempt to parse the server's
  HTTP/2 response as HTTP/1.1 and fail.

- Remove code that disabled HTTP/2 in the ALPN request that was added as
  a workaround in 8630d84b4615fe0f3efc9a31dd50a9f225cc923b

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1119,6 +1119,31 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "f6503fe142514ca4799d4c26297c4248239fe8838d827db6bd6065c6ed29a6ce"
 
 [[package]]
+name = "h2"
+version = "0.3.1"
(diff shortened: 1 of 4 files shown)


massand (approved):
Thanks for removing the extra clone I just checked in!

PR #157, opened 2021-03-10
Default always_reprovision_on_startup to true

gordonwang0:
Default `always_reprovision_on_startup` in Identity Service configuration to true, as recommended by DPS documentation.

Also updates the template config file for the latest IS changes.


arsing:
But why? We have more users complaining about unnecessary reprovisions than not.

PR #159, opened 2021-03-15
Add `aziotctl config mp` to create a super-config with a manual-provisioning connection string.

arsing:
Other changes:

- Remove leftover TODO in super_config.rs

- Remove another piece of code that disabled HTTP/2 that
  38075501330c10516312699ebbf2b657ebb15b58 forgot to revert.

### DIFF
--- a/aziotctl/src/config/mp.rs
+++ b/aziotctl/src/config/mp.rs
@@ -24,7 +24,7 @@ pub(crate) struct Options {
     out_config_file: std::path::PathBuf,
 
     /// Overwrite the new configuration file if it already exists
-    #[structopt(short = "f", long, value_name = "FILE")]
+    #[structopt(short = "f", long)]
     force: bool,
 }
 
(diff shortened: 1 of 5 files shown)


aziotctl/src/config/mp.rs
```diff
@@ -0,0 +1,100 @@
...
+    #[structopt(
+        short = "o",
+        long,
+        value_name = "FILE",
+        default_value = "/etc/aziot/config.toml"
+    )]
+    out_config_file: std::path::PathBuf,
+
+    /// Overwrite the new configuration file if it already exists
+    #[structopt(short = "f", long, value_name = "FILE")]
```
gordonwang0, line 27: Should the `value_name` be something else? Like `FORCE`?

arsing: Oops. Fixed. (It's not supposed to have a `value_name` since it's a flag.)

massand (approved):

gordonwang0 (approved):

PR #160, opened 2021-03-15
Add check up_to_date_configs

gordonwang0:
This check verifies that the super config files (`/etc/aziot/*/config.d/00-super.toml`) have a `last_modified` after `/etc/aziot/config.toml`. Prints a warning if `config.toml` was modified after `00-super.toml`, since that means the user probably forgot to run `aziotctl config apply` after updating `config.toml`.

Example message:
```
‼ daemon configurations up-to-date with config.toml - Warning
    /etc/aziot/config.toml was modified after keyd's config
    You must run 'aziotctl config apply' to update keyd's config with the latest config.toml
```

### DIFF
--- a/aziotctl/aziotctl-common/src/check_last_modified.rs
+++ b/aziotctl/aziotctl-common/src/check_last_modified.rs
@@ -1,0 +1,62 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+use std::fs;
+use std::path::Path;
+
+pub enum CheckResult {
+    Ok,
+    Ignored,
+    Warning(String),
+    Failed(std::io::Error),
+}
+
+pub fn check_last_modified(services: &[&str]) -> CheckResult {
+    let config = Path::new("/etc/aziot/config.toml");
+
+    if !config.exists() {
+        return CheckResult::Ignored;
+    }
+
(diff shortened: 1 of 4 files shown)


aziotctl/aziotctl-common/src/check_last_modified.rs
```diff
@@ -0,0 +1,62 @@
...
+        .expect("file metadata should contain valid last_modified");
+
+    for service in services {
+        let service_config = format!("/etc/aziot/{}/config.d/00-super.toml", service);
+        let service_config = Path::new(&service_config);
+
+        if !service_config.exists() {
+            return CheckResult::Warning(format!(
+                "{} does not exist.\n\
+                Did you run 'iotedge config apply'?",
```
arsing, line 36: It's not `iotedge` when you run it from `aziotctl`. Use `aziotctl_common::program_name()`

Same for the other `Warning`

arsing (changes requested):

aziotctl/aziotctl-common/src/check_last_modified.rs
gordonwang0: Thanks, fixed.

arsing (approved):

PR #163, opened 2021-03-17
Add aziotctl system reprovision command

gordonwang0:
Adds the `aziot system reprovision` command, which triggers the `/reprovision` API of Identity Service.

Also adds an async client for Identity Service. Currently only supports the `reprovision` API, as that's the only one used.

### DIFF
--- a/aziotctl/aziotctl-common/src/reprovision.rs
+++ b/aziotctl/aziotctl-common/src/reprovision.rs
@@ -1,0 +1,23 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+use anyhow::{anyhow, Result};
+
+pub async fn reprovision(uri: &url::Url) -> Result<()> {
+    let connector =
+        http_common::Connector::new(&uri).map_err(|err| anyhow!("Invalid uri {}: {}", uri, err))?;
+    let client = aziot_identity_client_async::Client::new(
+        aziot_identity_common_http::ApiVersion::V2020_09_01,
+        connector,
+    );
+
+    match client.reprovision().await {
+        Ok(_) => {
+            println!("Successfully reprovisioned with IoT Hub.");
+            Ok(())
+        }
+
+        Err(err) => {
+            return Err(anyhow!("Failed to reprovision: {}", err));
+        }
+    }
+}
--- a/aziotctl/aziotctl-common/src/reprovision.rs
+++ b/aziotctl/aziotctl-common/src/reprovision.rs
@@ -16,8 +16,6 @@ pub async fn reprovision(uri: &url::Url) -> Result<()> {
             Ok(())
         }
 
-        Err(err) => {
-            return Err(anyhow!("Failed to reprovision: {}", err));
-        }
+        Err(err) => Err(anyhow!("Failed to reprovision: {}", err)),
     }
 }
--- a/aziotctl/aziotctl-common/src/reprovision.rs
+++ b/aziotctl/aziotctl-common/src/reprovision.rs
@@ -4,7 +4,7 @@ use anyhow::{anyhow, Result};
 
 pub async fn reprovision(uri: &url::Url) -> Result<()> {
     let connector =
-        http_common::Connector::new(&uri).map_err(|err| anyhow!("Invalid uri {}: {}", uri, err))?;
+        http_common::Connector::new(uri).map_err(|err| anyhow!("Invalid URI {}: {}", uri, err))?;
     let client = aziot_identity_client_async::Client::new(
         aziot_identity_common_http::ApiVersion::V2020_09_01,
         connector,
--- a/aziotctl/src/system.rs
+++ b/aziotctl/src/system.rs
@@ -44,10 +44,11 @@ pub struct LogLevelOptions {
 #[derive(StructOpt)]
 #[structopt(about = "Reprovision device with IoT Hub")]
 pub struct ReprovisionOptions {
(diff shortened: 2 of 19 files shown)


aziotctl/src/system.rs
```diff
@@ -38,12 +41,24 @@ pub struct LogLevelOptions {
     log_level: log::Level,
 }
 
-pub fn system(options: Options) -> Result<()> {
+#[derive(StructOpt)]
+#[structopt(about = "Reprovision device with IoT Hub")]
+pub struct ReprovisionOptions {
+    #[structopt(
+        value_name = "Identity Service URI",
+        long = "--uri",
```
arsing, line 49:
```diff,rust
-long = "--uri",
+long = "uri",
```

aziotctl/aziotctl-common/src/reprovision.rs
```diff
@@ -0,0 +1,23 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+use anyhow::{anyhow, Result};
+
+pub async fn reprovision(uri: &url::Url) -> Result<()> {
+    let connector =
+        http_common::Connector::new(&uri).map_err(|err| anyhow!("Invalid uri {}: {}", uri, err))?;
```
arsing, line 7: `&uri` -> `uri`

`"Invalid uri` -> `"Invalid URI`

http-common/src/lib.rs
```diff
@@ -26,6 +26,9 @@ mod proxy;
 #[cfg(feature = "tokio1")]
 pub use proxy::{get_proxy_uri, MaybeProxyConnector};
 
+mod request;
+pub use request::{request, request_no_content};
```
arsing, line 30: These need to be under `#[cfg(feature = "tokio1")]`

identity/aziot-identityd/src/lib.rs
```diff
@@ -383,6 +383,18 @@ impl Api {
                     .await?
             }
             ReprovisionTrigger::Api => {
+                // Clear the backed up device state before reprovisioning.
+                // If this fails, log a warning but continue with reprovisioning.
+                let mut backup_file = self.settings.homedir.clone();
+                backup_file.push("device_info");
+
+                if let Err(err) = std::fs::remove_file(backup_file) {
+                    log::warn!(
```
arsing, line 392: Need to treat it as successful if `err.kind() == std::io::ErrorKind::NotFound`

arsing (changes requested):

aziotctl/src/system.rs
arsing: Actually, this shouldn't even be a parameter. The user doesn't get to override the endpoints anyway.

gordonwang0: True, but it's useful for development. For example, on my test VM I'm not using the default endpoints during development. This lets me use the reprovision command on my dev endpoints.

aziotctl/aziotctl-common/src/reprovision.rs
gordonwang0: Fixed.

http-common/src/lib.rs
gordonwang0: Fixed

identity/aziot-identityd/src/lib.rs

aziotctl/src/system.rs
arsing: My point is it shouldn't be a parameter of the CLI that the user can set.

Either `#[cfg(debug_assertions)]` it, or read it from an env var.

aziotctl/src/system.rs
```diff
@@ -38,12 +41,24 @@ pub struct LogLevelOptions {
...
 }
 
-pub fn system(options: Options) -> Result<()> {
+#[derive(StructOpt)]
+#[structopt(about = "Reprovision device with IoT Hub")]
+pub struct ReprovisionOptions {
+    #[structopt(
+        value_name = "Identity Service URI",
+        long,
+        default_value = "unix:///var/run/aziot/identityd.sock"
```
arsing, line 50: If you're going to keep it as a parameter, remove the `/var` from this to be consistent. All the other endpoints use `/run/aziot/...

aziotctl/src/system.rs
gordonwang0: Fixed

aziotctl/src/system.rs
gordonwang0: Fixed

aziotctl/src/system.rs
```diff
@@ -38,12 +41,24 @@ pub struct LogLevelOptions {
...
+    )]
+    uri: url::Url,
```
massand, line 53: The default URL should be preceded by grabbing the URL from the config.toml file (in case it was overriden there). Right?

gordonwang0: We only allow overriding the default URL for debug builds, so it should be ok to always use "/run/aziot/identityd.sock". Anyways, this option is only available for debug builds too.

massand: Ah yes. Thanks for clarifying.

massand (approved):

aziotctl/src/system.rs
arsing: Also the CLI shouldn't read the service's config file anyway.

arsing (approved):

PR #164, opened 2021-03-17
Adding checks for permission user and group on config files

huguesBouvier:
In the test the "file is well formed"
IotEdge check will now verify:
That the file is read only
That the group is correct
That the user is correct

Test:
Created a dummy file that was not read only, with root as grouID and root as user ID.
Check error was issued for each and that the script was proposing for a way to fix it.
Example of error:
× certd configuration is well-formed - Error
    The file /etc/aziot/certd/config.toml has user root, expected aziotcs. Please run 'sudo chown aziotcs /etc/aziot/certd/config.toml'


aziotctl/Cargo.toml
```diff
@@ -27,6 +27,7 @@ sysinfo = "0.15.3"
 tokio = { version = "1", features = ["macros", "fs", "io-util", "rt-multi-thread"] }
 toml = "0.5"
 url = "2"
+users = "0.11"
```
arsing, line 30: Use the `nix` crate that we already use.

arsing (changes requested):

aziotctl/Cargo.toml
huguesBouvier: Thanks! I fixed it.

aziotctl/src/internal/check/checks/well_formed_configs.rs
```diff
@@ -171,14 +189,71 @@ impl Checker for WellFormedIdentitydConfig {
...
 
 async fn load_daemon_cfg<T: serde::de::DeserializeOwned>(
     daemon: &str,
+    uid: &str,
     config_path: &Path,
     config_directory_path: Option<&Path>,
     shared: &CheckerShared,
 ) -> Result<DaemonCfg<T>> {
+    use std::os::linux::fs::MetadataExt;
+    if let Ok(metadata) = fs::metadata(config_path) {
```
arsing, line 205: This is wrong for two reasons:

1. `config_path` doesn't necessarily exist. In fact the situation that it doesn't exist is the default - `config apply` does not create it.

1. None of the files in `config_directory_path` are being checked.

aziotctl/src/internal/check/checks/well_formed_configs.rs
```diff
@@ -4,7 +4,8 @@ use anyhow::{Error, Result};
 use serde::Serialize;
 
 use crate::internal::check::{CheckResult, Checker, CheckerCache, CheckerMeta, CheckerShared};
-
+use nix::unistd::{Gid, Group, Uid, User};
```
arsing, line 7: Move this to be in the group of third-party imports, ie with `anyhow` and `serde` above. Also make sure to sort it, ie put it between `anyhow` and `serde`.

aziotctl/src/internal/check/checks/well_formed_configs.rs
```diff
@@ -171,14 +189,71 @@ impl Checker for WellFormedIdentitydConfig {
 enum DaemonCfg<T> {
     Cfg(T),
     PermissionDenied(Error),
+    IncorrectPermissions(Error),
+    IncorrectGroupId(Error),
+    IncorrectUserID(Error),
```
arsing, line 194: `IncorrectUserId` - lower-case `d`

arsing, line 194: Sort enum variants in alphabetical order.

aziotctl/src/internal/check/checks/well_formed_configs.rs
```diff
@@ -171,14 +189,71 @@ impl Checker for WellFormedIdentitydConfig {
...
 }
 
 async fn load_daemon_cfg<T: serde::de::DeserializeOwned>(
     daemon: &str,
+    uid: &str,
```
arsing, line 199: `uid` is a `nix::unistd::Uid`, not a `&str`. A `&str` is a username.

aziotctl/src/internal/check/checks/well_formed_configs.rs
```diff
@@ -171,14 +189,71 @@ impl Checker for WellFormedIdentitydConfig {
...
+            Ok(Some(group)) => group.name,
+            Err(_) | Ok(None) => {
+                return Err(anyhow::anyhow!(format!(
+                    "Could not find user group assigned to {}.",
+                    config_path.to_string_lossy()
+                )))
+            }
+        };
+
+        if !user.eq(uid) {
```
arsing, line 234: Compare by UID, not by username.

aziotctl/src/internal/check/checks/well_formed_configs.rs
```diff
@@ -171,14 +189,71 @@ impl Checker for WellFormedIdentitydConfig {
...
 async fn load_daemon_cfg<T: serde::de::DeserializeOwned>(
     daemon: &str,
+    uid: &str,
     config_path: &Path,
     config_directory_path: Option<&Path>,
     shared: &CheckerShared,
 ) -> Result<DaemonCfg<T>> {
+    use std::os::linux::fs::MetadataExt;
+    if let Ok(metadata) = fs::metadata(config_path) {
+        if !metadata.permissions().readonly() {
```
arsing, line 206: Why? What's the problem if the file is writable?

aziotctl/src/internal/check/checks/well_formed_configs.rs
```diff
@@ -171,14 +189,71 @@ impl Checker for WellFormedIdentitydConfig {
...
+                "The file {} has user {}, expected {}. Please run 'sudo chown {} {}'",
+                config_path.to_string_lossy(),
+                user,
+                uid,
+                uid,
+                config_path.to_string_lossy()
+            )));
+        }
+
+        if !group.eq(uid) {
```
arsing, line 245: Compare by GID, not by group name.

aziotctl/src/internal/check/checks/well_formed_configs.rs
```diff
@@ -171,14 +189,71 @@ impl Checker for WellFormedIdentitydConfig {
...
+            return Ok(DaemonCfg::IncorrectGroupId(anyhow::anyhow!(
+                "The file {} has group {}, expected {}. Please run 'sudo chgrp {} {}' ",
+                config_path.to_string_lossy(),
+                group,
+                uid,
+                uid,
+                config_path.to_string_lossy()
+            )));
+        }
+    }
```
arsing, line 255: Why does it matter what uid and gid the file is ACLed to? It could be readable to an arbitrary group that the user created which contains the service user. Or it could be owned by root but world-readable. Or anything else.

aziotctl/src/internal/check/checks/well_formed_configs.rs
```diff
@@ -171,14 +189,71 @@ impl Checker for WellFormedIdentitydConfig {
...
+
```
arsing, line 256: Do the same thing as: https://github.com/Azure/iot-identity-service/blob/18f5e5c92a68d08f37ddafeceb16075f1260bcb7/aziotctl/src/config/apply.rs#L41-L69 ie if the current user is root then do an actual check. Otherwise if this is a debug build, allow the file to be owned by the current user.

aziotctl/src/internal/check/checks/well_formed_configs.rs
```diff
@@ -171,14 +189,71 @@ impl Checker for WellFormedIdentitydConfig {
...
+    uid: &str,
     config_path: &Path,
     config_directory_path: Option<&Path>,
     shared: &CheckerShared,
 ) -> Result<DaemonCfg<T>> {
+    use std::os::linux::fs::MetadataExt;
+    if let Ok(metadata) = fs::metadata(config_path) {
+        if !metadata.permissions().readonly() {
+            return Ok(DaemonCfg::IncorrectPermissions(anyhow::anyhow!(
+                "The file {} must be read only. Please run 'sudo chmod 0444 {}'",
```
arsing, line 208: `0400`

arsing (changes requested):

aziotctl/src/internal/check/checks/well_formed_configs.rs
huguesBouvier: Then what can we check? I the PR useful? I set up a meeting to discuss with you.

aziotctl/src/internal/check/checks/well_formed_configs.rs
huguesBouvier: That's straightforward but what's wrong with that?
instead I will have to conver aziotcs (for example) to a uid/gid.

aziotctl/src/internal/check/checks/well_formed_configs.rs
huguesBouvier: What check could we do?

PR #165, opened 2021-03-18
Lift key import out of `get_sas_connector`

daprilik:
This should fix the issue where module identities were not working with DPS-TPM provisioning.
I've also included some light refactoring, which should make the code a bit easier to grok in the future.

### DIFF
--- a/identity/aziot-cloud-client-async-common/src/lib.rs
+++ b/identity/aziot-cloud-client-async-common/src/lib.rs
@@ -11,7 +11,7 @@ use http_common::MaybeProxyConnector;
 pub const ENCODE_SET: &percent_encoding::AsciiSet = &http_common::PATH_SEGMENT_ENCODE_SET.add(b'=');
 
 /// `key_client` must be either an `aziot_key_client_async::Client` or an `aziot_tpm_client_async::Client`.
-pub async fn get_sas_connector<K: KeyClient>(
+pub async fn get_sas_connector<K>(
     audience: &str,
     key_handle: K::KeyHandle,
     key_client: &K,
@@ -20,7 +20,10 @@ pub async fn get_sas_connector<K: KeyClient>(
 ) -> io::Result<(
     MaybeProxyConnector<hyper_openssl::HttpsConnector<hyper::client::HttpConnector>>,
     String,
-)> {
+)>
+where
+    K: KeyClient,
+{
     let token = {
         let expiry = chrono::Utc::now()
             + chrono::Duration::from_std(std::time::Duration::from_secs(30))
@@ -117,8 +120,7 @@ impl KeyClient for aziot_key_client_async::Client {
 
 #[async_trait::async_trait]
 impl KeyClient for aziot_tpm_client_async::Client {
-    // At the moment, we only store a single key in the TPM, so there isn't any
-    // need for distinct key handles.
+    // We only store a single key in the TPM, so there isn't any need for distinct key handles.
     type KeyHandle = ();
 
     async fn sign_with_key(&self, _key_handle: &(), data: &[u8]) -> io::Result<Vec<u8>> {
--- a/identity/aziot-cloud-client-async-common/src/lib.rs
+++ b/identity/aziot-cloud-client-async-common/src/lib.rs
@@ -11,18 +11,16 @@ use http_common::MaybeProxyConnector;
 pub const ENCODE_SET: &percent_encoding::AsciiSet = &http_common::PATH_SEGMENT_ENCODE_SET.add(b'=');
 
 /// `key_client` must be either an `aziot_key_client_async::Client` or an `aziot_tpm_client_async::Client`.
-pub async fn get_sas_connector(
+pub async fn get_sas_connector<K: KeyClient>(
     audience: &str,
-    key_handle: impl AsRef<[u8]>,
-    key_client: &impl KeyClient,
+    key_handle: K::KeyHandle,
+    key_client: &K,
     proxy_uri: Option<hyper::Uri>,
     is_tpm_registration: bool,
 ) -> io::Result<(
     MaybeProxyConnector<hyper_openssl::HttpsConnector<hyper::client::HttpConnector>>,
     String,
 )> {
-    let key_handle = key_client.insert_key(key_handle.as_ref()).await?;
-
     let token = {
         let expiry = chrono::Utc::now()
             + chrono::Duration::from_std(std::time::Duration::from_secs(30))
@@ -95,7 +93,6 @@ mod private {
 pub trait KeyClient: private::Sealed {
     type KeyHandle;
 
-    async fn insert_key(&self, key_handle: &[u8]) -> io::Result<Self::KeyHandle>;
     async fn sign_with_key(&self, key_handle: &Self::KeyHandle, data: &[u8])
         -> io::Result<Vec<u8>>;
 }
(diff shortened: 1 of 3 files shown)


identity/aziot-hub-client-async/src/lib.rs
```diff
@@ -275,6 +276,7 @@ impl Client {
         let body = hyper::body::to_bytes(body)
             .await
             .map_err(|err| std::io::Error::new(std::io::ErrorKind::Other, err))?;
+        log::debug!("IoTHub response body {:?}", body);
```
arsing, line 279: Remove this.

identity/aziot-cloud-client-async-common/src/lib.rs
```diff
@@ -133,12 +117,10 @@ impl KeyClient for aziot_key_client_async::Client {
 
 #[async_trait::async_trait]
 impl KeyClient for aziot_tpm_client_async::Client {
+    // At the moment, we only store a single key in the TPM, so there isn't any
```
arsing, line 120: Remove "At the moment". There'll only ever be one key.

identity/aziot-cloud-client-async-common/src/lib.rs
```diff
@@ -11,18 +11,16 @@ use http_common::MaybeProxyConnector;
 pub const ENCODE_SET: &percent_encoding::AsciiSet = &http_common::PATH_SEGMENT_ENCODE_SET.add(b'=');
 
 /// `key_client` must be either an `aziot_key_client_async::Client` or an `aziot_tpm_client_async::Client`.
-pub async fn get_sas_connector(
+pub async fn get_sas_connector<K: KeyClient>(
```
arsing, line 14: Put it in a `where` clause for consistency.

arsing (approved):
Approved with nits.

identity/aziot-hub-client-async/src/lib.rs
daprilik: I added this to mirror https://github.com/Azure/iot-identity-service/blob/main/identity/aziot-dps-client-async/src/lib.rs#L309.

It proved to be reasonably useful when debugging, so I'd like to keep it in. Alternatively, if you want it gone, we should get rid of the other one as well.

arsing: @massand I vaguely remember you saying we shouldn't log cloud responses that had secrets. Is that correct?

massand: We shouldn't log any responses from IoT Hub due to credential information disclosure issues. DPS doesn't log any keys in its responses for 2 APIs. Maybe adding a comment in DPS client about the responses being clean would help avoid copying that behavior next time.

identity/aziot-dps-client-async/src/lib.rs
```diff
@@ -161,7 +158,7 @@ impl Client {
             .await;
         };
 
-        if matches!(auth_kind, DpsAuthKind::TpmWithAuth { .. }) {
+        if matches!(auth_kind, DpsAuthKind::TpmPostChallenge { .. }) {
```
massand, line 161:
```suggestion
        if matches!(auth_kind, DpsAuthKind::TpmDpsNonce{ .. }) {
```

massand, line 161:
```suggestion
        if matches!(auth_kind, DpsAuthKind::TpmDpsNonce { .. }) {
```

identity/aziot-dps-client-async/src/lib.rs
```diff
@@ -191,7 +188,7 @@ impl Client {
     // TPM provisioning has special 2-step challenge/response flow which is
     // basically identical to the symmetric key flow, except that the TPM client
     // is used instead of the key client. To keep things DRY, a recursive call
-    // is used, passing `DpsAuthKind::TpmWithAuth` as the auth kind.
+    // is used, passing `DpsAuthKind::TpmPostChallenge` as the auth kind.
```
massand, line 191:
```suggestion
    // is used, passing `DpsAuthKind::TpmDpsNonce` as the auth kind.
```

identity/aziot-dps-client-async/src/lib.rs
```diff
@@ -228,29 +225,33 @@ impl Client {
...
+                    false,
+                )
+                .await?;
+                let authorization_header_value = hyper::header::HeaderValue::from_str(&token)
+                    .map_err(|err| std::io::Error::new(std::io::ErrorKind::Other, err))?;
+                req.headers_mut()
+                    .append(hyper::header::AUTHORIZATION, authorization_header_value);
+                connector
+            }
+            DpsAuthKind::TpmPostChallenge => {
```
massand, line 245:
```suggestion
            DpsAuthKind::TpmDpsNonce => {
```

identity/aziot-dps-client-async/src/lib.rs
```diff
@@ -331,11 +332,15 @@ impl Client {
...
-                };
+                self.tpm_client
+                    .import_auth_key(
+                        base64::decode(reg_result.authentication_key)
+                            .map_err(|err| std::io::Error::new(std::io::ErrorKind::Other, err))?,
+                    )
+                    .await?;
 
+                // update auth method, and recurse
+                *auth_kind = DpsAuthKind::TpmPostChallenge;
```
massand, line 343:
```suggestion
                *auth_kind = DpsAuthKind::TpmDpsNonce;
```

massand (changes requested):
Lets follow the general naming convention for challenge response (which is also part of the DPS TPM docs)

identity/aziot-dps-client-async/src/lib.rs
```diff
@@ -306,6 +307,9 @@ impl Client {
         let body = hyper::body::to_bytes(body)
             .await
             .map_err(|err| std::io::Error::new(std::io::ErrorKind::Other, err))?;
+
+        // Unlike IoT Hub, DPS responses don't contain any potentially sensitive
```
massand, line 311: I'll eat my own words now looking at the response for TPM registration. DPS responses will contain the nonce, which is used as the signing key for registrations for a limited time period. So, let us remove the DPS response body too :)

arsing (approved):

identity/aziot-dps-client-async/src/lib.rs
```diff
@@ -161,7 +158,7 @@ impl Client {
             .await;
         };
 
-        if matches!(auth_kind, DpsAuthKind::TpmWithAuth { .. }) {
+        if matches!(auth_kind, DpsAuthKind::TpmDpsNonce { .. }) {
```
massand, line 161: It seems we are recursing within `request()` itself. The nonce is imported into TPM when the `auth_kind` within `request()` method changes from `Tpm` to `TpmDpsNonce`. So, why are we reimporting the nonce into the TPM again after exiting `request()` back into `register()`?

massand (changes requested):
Noticed something off about the two imports happening during TPM registration. And a DPS response logging problem I just saw

identity/aziot-dps-client-async/src/lib.rs
arsing: The first response from DPS contains a challenge key. The second response from DPS contains the symmetric key registered with IoT Hub.

massand: Ah yes. Thanks.

identity/aziot-dps-client-async/src/lib.rs
```diff
@@ -331,11 +335,15 @@ impl Client {
                 let reg_result: model::TpmRegistrationResult = serde_json::from_slice(&body)
                     .map_err(|err| std::io::Error::new(std::io::ErrorKind::Other, err))?;
 
-                // update auth method
-                *auth_kind = DpsAuthKind::TpmWithAuth {
-                    auth_key: reg_result.authentication_key,
-                };
+                self.tpm_client
```
massand, line 338: Maybe a comment here would remind readers (slow readers like me) about the difference between this `import_auth_key` (which is technically importing the DPS nonce and not the primary key for IoT Hub) and the other one outside.

massand: And another suggestion for future maintainers/reviewers: to add extra emphasis on the use of an auth key, I would consider renaming `aziot_identity_common::Credentials::Tpm` to `aziot_identity_common::Credentials::TpmAuthKey`

daprilik: We already had a comment in this file (at the top of the `request` method itself), so I just moved it to a more appropriate place + tweaked the wording a bit. I just linked back to the relevant docs about how the DPS TPM flow works.

Let me know if you want to change and wording.

massand (approved):

arsing (approved):

PR #171, opened 2021-03-22
[aziotctl] Add checks to load preloaded certs and private keys from certd and keyd

arsing:
... and test that certs and private keys with the same ID
correspond to each other.

Also re-export some common check utility functions from aziotctl-common
so that iotedge-check can reuse them.


aziotctl/src/internal/check/checks/read_certs.rs
```diff
@@ -0,0 +1,60 @@
...
+
+        for id in preloaded_certs.keys() {
+            let cert = cert_client
+                .get_cert(id)
+                .await
+                .with_context(|| format!("could not load cert with ID {:?}", id))?;
+
+            // PEM blob might have multiple certs, but we only care about the first one,
+            // so we use `openssl::x509::X509::from_pem` instead of `stack_from_pem`
+            let cert = openssl::x509::X509::from_pem(&cert)
```
gordonwang0, line 52: Does from_pem still check that all certs in the file parse correctly? If not, I think we should still use stack_from_pem to check that all certs parse and then take only the first one.

aziotctl/src/internal/check/checks/read_key_pairs.rs
```diff
@@ -0,0 +1,82 @@
...
+            //
+            // We also can't go behind keyd's back and load the keys ourselves, because the point of this check
+            // is to test keyd's ability to load the keys, but also because only keyd is allowed to load PKCS#11 keys
+            // since PKCS#11 implementations are generally not cross-process-safe.
+            //
+            // So as a best effort, ignore errors from loading key pairs entirely.
+            let key_handle = if let Ok(key_handle) = key_client.load_key_pair(id).await {
+                key_handle
+            } else {
+                continue;
```
gordonwang0, line 68: Maybe print a info message here? Something like "{id} couldn't be loaded, so will not be checked"

daprilik (changes requested):
Since we are adding new checks that require the various services to be up and running, I think we should modify the checks under `daemons_running.rs` to set a couple booleans in `CheckerCache`, and then use an `unwrap_or_skip!` guard before communicating with them.

aziotctl/src/internal/check/checks/certs_match_private_keys.rs
```diff
@@ -0,0 +1,47 @@
...
+impl CertsMatchPrivateKeys {
+    async fn inner_execute(
+        &mut self,
+        _shared: &CheckerShared,
+        cache: &mut CheckerCache,
+    ) -> Result<CheckResult> {
+        for (id, private_key) in &cache.private_keys {
+            if let Some(cert) = cache.certs.get(id) {
+                unsafe {
+                    openssl2::openssl_returns_1(openssl_sys2::X509_check_private_key(
```
massand, line 37: I am curious why this method is missing from the openssl_sys crate?

aziotctl/src/internal/check/checks/read_key_pairs.rs
massand: Putting that info behind the `shared.cfg.verbose` would help reduce the spew from several keys.

aziotctl/src/internal/check/checks/read_certs.rs
```diff
@@ -0,0 +1,60 @@
...
+                .with_context(|| format!("could not load cert with ID {:?}", id))?;
+
+            cache.certs.insert(id.clone(), cert);
```
massand, line 55: Failure to read a cert in `preloaded_certs` will ignore subsequent certs. This will leave the cache incomplete for the subsequent `read_key_pairs` check. Consider logging for failed certs and continuing with the remaining `preloaded_certs`

massand (changes requested):

aziotctl/src/internal/check/checks/read_certs.rs
arsing: It doesn't. Done.

aziotctl/src/internal/check/checks/read_key_pairs.rs
arsing: There is no logging possible from `check`, and even verbose messages wouldn't get seen if the check eventually succeeds.

aziotctl/src/internal/check/checks/read_certs.rs
arsing: I assume you meant `cert_matches_key_pairs`, not `read_key_pairs`.

To be clear, it's not a problem that `cert_matches_key_pairs` doesn't consider all the certs if there's an invalid cert. Even if the errors were aggregated till the end, `cert_matches_key_pairs` still would not cover all the certs, namely the ones that failed to parse. So there'll always be the possibility that once the user fixes `read_certs`, `cert_matches_key_pairs` will start failing where it was succeeding before.

But anyway, collecting all the errors helps with diagnosing all the issues in one shot instead of discovering them one at a time, so I've done that.

arsing:
>Since we are adding new checks that require the various services to be up and running, I think we should modify the checks under `daemons_running.rs` to set a couple booleans in `CheckerCache`, and then use an `unwrap_or_skip!` guard before communicating with them.

Done.

daprilik (changes requested):
Please also include `cache.identityd_running` and `cache.tpmd_running` as part of this PR (even if they are currently unused).

Also, at that point it'd be cleaner to define a separate `struct DaemonsActive` that mirrors the existing `struct DaemonConfigs`, and have the path to the options be `cache.daemon_running.{certd, keyd, tpmd, identityd}`

arsing:
>Please also include `cache.identityd_running` and `cache.tpmd_running` as part of this PR (even if they are currently unused).
>
>Also, at that point it'd be cleaner to define a separate `struct DaemonsActive` that mirrors the existing `struct DaemonConfigs`, and have the path to the options be `cache.daemon_running.{certd, keyd, tpmd, identityd}`

Done.

daprilik (approved):
:shipit:

massand (approved):
:shipit:

gordonwang0 (approved):

PR #172, opened 2021-03-22
Reprovision when identity reconciliation fails

gordonwang0:
If Identity Service fails to contact Hub during identity reconciliation, it will try once to reprovision the device.

### DIFF
--- a/identity/aziot-identityd/src/identity.rs
+++ b/identity/aziot-identityd/src/identity.rs
@@ -516,7 +516,11 @@ impl IdentityManager {
                             self.get_backup_provisioning_info(credential.clone())?;
 
                         if skip_if_backup_is_valid && backup_device.is_some() {
-                            backup_device.expect("backup device cannot be none")
+                            let backup_device =
+                                backup_device.expect("backup device cannot be none");
+                            log::info!("Provisioned with backup for {}.", backup_device.device_id);
+
+                            backup_device
                         } else {
                             let operation = dps_client
                                 .register(&registration_id, &dps_auth_kind)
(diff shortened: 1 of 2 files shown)


identity/aziot-identityd/src/lib.rs
```diff
@@ -412,10 +412,37 @@ impl Api {
...
+                    ReprovisionTrigger::Startup | ReprovisionTrigger::ConfigurationFileUpdate => {
+                        log::info!("Could not reconcile Identities with current device data. Reprovisioning.");
+
+                        self.id_manager
+                            .provision_device(self.settings.provisioning.clone(), false)
+                            .await?;
+                        log::info!("Successfully reprovisioned.");
+
+                        self.id_manager
+                            .reconcile_hub_identities(self.settings.clone())
```
massand, line 432: Tracing is beneficial for the second reconcile attempt as well

gordonwang0: If this fails, the error gets logged by the caller (and if it succeeds, the "Identity reconciliation complete message" below gets logged).

identity/aziot-identityd/src/identity.rs
```diff
@@ -516,7 +516,11 @@ impl IdentityManager {
                             self.get_backup_provisioning_info(credential.clone())?;
 
                         if skip_if_backup_is_valid && backup_device.is_some() {
-                            backup_device.expect("backup device cannot be none")
+                            let backup_device =
```
massand, line 519: Can we DRY these statements into the `operation_to_iot_hub_device` method and rename the method?

gordonwang0: This log message is only printed when a valid backup exists; i.e. when IS is not making an operation to IoT Hub.

massand: I forgot to tag my comment for all the statements from lines 515-527. I meant DRY all those lines for all DPS auth types. I think that can be done?

gordonwang0: Got it. Did that.

massand (approved):

PR #175, opened 2021-03-24
Add local_gateway_hostname (i.e: parent_hostname) configuration 

daprilik:
This is the first half of a 2 PR series that pushes `parent_hostname` configuration down into `aziot`. I'm still hammering out the details up the upstream changes (turns out there's some tricky re-plumbing that needs to get done, since _a lot_ of code was reading directly from `settings.parent_hostname()`, which likely won't exist...)

In any case, these changes _should_ be pretty straighforward, so if these look reasonable, we ought to merge them in and unblock folks waiting on [this half of] the feature.


arsing (changes requested):
Need to update [`template.toml`](https://github.com/Azure/iot-identity-service/blob/main/aziotctl/config/unix/template.toml) too. Copy the content from https://github.com/Azure/iotedge/blob/560d223fdb601aba994b51aa1b16c1ea08d72bcc/edgelet/contrib/config/linux/template.toml#L10-L17

arsing:
Don't you also need to change https://github.com/Azure/iot-identity-service/blob/main/identity/aziot-hub-client-async/src/lib.rs#L162 to use `local_gateway_hostname` over `iothub_hostname` ?

daprilik:
> Don't you also need to change https://github.com/Azure/iot-identity-service/blob/main/identity/aziot-hub-client-async/src/lib.rs#L162 to use local_gateway_hostname over iothub_hostname?

@massand?

aziotctl/aziotctl-common/src/config/super_config.rs
```diff
@@ -26,6 +26,14 @@ use url::Url;
 pub struct Config {
     pub hostname: Option<String>,
 
+    // DEVNOTE: ideally, this would be called `local_gateway_hostname`, to match
```
arsing, line 29: I think you can `#[serde(alias)]` it so that it gets deserialized with either name, then use different names in the `template.toml`s.

daprilik: ooooh, fancy. I guess the better question is whether or not that'd be more confusing than just having a single [somewhat mismatched] name for the field?

arsing: Should be fine. a-i-s users won't have the edge-specific file in the first place, and a-e users hopefully don't look at the a-i-s template and only at the a-e template.

daprilik: so you're thinking it makes sense to use `#[serde(alias)]` then?

arsing: Yes. It also solves the `(?)` in

>and rephrase the "Nested Edge" part to taste (?)

because you can just describe it as the local gateway.

arsing (approved):

massand:
> > Don't you also need to change https://github.com/Azure/iot-identity-service/blob/main/identity/aziot-hub-client-async/src/lib.rs#L162 to use local_gateway_hostname over iothub_hostname?
> 
> @massand?

Yes. If `local_gateway_hostname` is non-empty, that is the only path to create identities and `iothub_hostname` is only used in IS API responses in `hubName` field (along with `local_gateway_hostname` as the new `gatewayHost` field)

identity/aziot-identityd/src/identity.rs
```diff
@@ -532,6 +541,7 @@ impl IdentityManager {
                         dps_auth_kind,
                         registration_id,
                         credentials,
+                        provisioning.local_gateway_hostname,
```
massand, line 544: @huguesBouvier - Does the proxy route HTTPS traffic to DPS endpoints correctly? I assume it will, however the X509 case will not work. Can you confirm?

https://github.com/Azure/iot-identity-service/blob/main/identity/aziot-identityd/src/identity.rs#L497-L507
If X509 doesn't work in nested mode, we need to failfast in `DpsAttestationMode::X509` case, possibly before grabbing the identity cert, by throwing an error.

identity/aziot-hub-client-async/src/lib.rs
```diff
@@ -159,7 +159,7 @@ impl Client {
         TRequest: serde::Serialize,
         TResponse: serde::de::DeserializeOwned,
     {
-        let uri = format!("https://{}{}", hub_device.iothub_hostname, uri);
+        let uri = format!("https://{}{}", hub_device.local_gateway_hostname, uri);
```
massand, line 162: https://github.com/Azure/iot-identity-service/blob/main/identity/aziot-hub-client-async/src/lib.rs#L311-L321
Needs to be changed in `request_no_content` too

identity/aziot-identityd/src/identity.rs
massand: I believe the same issue will hit the IoT Hub `ManualAuthMethod::X509` case as well

aziotctl/aziotctl-common/src/config/apply.rs
```diff
@@ -207,7 +208,10 @@ pub fn run(
             super_config::ProvisioningType::None => aziot_identityd_config::ProvisioningType::None,
         };
 
-        aziot_identityd_config::Provisioning { provisioning }
+        aziot_identityd_config::Provisioning {
+            provisioning,
+            local_gateway_hostname,
```
massand, line 213: See my other comment about X509 auth types and nested mode. It is likely we need to validate the super-config for avoiding nested mode configuration with a device identity that uses X509 auth type (DPS or IoT Hub). I believe IoT Hub X509 auth type will be supported soon however (check with @huguesBouvier)

massand (changes requested):
Nested mode definitely doesn't support DPS X509. IoT Hub X509 support in nested mode is WIP I believe

massand (approved):
We can move the X509 changes to a separate PR

identity/aziot-identityd/src/identity.rs
huguesBouvier: Right now we don't support DPS

aziotctl/config/unix/template.toml
```diff
@@ -6,6 +6,14 @@
...
 # hostname = "my-device"
 
+# ==============================================================================
+# Local Gateway Hostname
+# ==============================================================================
+#
+# If this device is running in a nested deployment, uncomment the next line to
+# set the local gateway hostname of this device.
+#
+# local_gateway_hostname = "my-parent-device"
```
varunpuranik, line 16: @daprilik - I thought this was going to be called parent_hostname. 
Also, should this not be in identityd config as well?

aziotctl/aziotctl-common/src/config/apply.rs
varunpuranik, line 213: I missed this - why is this part of provisioning, instead of at the top level?

PR #179, opened 2021-03-25
[aziotctl] Add a check that the latest version of the package is installed.

arsing:


aziotctl/src/internal/check/checks/aziot_version.rs
```diff
@@ -0,0 +1,123 @@
...
+        self.expected_version = Some(expected_version.clone());
+
+        let actual_version = env!("CARGO_PKG_VERSION");
+        self.actual_version = Some(actual_version.to_owned());
+
+        if expected_version != actual_version {
+            return Ok(CheckResult::Warning(
+                anyhow!(
+                    "Installed aziot-identity-service package has version {} but {} is the latest stable version available.\n\
+                    Please see https://aka.ms/iotedge-update-runtime for update instructions.",
```
arsing, line 110: @micahl Do we have any docs planned for this? I don't believe we do.

If not, I can proactively create an `https://aka.ms/aziot-update-runtime` that points to the same page as this one for now, but gives us the flexibility to make it point to something else in the future.

micahl: @arsing correct, we don't have something for this yet.  Please create a redirect link that we can update later.  Also see my separate email on the difference between fwlink and aka.ms.

arsing: Okay, I've created the new aka.ms URL.

re: fwlink - We use target links without the language code, eg this one redirects to `https://docs.microsoft.com/azure/iot-edge/how-to-update-iot-edge`, not `https://docs.microsoft.com/en-us/azure/iot-edge/how-to-update-iot-edge`. The former redirects to the latter or to the appropriate language based on user agent's `accept-language` header. So eg a user whose browser sends `accept-language: it-it` will get redirected to the `/it-it/` page. So I think it's fine in that regard.

At any rate, I don't know how to fill the required fields it wants me to fill for a fwlink URL, and if we really want to switch to fwlink we have a bunch of these links to update (both in this repo and in iotedge), so let's do that after GA.

massand (approved):

gordonwang0 (approved):

PR #180, opened 2021-03-25
Add check for preventing DPS provisioning in nested mode

massand:
Nested Edge deployments only support IoT Hub provisioning for release/1.2. This PR introduces informative errors on invoking `aziotctl config apply` and daemon logs, if Identity Service is configured with `parent_hostname` AND DPS provisioning parameters.

### DIFF
--- a/aziotctl/aziotctl-common/src/config/apply.rs
+++ b/aziotctl/aziotctl-common/src/config/apply.rs
@@ -529,7 +529,7 @@ mod tests {
     #[should_panic(expected = "DPS provisioning is not supported in nested mode")]
     fn dps_not_supported_in_nested() {
         let super_config = r#"
-parent_hostname = "device.hostname"
+local_gateway_hostname = "device.hostname"
 
 [provisioning]
 source = "dps"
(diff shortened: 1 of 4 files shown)


aziotctl/aziotctl-common/src/config/apply.rs
```diff
@@ -520,4 +524,29 @@ mod tests {
             );
         }
     }
+
+    #[test]
+    #[should_panic(expected = "DPS provisioning is not supported in nested mode")]
+    fn dps_not_supported_in_nested() {
+        let super_config = r#"
+parent_hostname = "device.hostname"
```
arsing, line 532:
```diff,toml
-parent_hostname = "device.hostname"
+local_gateway_hostname = "device.hostname"
```

arsing (changes requested):

aziotctl/aziotctl-common/src/config/apply.rs
massand: Fixed

arsing (approved):

PR #181, opened 2021-03-25
Remove usage of 'foo' in URLs

gordonwang0:
Replace 'foo' with the socket file name.


cert/aziot-cert-client-async/src/lib.rs
```diff
@@ -43,7 +43,7 @@ impl Client {
         let res: aziot_cert_common_http::get_cert::Response = http_common::request(
             &self.inner,
             http::Method::POST,
-            &format!("http://foo/certificates?api-version={}", self.api_version),
+            &format!("http://certd.sock/certificates?api-version={}", self.api_version),
```
arsing, line 46: Like I said in the scrum, for these clients, can you just remove the scheme and authority entirely (ie everything before the first `/` of the path) and see if that works.

gordonwang0: That doesn't work. I get the error: `client requires absolute-form URIs`

arsing: Ah indeed. `http::Uri` is fine with relative URIs but `hyper::Client` isn't.

arsing (approved):

PR #183, opened 2021-03-25
Revert use of `Arc::{in,de}crement_strong_count` since iotedge still compiles this code with Rust <1.51

arsing:


daprilik (approved):
Soon :tm:

PR #185, opened 2021-03-26
[aziotctl check] Add supported platform check

daprilik:
Opening as draft while I solicit feedback on details such as warning messages, what platforms constitute which "Tier", and confirming that the implementation actually works as intended across multiple distros.

Two important notes:

- Please ignore the current Tier lists in the code. These are just placeholders until we nail down the tiers.
- *I have not tested the `os_info` library across all distros yet - it may or may not be working as intended!*


aziotctl/src/internal/check/checks/supported_platform.rs
```diff
@@ -0,0 +1,89 @@
...
+            | Type::Redhat
+            | Type::RedHatEnterprise
+            | Type::SUSE => Tier::Tier2,
+            _ => Tier::Tier3,
+        };
+
+        Ok(match tier {
+            Tier::Tier1 => CheckResult::Ok,
+            Tier::Tier2 => CheckResult::Warning(anyhow!(
+                "Running on a partially supported, Tier 2 platform ({}). lorem ipsum",
```
lt72, line 80: Can we add the URL to the tiering docs?

lt72: https://docs.microsoft.com/en-us/azure/iot-edge/support?view=iotedge-2018-06

lt72: You should check with Kelly if that is the URL to use or if there a lasting one that can be redirected as needed

arsing: We need an aziot equivalent of https://aka.ms/iotedge-platsup . I'll make https://aka.ms/aziot-platsup for now and we can redirect it to aziot-specific documentation later (same approach we used for https://github.com/Azure/iot-identity-service/pull/179#discussion_r600971063 )

arsing: Done. @daprilik Use https://aka.ms/aziot-platsup

daprilik: Awesome. Didn't know we already had this list. I'll update the PR tomorrow.

PR #188, opened 2021-03-27
Fix `iotedge config apply` not picking up parent hostname because of serde bug.

arsing:


aziotctl/aziotctl-common/src/config/super_config.rs
```diff
@@ -26,11 +26,17 @@ use url::Url;
...
+    // DEVNOTE:
+    //
+    // This field is called local_gateway_hostname in the identity-service super-config and parent_hostname in the iotedge super-config.
+    // Since this super-config type is shared between the two, we use an alias to allow both names to work.
+    //
+    // Ideally this would be `#[serde(alias = "parent_hostname")] local_gateway_hostname: Option<String>`, but due to a serde bug [1]
+    // this does not work. So we have to do it the other way around.
+    //
+    // [1]: https://github.com/serde-rs/serde/issues/1976
+    #[serde(alias = "local_gateway_hostname")]
```
varunpuranik, line 38: Why do we need to do this? Can we just call it parent_hostname in both places?

varunpuranik: Note that parent/child is how it is called in all documentation (even for edge-leaf scenarios)

varunpuranik (approved):

PR #191, opened 2021-03-30
Update version check for 1.2.0

gordonwang0:
Updates the version check for the latest https://raw.githubusercontent.com/Azure/azure-iotedge/master/latest-versions-non-lts.json


arsing:
It was supposed to be `aziot`. Why change the old code to match the new file instead of writing the new file to match the old code?

gordonwang0:
Consistency with the package name. But sure, it's fine to keep it aziot.

arsing:
Consistency with the package name is fine. In that case make sure to change the first one to `aziot-edge` too.

gordonwang0:
Ok, I'll do that.

arsing (approved):

PR #192, opened 2021-03-31
Identity Service offline support for get module

massand:
The Identity Service always required an active connection to IoT Hub to handle all incoming API requests. This caused regressions
in the previous offline behavior of IoT Edge (which relied on libiothsm for all module identity signing requests), while the device
was running and IoT Edge needed to fulfill signing requests (which simply rely on module_id and generation_id as input parameters).

This change adds module backups, which comprise of module identity information (excluding credentials), from IoT Hub. The backup files are always overwritten with fresh information from IoT Hub. When the device is offline, the module backups are read into memory and used in API responses. Key handles are still freshly generated for each API request.

Device reprovisioning will purge old module information, if the device information has changed.


massand:
@gordonwang0 / @arsing - I have run this through our single-node Edge E2E tests against the latest iot-identity-service package built for this PR and there were no regressions reported. I manually tested the Edge offline mode works as 1.1.

identity/aziot-identityd/src/lib.rs
```diff
@@ -386,7 +386,7 @@ impl Api {
                 // Clear the backed up device state before reprovisioning.
                 // If this fails, log a warning but continue with reprovisioning.
                 let mut backup_file = self.settings.homedir.clone();
-                backup_file.push("device_info");
+                backup_file.push(identity::DEVICE_BACKUP_LOCATION);
 
                 if let Err(err) = std::fs::remove_file(backup_file) {
```
arsing, line 391: Need to delete the new files too?

massand: If /reprovision API is called, it means the user wants to wipe the device backup completely. The module backup for that device is purged only on setting the device with new provisioning information in IdentityManager. Even that is purged only when the device has changed.

identity/aziot-identityd/src/identity.rs
```diff
@@ -830,3 +935,149 @@ impl HubDeviceInfo {
...
+            );
+        }
+    }
+
+    pub fn set_module_backup(
+        homedir_path: &Path,
+        iothub_hostname: &str,
+        device_id: &str,
+        module_id: &str,
+        data: &Option<aziot_identity_common::hub::Module>,
```
arsing, line 981: `&Option<T>` -> `Option<&T>`

identity/aziot-identityd/src/lib.rs
```diff
@@ -537,9 +537,17 @@ impl Api {
...
-        let _ = self
+        // Attempt to re-provision the device. Failures need to be logged and the device should
+        // run offline.
+        if let Err(err) = self
             .reprovision_device(auth::AuthId::LocalRoot, trigger)
-            .await?;
+            .await
+        {
+            log::warn!(
+                "Failed to reprovisioning device. Running offline. Reprovisioning failure reason: {}. ",
```
arsing, line 547: `Failed to reprovisioning` -> `Failed to reprovision`

identity/aziot-identityd/src/lib.rs
arsing: I'm asking why it makes sense to leave the module backup around when the device backup has been deleted. If the device hasn't been reprovisioned in the cloud the same backup will be written back at the end, so in the same way the same backup for the modules can also be written back.

identity/aziot-identityd/src/identity.rs
```diff
@@ -266,6 +344,20 @@ impl IdentityManager {
                 let identities = response
```
arsing, line 344: You don't need to fall back to calling `get_module_backup` if the `client.get_modules()` on line 342 fails?

identity/aziot-identityd/src/lib.rs
massand: That is intentional to leave a backup if any provisioning / reconciliation fails. Two reasons- 1) IdentityManager always prioritizes and overwrites fresh identity data from the cloud for all operations and b) IdentityManager will cleanup the module identity backups on reconciliation. Leaving the module backup alone doesn't cause any issues (even though in some very corner cases, modules maybe left behind) because it is only accessible after a) the device is set in Identity Manager (after reprovisioning) and b) the "iothub + device" match. The purging should only happen when step (b) has a new device. Setting it prior to that isn't necessary. If we always deleted module backup before both (a) and (b) happen AND reconcile ever fails, the backup won't be available for /sign operations in some valid scenarios.

identity/aziot-identityd/src/identity.rs
massand: Only single identity read operations are supported offline. All other APIs fail when the device isn't connected.

identity/aziot-identityd/src/identity.rs
massand: Turns out I didn't need to send a reference. I fixed some instances of this in `lib.rs` for `issue_local_identity` though.

identity/aziot-identityd/src/lib.rs
massand: Fixed

gordonwang0 (approved):

arsing (approved):

massand:
Kodiak seems down. I will manually squash and merge this in, if it persists after latest CI run.

PR #193, opened 2021-03-31
Fix connect upstream test in nested edge.

huguesBouvier:
Connect upstream test will now use parent hostname instead of iothubhostname when nested.

Decorate error message in nested to help user troubleshoot.

Test:
Tested happy path on lvl5(non nested)
Tested happy path on lvl4 and made sure it was connecting to parent.
Removed trust bundle from store and correctly got the following error message:
host can connect to and perform TLS handshake with iothub AMQP port - Error
    Make sure the parent device is reachable using 'curl https://parenthostname'. Make sure the the trust bundle has been added to the trusted store: sudo cp <path>/azure-iot-test-only.root.ca.cert.pem /usr/local/share/ca-certificates/azure-iot-test-only.root.ca.cert.pem.crt
                sudo update-ca-certificates

### DIFF
--- a/aziotctl/src/internal/check/checks/host_connect_iothub.rs
+++ b/aziotctl/src/internal/check/checks/host_connect_iothub.rs
@@ -68,17 +68,38 @@ impl HostConnectIotHub {
     ) -> Result<CheckResult> {
         use aziot_identityd_config::ProvisioningType;
 
-        let iothub_hostname = match &shared.cfg.iothub_hostname {
-            Some(s) => s,
+        let identityd_config = unwrap_or_skip!(&cache.cfg.identityd);
+
+        let (iothub_hostname, nested) = match &shared.cfg.iothub_hostname {
+            Some(s) => (s, false),
             None => {
-                let iothub_hostname = match &unwrap_or_skip!(&cache.cfg.identityd)
-                    .provisioning
-                    .provisioning
-                {
+                let (iothub_hostname, nested) = match &identityd_config.provisioning.provisioning {
                     ProvisioningType::Manual {
                         iothub_hostname, ..
-                    } => iothub_hostname,
+                    } => {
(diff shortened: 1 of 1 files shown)


aziotctl/src/internal/check/checks/host_connect_iothub.rs
```diff
@@ -68,17 +68,35 @@ impl HostConnectIotHub {
     ) -> Result<CheckResult> {
         use aziot_identityd_config::ProvisioningType;
 
-        let iothub_hostname = match &shared.cfg.iothub_hostname {
-            Some(s) => s,
+        let identityd_config = unwrap_or_skip!(&cache.cfg.identityd);
+
+        let (iothub_hostname, nested) = match &shared.cfg.iothub_hostname {
+            Some(s) => (s, false),
```
daprilik, line 74: this raises the question of whether or not we should add a `--gateway-hostname` flag to compliment the existing `--iothub-hostname` flag that already exists.

Obviously not part of this PR, but this is nonetheless something we should think about.

aziotctl/src/internal/check/checks/host_connect_iothub.rs
```diff
@@ -68,17 +68,35 @@ impl HostConnectIotHub {
...
-                    } => iothub_hostname,
+                    } => {
+                        // use `local_gateway_hostname` if available
+                        (
+                            identityd_config
+                                .provisioning
+                                .local_gateway_hostname
+                                .as_ref()
+                                .unwrap_or(iothub_hostname),
+                            true,
```
daprilik, line 87: shouldn't this be `identityd_config.provisioning.local_gateway_hostname.is_some()`?

huguesBouvier: Good catch it was a bug

daprilik (approved):

PR #194, opened 2021-04-01
Fix connect upstream test in nested edge.

huguesBouvier:
Fix amqp tests:

https://dev.azure.com/msazure/One/_build/results?buildId=40970480&view=logs&j=e8afecc0-c259-5cc1-62e0-5c5d1c2a3db4&t=a5111985-4f75-51bb-4ed8-1624e65f53ff


huguesBouvier (commented):
Is that how you update the branch? I cannot do a direct push from my computer.

PR #202, opened 2021-04-14
Add options to config EST URL and auth per cert

gordonwang0:
Allows the auth and url options to be set for a specific EST-issued certs. The existing option `[cert_issuance.est]` is now a default to fall back on if options are not set per cert.

### DIFF
--- a/aziotctl/aziotctl-common/src/config/apply.rs
+++ b/aziotctl/aziotctl-common/src/config/apply.rs
@@ -1,5 +1,7 @@
 // Copyright (c) Microsoft. All rights reserved.
 
+use std::str::FromStr;
+
 use anyhow::anyhow;
 
 use super::super_config;
@@ -56,6 +58,12 @@ pub fn run(
         certs: vec![],
     };
 
+    // Authorization of CS with KS.
+    let mut aziotcs_keys = aziot_keyd_config::Principal {
+        uid: aziotcs_uid.as_raw(),
+        keys: vec![],
+    };
+
     let provisioning = {
         let super_config::Provisioning { provisioning } = provisioning;
 
@@ -103,7 +111,20 @@ pub fn run(
 
                     super_config::ManualAuthMethod::X509 { identity } => {
                         match identity {
-                            super_config::X509Identity::Issued { identity_cert } => {
+                            super_config::X509Identity::Issued { mut identity_cert } => {
+                                if let aziot_certd_config::CertIssuanceMethod::Est {
+                                    url: _,
+                                    ref mut auth,
+                                } = &mut identity_cert.method
+                                {
+                                    set_device_id_est_auth(
+                                        auth,
+                                        &mut preloaded_certs,
+                                        &mut preloaded_keys,
+                                        &mut aziotcs_keys,
+                                    )?;
+                                }
+
                                 aziotid_keys.keys.push(super::DEVICE_ID_ID.to_owned());
 
                                 cert_issuance_certs
@@ -168,7 +189,20 @@ pub fn run(
                         identity,
                     } => {
                         match identity {
-                            super_config::X509Identity::Issued { identity_cert } => {
+                            super_config::X509Identity::Issued { mut identity_cert } => {
+                                if let aziot_certd_config::CertIssuanceMethod::Est {
+                                    url: _,
+                                    ref mut auth,
+                                } = &mut identity_cert.method
+                                {
+                                    set_device_id_est_auth(
+                                        auth,
+                                        &mut preloaded_certs,
+                                        &mut preloaded_keys,
+                                        &mut aziotcs_keys,
+                                    )?;
+                                }
+
                                 aziotid_keys.keys.push(super::DEVICE_ID_ID.to_owned());
 
                                 cert_issuance_certs
@@ -262,12 +296,6 @@ pub fn run(
         "homedir_path".to_owned(),
         super::AZIOT_KEYD_HOMEDIR_PATH.to_owned(),
     );
-
-    // Authorization of CS with KS.
-    let mut aziotcs_keys = aziot_keyd_config::Principal {
-        uid: aziotcs_uid.as_raw(),
-        keys: vec![],
-    };
 
     let certd_config = {
         let super_config::CertIssuance { est, local_ca } = cert_issuance;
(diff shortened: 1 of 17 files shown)


arsing:
Rather than make a magical ID like this, the way I was imagining this would be solved is to extend the `CertIssuanceOptions` struct to have the EST URL if `method == "EST"`. Not the `CertIssuanceOptions` in the certd-config crate, but a new derived one for the super-config.

gordonwang0:
Yeah, that makes sense. So the config would be:

```
[provisioning.attestation]
identity_cert = { method = "est", common_name = "gowa-test-1", url = "https://example.com/.well-known/est" }
```

As an aside, we should make the auth options specific to each URL too, as there might be different username/passwords/certs for each server.

arsing:
>Yeah, that makes sense. So the config would be:

Yes. Something like:

```rust
struct CertIssuanceOptions {
    #[serde(flatten)]
    method: CertIssuanceMethod,
    ...
}

#[serde(tag = "method")]
enum CertIssuanceMethod {
    #[serde(rename = "est")]
    Est {
        url: Option<Url>,
    },
    ...
}
```

>As an aside, we should make the auth options specific to each URL too, as there might be different username/passwords/certs for each server. But that requires modifying certd too, so let's do that later.

Might be hard to do that backward-incompatibly.

gordonwang0:
We could add to CertIssuanceMethod:
```
[provisioning.attestation]
identity_cert = { method = "est", common_name = "gowa-test-1", url = "https://example.com/.well-known/est", username = "user", password = "pass", identity_cert = "file:///...", identity_pk = "file:///..." }
```

`[cert_issuance.est.auth]` would then just be the default if it's not set per URL. That would be backwards-compatible. Though I don't like how long the CertIssuance object is going to be. There's no way to line break the TOML, is there?

Or we could do:
```
[provisioning.attestation]
identity_cert = { method = "est", common_name = "gowa-test-1" }

[cert_issuance.est.urls]
identity_cert = "https://example.com/.well-known/est"

[cert_issuance.est.urls.identity_cert]
username = "user"
password = "pass"
identity_cert = "file:///..."
identity_pk = "file:///..."
```

Where again `[cert_issuance.est.auth]` would then just be the default if it's not set per URL. I think this looks cleaner, though if we do this way then we should keep the magic `identity_cert` value and not add the URL to CertIssuance.

Thoughts?

arsing:
>`[cert_issuance.est.auth]` would then just be the default if it's not set per URL.

True, keeping it as a fallback will work.

>Though I don't like how long the CertIssuance object is going to be. There's no way to line break the TOML, is there?

You can break the inline table `{ ... }` into more than one line so that each property is on its own line.

>Or we could do: [...] I think this looks cleaner, though if we do this way then we should keep the magic `identity_cert` value and not add the URL to CertIssuance.

Right, the magic value is what I want to avoid.

gordonwang0:
> You can break the inline table { ... } into more than one line so that each property is on its own line.

How does that work? I'm not able to get it to validate with https://www.toml-lint.com/.

> Right, the magic value is what I want to avoid.

Is the magic value really so bad? We're already using one right now, it's just called `device-id` instead of `identity_cert`.

arsing:
Oh, never mind. Line breaks are only allowed inside arrays, not inline tables.

I think it's fine. It's probably unlikely that the user will have multiple URLs with different auth. If they do, they can live with the line being long, or they can manually convert it to a non-inline table; it's all the same to the parser in the end.

>Is the magic value really so bad? We're already using one right now, it's just called `device-id` instead of `identity_cert`.

We don't have any magic values right now.

gordonwang0:
I've implemented the changes discussed. Still need to update the super config (and iotedge), but please take an initial look.

aziotctl/aziotctl-common/src/config/apply.rs
```diff
@@ -103,7 +111,20 @@ pub fn run(
 
                     super_config::ManualAuthMethod::X509 { identity } => {
                         match identity {
-                            super_config::X509Identity::Issued { identity_cert } => {
+                            super_config::X509Identity::Issued { mut identity_cert } => {
+                                if let aziot_certd_config::CertIssuanceMethod::Est {
+                                    url: _,
+                                    ref mut auth,
```
arsing, line 117: You don't need the `ref mut`. You're matching a `&mut` so `auth` is already a `&mut` itself.

cert/aziot-certd-config/src/lib.rs
```diff
@@ -277,6 +233,39 @@ pub enum CertIssuanceMethod {
...
 }
 
+fn deserialize_est_auth<'de, D>(deserializer: D) -> Result<Option<EstAuth>, D::Error>
+where
+    D: serde::de::Deserializer<'de>,
+{
+    let inner: Option<EstAuthInner> = serde::Deserialize::deserialize(deserializer)?;
+
+    if let Some(inner) = inner {
+        let auth = match deserialize_auth_inner(&inner) {
```
arsing, line 243: `let auth = deserialize_auth_inner(&inner).map_err(serde::de::Error::missing_field)?;`

cert/aziot-certd/src/lib.rs
```diff
@@ -394,7 +397,8 @@ fn create_cert<'a>(
                         ))
                     })?;
 
-                    let url = urls
+                    let auth = cert_auth.as_ref().unwrap_or(&default_auth);
+                    let url = cert_url.as_ref().unwrap_or(urls
```
arsing, line 401: `.unwrap_or_else(|| urls.get(id)).or_else(|| ...` so that it's evaluated lazily.

arsing (changes requested):

aziotctl/aziotctl-common/src/config/apply.rs
gordonwang0: Fixed

cert/aziot-certd-config/src/lib.rs
gordonwang0: Fixed

gordonwang0:
@arsing How do you suggest solving the issue with the difference between `aziot_certd_config::EstAuth` and `super_config::EstAuth`? I'm trying to add the new EST options to the super config, but it's proving problematic.

In `super_config`, either EST `bootstrap_id` or `id` certs are provided. In `aziot_certd_config`, `id` must be provided and `bootstrap_id` is optional. The `super_config` uses a placeholder `est-id` if `bootstrap_id` is provided, but now with the option to set the auth per-cert, we can't know which placeholder name to use (maybe just the cert id, but that's not guarenteed to work).

arsing (approved):

PR #203, opened 2021-04-15
Export connection string parsing + validation function

daprilik:
Pre-requisite for enabling `iotedge config mp` to perform early validation of the passed-in connection string.

### DIFF
--- a/aziotctl/aziotctl-common/src/config/mod.rs
+++ b/aziotctl/aziotctl-common/src/config/mod.rs
@@ -63,7 +63,7 @@ pub fn write_file(
     Ok(())
 }
 
-fn parse_manual_connection_string(
+pub fn parse_manual_connection_string(
     connection_string: &str,
 ) -> Result<(String, String, Vec<u8>), String> {
     const HOSTNAME_KEY: &str = "HostName";


arsing:
Instead of making it `pub` and then ad-hoc calling it, change the super-config type to use the parsed type instead of a `String` so that (de)serializing it is already an error.

daprilik:
I'm not entirely sure what you mean. `iotedge config mp` bypasses any deserialization, since it crafts a super_config struct in-memory and then serializes it into a file.

Unless you mean going through the effort of making a whole new newtype (e.g: `pub struct ConnectionString(String)`) that enforces well-formed contents through a custom `Deserialize` implementation?

arsing:
>since it crafts a super_config struct in-memory and then serializes it into a file.

Yes, and "serializes" is part of "(de)serializing"


>Unless you mean going through the effort of making a whole new newtype

No. Just `ManualProvisioning::ConnectionString` needs to be changed from `connection_string: String` to the same as its `Explicit` neighbor.

daprilik:
I'm not entirely sure what you're suggesting, so I shot you a DM on teams.

daprilik:
Updated following our discussion. Will need to update downstream `iotedge` PR as well...

aziotctl/aziotctl-common/src/config/super_config.rs
```diff
@@ -99,6 +99,48 @@ pub enum ManualProvisioning {
...
+        struct Visitor;
+
+        impl<'de> serde::de::Visitor<'de> for Visitor {
+            type Value = ConnectionString;
+
+            fn expecting(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+                f.write_str("connection string")
+            }
+
+            fn visit_str<E>(self, s: &str) -> Result<Self::Value, E>
```
arsing, line 132: Also impl `visit_string`

aziotctl/src/config/mp.rs
```diff
@@ -59,7 +59,10 @@ To reconfigure IoT Identity Service, run:
         provisioning: common_config::super_config::Provisioning {
             provisioning: common_config::super_config::ProvisioningType::Manual {
                 inner: common_config::super_config::ManualProvisioning::ConnectionString {
-                    connection_string,
+                    connection_string: common_config::super_config::ConnectionString::new(
+                        connection_string,
+                    )
+                    .map_err(|e| anyhow!("Invalid connection string: {}", e))?,
```
arsing, line 65: `"invalid connection string: {}"` - lower-case "i", for consistency.

arsing (changes requested):

arsing (approved):

PR #207, opened 2021-04-23
aziotctl check: warn when using unrecommended key types

gordonwang0:
Extends the read_key_pairs check to warn when using RSA keys smaller than 2048b or EC keys that don't use curve P-256.

### DIFF
--- a/aziotctl/src/internal/check/checks/read_key_pairs.rs
+++ b/aziotctl/src/internal/check/checks/read_key_pairs.rs
@@ -8,6 +8,8 @@ use crate::internal::check::{CheckResult, Checker, CheckerCache, CheckerMeta, Ch
 #[derive(Serialize, Default)]
 pub struct ReadKeyPairs {}
 
+const RSA_RECOMMENDED_MIN_BITS: u32 = 2048;
+
 #[async_trait::async_trait]
 impl Checker for ReadKeyPairs {
     fn meta(&self) -> CheckerMeta {
@@ -56,7 +58,8 @@ impl ReadKeyPairs {
             key_engine
         };
 
-        let mut err_aggregated = String::new();
+        let mut err_aggregated = vec![];
+        let mut warn_aggregated = vec![];
 
         for id in preloaded_keys.keys() {
             // We don't know whether `id` is a symmetric or asymmetric key,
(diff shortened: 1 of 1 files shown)


aziotctl/src/internal/check/checks/read_key_pairs.rs
```diff
@@ -82,23 +85,54 @@ impl ReadKeyPairs {
...
+                        openssl::pkey::Id::RSA => {
+                            if key_length < RSA_RECOMMENDED_MIN_BITS {
+                                warn_aggregated.push(format!(
+                                    "RSA key {} has length {} (min recommended: {})",
+                                    id, key_length, RSA_RECOMMENDED_MIN_BITS
+                                ));
+                            }
+                        }
+                        openssl::pkey::Id::EC => {
+                            if key_length != 256 {
```
arsing, line 103: You should check the curve rather than the number of bits, because it could be another curve that's 256 bits.

```rust
key.ec_key().group().curve_name() == Some(openssl::nid::Nid::X9_62_PRIME256V1)
```

aziotctl/src/internal/check/checks/read_key_pairs.rs
```diff
@@ -82,23 +85,54 @@ impl ReadKeyPairs {
                 })?;
                 Ok::<_, anyhow::Error>(key)
             };
+
             match key() {
                 Ok(key) => {
+                    let key_length = key.bits();
+
+                    match key.id() {
```
arsing, line 93: Since you need `ec_key` per my other comment, it's easier to do `if let Ok(rsa) = key.rsa() { ... } else if let Ok(ec_key) = key.ec_key() { ... } else { Err(...) }` which combines the type check with getting the type-specific parameters.

With `.rsa()` you get the modulus length using `rsa.size()`, but note that it's in bytes, not bits.

arsing (changes requested):

aziotctl/src/internal/check/checks/read_key_pairs.rs
gordonwang0: Fixed

aziotctl/src/internal/check/checks/read_key_pairs.rs
gordonwang0: Fixed

arsing (approved):

PR #208, opened 2021-04-26
Make set_est_auth public

gordonwang0:
This function need to be public so `iotedge config apply` can use it.


arsing:
`iotedge config apply` needs it for the Edge CA?

gordonwang0:
Yes

arsing (approved):

PR #210, opened 2021-04-27
Add cert_name parameter to set_est_auth

gordonwang0:
This parameter is needed so that different cert IDs for the EST auth certs can be set for device ID and Edge CA certs.

### DIFF
--- a/aziotctl/aziotctl-common/src/config/apply.rs
+++ b/aziotctl/aziotctl-common/src/config/apply.rs
@@ -119,6 +119,7 @@ pub fn run(
                                             &mut preloaded_certs,
                                             &mut preloaded_keys,
                                             &mut aziotcs_keys,
+                                            super::DEVICE_ID_ID,
                                         )
                                     } else {
                                         None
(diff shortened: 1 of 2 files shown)


aziotctl/aziotctl-common/src/config/apply.rs
```diff
@@ -500,33 +502,30 @@ pub fn set_est_auth(
         aziot_keys_common::PreloadedKeyLocation,
     >,
     aziotcs_keys: &mut aziot_keyd_config::Principal,
+    cert_name: &str,
 ) -> Option<aziot_certd_config::EstAuth> {
     auth.as_ref().map(|auth| {
         let auth_x509 = auth.x509.as_ref().map(|x509| {
+            let identity_cert_id = format!("{}-{}", super::EST_ID_ID, cert_name);
```
arsing, line 509: Can you replace the `EST_ID_DEVICE_ID` with this too, then? So that there aren't two different implementations of this algorithm.

gordonwang0: I don't understand. What are you referring to by `EST_ID_DEVICE_ID`? I removed that constant and replaced it with this.

arsing: Sorry, didn't notice the second file's change.

arsing (approved):

PR #211, opened 2021-04-27
Rate-limit + timeout fix for incomplete socket reads

massand:
There are two parts to the fix:
1. ensure request content read is timed out within 5 seconds to avoid content-length being set to greater value than actual content 
2. rate-limit all clients to 10 concurrent requests per uid per service (this applies uniformly to all uids)

### DIFF
--- a/http-common/Cargo.toml
+++ b/http-common/Cargo.toml
@@ -21,6 +21,7 @@ percent-encoding = "2"
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 tokio = { version = "1", features = ["net", "rt-multi-thread"], optional = true }
+tokio-io-timeout = { version = "1" }
 tracing = { version = "0.1", features = ["log"] }
 url = { version = "2", features = ["serde"] }
 
--- a/http-common/Cargo.toml
+++ b/http-common/Cargo.toml
@@ -20,7 +20,7 @@ openssl = { version = "0.10", optional = true }
 percent-encoding = "2"
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
-tokio = { version = "1", features = ["net", "rt-multi-thread"], optional = true }
+tokio = { version = "1", features = ["net", "rt-multi-thread", "sync"], optional = true }
 tokio-io-timeout = { version = "1" }
 tracing = { version = "0.1", features = ["log"] }
 url = { version = "2", features = ["serde"] }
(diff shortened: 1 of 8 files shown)


http-common/Cargo.toml
```diff
@@ -20,7 +20,8 @@ openssl = { version = "0.10", optional = true }
 percent-encoding = "2"
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
-tokio = { version = "1", features = ["net", "rt-multi-thread"], optional = true }
+tokio = { version = "1", features = ["net", "rt-multi-thread", "sync"], optional = true }
+tokio-io-timeout = { version = "1" }
```
arsing, line 24: `tokio-io-timeout` by itself is not enough. That sets the limit for every individual read, not for the whole read of the request. That means the client can drip-feed one byte every 5s and the request won't time out.

You also need to apply a [`tokio::time::Timeout`](https://docs.rs/tokio/1.5.0/tokio/time/fn.timeout.html) on all the calls to `hyper::body::to_bytes(body)` inside the `make_service` macro.

arsing (changes requested):

http-common/src/connector.rs
```diff
@@ -27,8 +27,15 @@ pub enum AsyncStream {
...
 #[derive(Debug)]
 pub enum Incoming {
-    Tcp(tokio::net::TcpListener),
-    Unix(tokio::net::UnixListener),
+    Tcp {
+        listener: tokio::net::TcpListener,
+    },
+    Unix {
+        listener: tokio::net::UnixListener,
+        user_state: futures_util::lock::Mutex<
```
arsing, line 35: This doesn't need to be an async mutex. Just use `std::sync::Mutex`.

http-common/Cargo.toml
massand: Fixed

http-common/src/connector.rs
massand: Fixed

http-common/src/connector.rs
```diff
@@ -80,8 +80,8 @@ impl Incoming {
                 let (unix_stream, _) = listener.accept().await?;
                 let ucred = unix_stream.peer_cred()?;
                 let user_state = user_state
-                    .lock()
-                    .await
+                    .get_mut()
```
arsing, line 83: Oh, this already receives `&mut self`, so even a `std::sync::Mutex` is unnecessary. You can just have `user_state: BTreeMap` directly.

massand: Good catch. Fixed.

arsing (approved):

PR #212, opened 2021-05-01
CN fallback for X.509 registration

massand:
In 1.1, iotedge had a fallback mechanism for DPS X.509 attestation. In 1.2, however, new options for obtaining device identity certificate became available. In the new world, CS could be configured to issue identity certificate in one of these configurations:

1. receive a CSR with any subject name. This subject would be IS's config.toml registration_id value. 
2. receive a CSR and common name is configured within CS config.toml. This will override any common name specified in the CSR.

With this fix, the behavior is changed to bring back parity with 1.1 behavior (i.e. if registration_id is missing, fallback to identity cert's subject name as registration_id). The CSR issued by IS to renew the identity cert will contain subject name set as one of the following values (in order of precedence): 
- First, the IS config.toml X.509 attestation registration_id is used in the CSR to renew the identity cert
- If registration_id is missing, the current identity certificate's subject name is used in the CSR to renew the identity cert
- When the identity certificate is returned, the latest valid/renewed cert's subject is used, if the registration_id is missing.

Note that if registration_id is set to a value different from the identity certificate subject name returned by CS, DPS will reject the registration request (as of DPS API version 2019-04-15). So, it is upto the user to avoid this scenario currently. 

TODO:

- [x] Validate E2E manually
- [x] Send a PR for iotedge import tool changes

### DIFF
--- a/identity/aziot-identityd/src/identity.rs
+++ b/identity/aziot-identityd/src/identity.rs
@@ -628,8 +628,8 @@ impl IdentityManager {
                         let registration_id = match registration_id {
                             Some(registration_id) => registration_id,
                             None => match cert_subject_name {
-                                std::option::Option::Some(cert_subject_name) => cert_subject_name?,
-                                std::option::Option::None => {
+                                Some(cert_subject_name) => cert_subject_name?,
+                                None => {
                                     return Err(Error::Internal(InternalError::CreateCertificate(
                                                 "device identity certificate subject is required to register with DPS, but it is not configured".to_string().into())));
                                 }
@@ -814,8 +814,8 @@ impl IdentityManager {
             let new_cert_subject = match subject {
                 Some(subject) => subject.to_string(),
                 None => match old_cert_subject_name {
-                    std::option::Option::Some(sn) => sn?,
-                    std::option::Option::None => {
+                    Some(sn) => sn?,
+                    None => {
                         return Err(Error::Internal(InternalError::CreateCertificate(
                                     "device identity certificate subject is required to register with DPS, but it is not configured".to_string().into())));
                     }
--- a/identity/aziot-identityd/src/identity.rs
+++ b/identity/aziot-identityd/src/identity.rs
@@ -627,13 +627,9 @@ impl IdentityManager {
 
                         let registration_id = match registration_id {
                             Some(registration_id) => registration_id,
-                            None => match cert_subject_name {
-                                Some(cert_subject_name) => cert_subject_name?,
-                                None => {
-                                    return Err(Error::Internal(InternalError::CreateCertificate(
-                                                "device identity certificate subject is required to register with DPS, but it is not configured".to_string().into())));
-                                }
-                            },
+                            None => cert_subject_name.map_err(|err| {
+                                Error::Internal(InternalError::CreateCertificate(Box::new(err)))
+                            })?,
                         };
 
                         let dps_auth_kind = aziot_dps_client_async::DpsAuthKind::X509 {
@@ -752,7 +748,7 @@ impl IdentityManager {
         identity_pk: &str,
         identity_cert: &str,
         subject: Option<&String>,
-    ) -> Result<Option<Result<String, Error>>, Error> {
+    ) -> Result<Result<String, Error>, Error> {
         // Retrieve existing cert and check it for expiry.
         let (device_id_cert, old_cert_subject_name) = match self
             .cert_client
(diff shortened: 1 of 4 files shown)


identity/aziot-identityd/src/identity.rs
```diff
@@ -553,12 +553,14 @@ impl IdentityManager {
                         identity_cert,
                         identity_pk,
                     } => {
-                        self.create_identity_cert_if_not_exist_or_expired(
-                            &identity_pk,
-                            &identity_cert,
-                            &device_id,
-                        )
-                        .await?;
+                        //IoT Hub doesn't require device ID to match identity certificate CN
```
gordonwang0, line 556: Nit: Should be space after `//`. Same with line 619 below.

identity/aziot-identityd/src/identity.rs
```diff
@@ -614,12 +616,25 @@ impl IdentityManager {
...
+                        //DPS requires registration ID to match identity certificate CN
+                        let cert_subject_name = self
+                            .create_identity_cert_if_not_exist_or_expired(
+                                &identity_pk,
+                                &identity_cert,
+                                registration_id.as_deref().to_owned(),
+                            )
+                            .await?;
+
+                        let registration_id = match registration_id {
```
gordonwang0, line 628: This block is more concise with `unwrap_or_else`.

identity/aziot-identityd/src/identity.rs
```diff
@@ -736,15 +751,33 @@ impl IdentityManager {
...
+        {
             Ok(pem) => {
                 let cert = openssl::x509::X509::from_pem(&pem).map_err(|err| {
                     Error::Internal(InternalError::CreateCertificate(Box::new(err)))
                 })?;
                 let cert_expiration = cert.as_ref().not_after();
+                let cert_subject = cert.as_ref().subject_name();
+                let cert_subject = match cert_subject.entries_by_nid(openssl::nid::Nid::COMMONNAME).next() {
+                    Some(common_name_entry) => {
+                        match String::from_utf8(common_name_entry.data().as_slice().into()) {
```
gordonwang0, line 770: This can be expressed more concisely with `map_err`.

identity/aziot-identityd/src/identity.rs
```diff
@@ -736,15 +751,33 @@ impl IdentityManager {
         &self,
         identity_pk: &str,
         identity_cert: &str,
-        subject: &str,
-    ) -> Result<(), Error> {
+        subject: Option<&str>,
```
gordonwang0, line 754: This should be `Option<String>`, since `registration_id` in the config is already `Option<String>`.

identity/aziot-identityd/src/identity.rs
```diff
@@ -736,15 +751,33 @@ impl IdentityManager {
...
+    ) -> Result<Option<Result<String, Error>>, Error> {
```
gordonwang0, line 755: Can this be flattened to just `Result<String, Error>`?

The `Ok(String)` value is the subject name, which must come from either the old cert or the config. Is there ever a scenario where the subject name won't be known by this function?

gordonwang0 (changes requested):

identity/aziot-identityd/src/identity.rs
massand: The subject name doesn't need to exist in the manual X.509 provisioning case. Since this logic is shared between both DPS and IoT Hub, we need to late-evaluate this result.

identity/aziot-identityd/src/identity.rs
massand: The unwrap_or_else doesn't permit lazy evaluatation of result of create_identity_cert_if_not_exist_or_expired. This is intentional

identity/aziot-identityd/src/identity.rs
massand: Fixed here and below.

identity/aziot-identityd/src/identity.rs
massand: Fixed the match at cert_subject level

identity/aziot-identityd/src/identity.rs
massand: Fixed

identity/aziot-identityd/src/identity.rs
gordonwang0: In X509 manual, this function is called with the subject parameter is set to `Some(&device_id)`. So even in that case, the subject name exists, right?

massand: The result isn't used in the manual x509 case. The `device_id` is only used to issue a new identity certificate (in case CS has EST method configured for `device-id` certificate).

gordonwang0: Which means it should be fine to flatten the return type to `Result<String, Error>`, right?
- In the manual X509 case, it's not used so doesn't matter
- In the DPS case, the parsing becomes simpler (currently you have nested matches to unwrap it).

massand: Flattened it to Result<Result<String, Error>> since we discussed the need for passing back a lazy-evaluated result.

gordonwang0:
LGTM. Just need to resolve the merge conflict and I'll sign off.

gordonwang0 (approved):

PR #213, opened 2021-05-03
Delete and regenerate private keys for new certs.

arsing:
This change defines libaziot_keys v2.1.0.0 with two new functions to delete keys
and key pairs respectively, then exports them via new keyd HTTP API under
a new API version 2021-05-01.

Secondly, this commit changes the code paths in certd and identityd that
create new certs for themselves - EST id cert in certd and
device-id cert in identityd - to delete the existing key first.

Deleting a key is a no-op success if the key doesn't exist, or if
the libaziot_keys version is less than v2.1.0.0


docs/keys-service.md
```diff
@@ -84,9 +84,31 @@ Required. See [API authentication](#api-authentication).
 
 ---
 
+### Delete Symmetric Key
+
+`DELETE /key?api-version=2021-05-01`
+
+#### Authentication
+
+Not required.
```
gordonwang0, line 93: Should these delete APIs require authentication? AFAIK all other key R/W operations do.

key/aziot-keyd/src/keys.rs
```diff
@@ -436,6 +610,44 @@ impl std::fmt::Display for GetKeyPairPublicParameterError {
...
 
+impl Keys {
+    pub(crate) fn delete_key_pair(
+        &mut self,
+        id: &std::ffi::CStr,
+    ) -> Result<(), DeleteKeyPairError> {
+        unsafe {
+            match self {
+                Keys::V2_0_0_0 { .. } => {
+                    // Old version don't support deleting key pairs, so treat it as a no-op.
```
gordonwang0, line 621: Old version *doesn't* support...

Same on line 755 below.

gordonwang0 (changes requested):
Could you also implement these todos?

https://github.com/Azure/iot-identity-service/blob/fcaade3b329994e2818a958f7f83c3216c5e075f/identity/aziot-identityd/src/identity.rs#L763-L768

https://github.com/Azure/iot-identity-service/blob/fcaade3b329994e2818a958f7f83c3216c5e075f/identity/aziot-identityd/src/identity.rs#L806-L810

docs/keys-service.md
arsing: They don't require authentication because they need a key handle. Same as encrypt, etc.

key/aziot-keyd/src/keys.rs
arsing: Fixed.

arsing:
>Could you also implement these todos?

I deleted them. Next time around the keys will get deleted anway.

docs/keys-service.md
gordonwang0: But is allowing anyone with the key handle to delete the key safe? I suppose we do trust whoever has the key handles since GET requires authentication.

arsing:
We want to add a more fine-grained permissions model eventually anyway - to limit a key handle can be used for in terms of caller UID and op. So "delete" would be one such op.
________________________________

gordonwang0 (approved):

PR #215, opened 2021-05-10
Documentation overhaul.

arsing:
This refactors the `docs/` directory into `docs/` and `docs-dev/` for
user documentation and developer documentation,
with the following overall motivations:

- Remove developer documentation from HTML documentation so that users do not
  land on it. Ideally this would be something like `docs/user/` and `docs/dev/`
  but GitHub does not allow using anything except `docs/` for the root.

- Change PKCS#11 library documentation to focus exclusively on the case of
  users installing it in production. This means replacing mentions of
  "the user you will run aziot-keyd as" with "aziotks", etc.

- Don't link to HTTP API docs from the index since they're only useful
  for host module authors. They're linked from the host-module doc instead.

  Also, the TPM service is exclusively an internal service, so move it to
  developer docs.

- Move `packaging.md` to developer docs since the user does not need
  the level of detail it provides about what the package contains.
  The user docs only cover the super-config and a new page about
  the `aziotctl` binary.

- Hard-code the nav bar instead of letting Jekyll generate it, so that
  it preserves the order of pages rather than sort them alphabetically.


arsing:
Generated: https://arsing.github.io/iot-identity-service/

gordonwang0 (approved):

PR #216, opened 2021-05-12
Fix Identity Service when starting offline

gordonwang0:
Fixes Identity Service when starting offline:
- Fix a bug where device backups weren't written unless at least one module was specified.
- Abort IS if provisioning fails and no backup can be read. This should be considered a fatal error because none of the APIs will work without device information.
- Fix documentation for /identities/device.
- Adds the `cloud_timeout_sec` and `cloud_retries` options to Identity Service config.

### DIFF
--- a/aziotctl/aziotctl-common/src/config/apply.rs
+++ b/aziotctl/aziotctl-common/src/config/apply.rs
@@ -24,6 +24,8 @@ pub fn run(
         parent_hostname,
         provisioning,
         localid,
+        cloud_timeout_sec,
+        cloud_retries,
         mut aziot_keys,
         mut preloaded_keys,
         cert_issuance,
@@ -267,9 +269,9 @@ pub fn run(
 
         homedir: super::AZIOT_IDENTITYD_HOMEDIR_PATH.into(),
 
-        cloud_timeout_sec: aziot_identityd_config::Settings::default_cloud_timeout(),
+        cloud_timeout_sec,
 
-        cloud_retries: aziot_identityd_config::Settings::default_cloud_retries(),
+        cloud_retries,
 
         principal: vec![],
 
--- a/aziotctl/aziotctl-common/src/config/apply.rs
+++ b/aziotctl/aziotctl-common/src/config/apply.rs
@@ -267,6 +267,10 @@ pub fn run(
 
         homedir: super::AZIOT_IDENTITYD_HOMEDIR_PATH.into(),
 
+        cloud_timeout_sec: aziot_identityd_config::Settings::default_cloud_timeout(),
+
+        cloud_retries: aziot_identityd_config::Settings::default_cloud_retries(),
+
         principal: vec![],
 
         provisioning,
--- a/http-common/src/connector.rs
+++ b/http-common/src/connector.rs
@@ -49,15 +49,14 @@ impl Incoming {
             + 'static,
         <H as hyper::service::Service<hyper::Request<hyper::Body>>>::Future: Send,
     {
-        const READ_WRITE_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(5);
+        const READ_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(5);
         const MAX_REQUESTS_PER_USER: usize = 10;
 
         match self {
             Incoming::Tcp { listener } => loop {
                 let (tcp_stream, _) = listener.accept().await?;
-                let mut timeout_stream = tokio_io_timeout::TimeoutStream::new(tcp_stream);
-                timeout_stream.set_read_timeout(Some(READ_WRITE_TIMEOUT));
-                timeout_stream.set_write_timeout(Some(READ_WRITE_TIMEOUT));
+                let mut timeout_stream = tokio_io_timeout::TimeoutReader::new(tcp_stream);
+                timeout_stream.set_timeout(Some(READ_TIMEOUT));
 
                 // TCP is available in test builds only (not production). Assume current user is root.
                 let server = crate::uid::UidService::new(0, server.clone());
@@ -83,9 +82,8 @@ impl Incoming {
                         std::sync::Arc::new(tokio::sync::Semaphore::new(MAX_REQUESTS_PER_USER))
                     })
                     .clone();
-                let mut timeout_stream = tokio_io_timeout::TimeoutStream::new(unix_stream);
-                timeout_stream.set_read_timeout(Some(READ_WRITE_TIMEOUT));
-                timeout_stream.set_write_timeout(Some(READ_WRITE_TIMEOUT));
+                let mut timeout_stream = tokio_io_timeout::TimeoutReader::new(unix_stream);
+                timeout_stream.set_timeout(Some(READ_TIMEOUT));
 
                 let server = crate::uid::UidService::new(ucred.uid(), server.clone());
                 tokio::spawn(async move {
--- a/http-common/src/connector.rs
+++ b/http-common/src/connector.rs
@@ -49,20 +49,17 @@ impl Incoming {
             + 'static,
         <H as hyper::service::Service<hyper::Request<hyper::Body>>>::Future: Send,
     {
-        const READ_TIMEOUT: std::time::Duration = std::time::Duration::from_secs(5);
         const MAX_REQUESTS_PER_USER: usize = 10;
 
         match self {
             Incoming::Tcp { listener } => loop {
                 let (tcp_stream, _) = listener.accept().await?;
-                let mut timeout_stream = tokio_io_timeout::TimeoutReader::new(tcp_stream);
-                timeout_stream.set_timeout(Some(READ_TIMEOUT));
 
                 // TCP is available in test builds only (not production). Assume current user is root.
                 let server = crate::uid::UidService::new(0, server.clone());
                 tokio::spawn(async move {
                     if let Err(http_err) = hyper::server::conn::Http::new()
-                        .serve_connection(Box::pin(timeout_stream), server)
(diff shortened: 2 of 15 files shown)


identity/aziot-identityd/src/identity.rs
```diff
@@ -893,6 +893,14 @@ impl IdentityManager {
                         std::collections::BTreeSet::default()
                     };
 
+                // Write out device state and settings.
+                // This overwrites any existing device state and settings backup.
+                std::fs::write(prev_device_info_path, device_status)
+                    .map_err(|err| Error::Internal(InternalError::SaveDeviceInfo(err)))?;
+
+                std::fs::write(prev_settings_path, &settings_serialized)
```
arsing, line 901: If creating module identities after saving the backup fails, identityd restarts, won't it think module identities already exist in cloud because the backup matches the settings, and thus it'll skip trying to create them again? I assume that's why the cloud stuff was done before saving the backup.

So rather than move this up, you could just return the early return altogether. In fact, currently it allows cloud to have extra identities that aren't present in the config or in the backup; removing the early return would allow cleaning those up too.

identity/aziot-identityd/src/lib.rs
```diff
@@ -51,7 +51,7 @@ macro_rules! match_id_type {
     };
 }
 
-#[derive(Debug)]
+#[derive(Clone, Debug)]
```
arsing, line 54: Derive `Copy` too because all the variants are empty, and then you won't need to `.clone()` it explicitly.

arsing (changes requested):

identity/aziot-identityd/src/lib.rs
gordonwang0: Fixed.

identity/aziot-identityd/src/identity.rs
gordonwang0: Fixed.

I remember discussing this with Rajeev and he mentioned there might be a scenario where users create identities in IoT Hub not managed by Identity Service, and we don't want IS to delete those. So that's why there's an early return there. Does that make sense?

gordonwang0:
@arsing Dylan was testing this and found that when offline, getting module identities doesn't work.

I'm able to repro:
```
sudo curl --unix-socket /run/aziot/identityd.sock http://localhost/identities/modules/SimulatedTemperatureSensor?api-version=2020-09-01\&type=aziot -H "content-type: application/json"
```

Logs:
```
May 13 10:35:19 gowa-virtualbox-vm aziot-identityd[16952]: 2021-05-13T17:35:19Z [INFO] - <-- GET /identities/modules/SimulatedTemperatureSensor?api-version=2020-09-01&type=aziot {"host": "localhost", "user-agent": "curl/7.58.0", "accept": "*/*", "content-type": "application/json"}
May 13 10:35:24 gowa-virtualbox-vm aziot-identityd[16952]: 2021-05-13T17:35:24Z [INFO] - Error while serving HTTP connection: connection error: timed out
```

This comes from:
https://github.com/Azure/iot-identity-service/blob/81b066b49104821025e65bde5798628e5fed194c/http-common/src/connector.rs#L94-L99

Any idea on why this happens? This only happens when offline (I put my laptop in airplane mode).

arsing:
Is the API itself taking longer than the timeout to compute a response?

Rajeev implemented the timeout for two things. One is for the difference between any two `read()`s from the client. The other for how long it takes to read the client's request body overall.

Now, if the server itself takes longer than the timeout to compute a response *after* it's received the request in its entirety, the assumption is that the first timeout should *not* fire. Can you verify that assumption?

http-common/src/connector.rs
```diff
@@ -83,16 +79,13 @@ impl Incoming {
...
-                timeout_stream.set_read_timeout(Some(READ_WRITE_TIMEOUT));
-                timeout_stream.set_write_timeout(Some(READ_WRITE_TIMEOUT));
 
                 let server = crate::uid::UidService::new(ucred.uid(), server.clone());
                 tokio::spawn(async move {
                     match user_state.try_acquire_owned() {
                         Ok(_permit) => {
                             if let Err(http_err) = hyper::server::conn::Http::new()
-                                .serve_connection(Box::pin(timeout_stream), server)
+                                .serve_connection(Box::pin(unix_stream), server)
```
gordonwang0, line 88: @arsing I wasn't able to get TimeoutStream (either by changing it to TimeoutReader or only setting the read timeout) to respect the timeouts in IS. From my testing, the READ_WRITE_TIMEOUT (5 sec) always stopped whatever API call I was making and returned an error if that API call was taking longer than 5 sec. So I changed it back to streams that don't use timeouts.

identity/aziot-identityd/src/error.rs
```diff
@@ -21,6 +21,25 @@ impl Error {
...
         Error::InvalidParameter(name, err.into())
     }
+
+    pub(crate) fn is_network(&self) -> bool {
+        match &self {
+            Error::DpsClient(err) | Error::HubClient(err) => match err.kind() {
+                std::io::ErrorKind::TimedOut => true, // Network timed out
+                std::io::ErrorKind::Other => {
+                    if let Some(err) = err.get_ref() {
+                        // TODO: check if error is hyper::Error
```
gordonwang0, line 31: @arsing Rust noob question: How do I check that err is a hyper::Err here? err.get_ref() just gives something that impl std::err::Error. I want to do something like:

```
if let Some(err) = err.get_ref() {
    if err is hyper::Error {
        if err.is_connect() {
            return true;
        }
    }
}
```

arsing: [`.downcast_ref()`](https://doc.rust-lang.org/stable/std/error/trait.Error.html#method.downcast_ref) but it's an anti-pattern. When you want to look at inner errors, you *should* make the outer error hold the inner error strongly-typed. In this case that would mean replacing `std::io::Error` with `enum Error { Http(hyper::Error), ... }`

http-common/src/connector.rs
arsing: Okay. We can revisit it later.

identity/aziot-identityd/src/error.rs
gordonwang0: Got it. Actually I figured we can just map hyper's `err.is_connected()` to `std::io::ErrorKind::NotConnected`, so I did that.

http-common/src/connector.rs
```diff
@@ -49,21 +49,17 @@ impl Incoming {
...
-                let mut timeout_stream = tokio_io_timeout::TimeoutStream::new(tcp_stream);
-                timeout_stream.set_read_timeout(Some(READ_WRITE_TIMEOUT));
-                timeout_stream.set_write_timeout(Some(READ_WRITE_TIMEOUT));
 
                 // TCP is available in test builds only (not production). Assume current user is root.
                 let server = crate::uid::UidService::new(0, server.clone());
                 tokio::spawn(async move {
                     if let Err(http_err) = hyper::server::conn::Http::new()
-                        .serve_connection(Box::pin(timeout_stream), server)
+                        .serve_connection(Box::pin(tcp_stream), server)
```
arsing, line 62: Don't need `Box::pin()`: https://github.com/Azure/iot-identity-service/commit/fcaade3b329994e2818a958f7f83c3216c5e075f#diff-8bd6e0799ebba2043c7b5d189504fdd1d5166ccc415dd9ef6675f4a0b3aaf76fL55

Same for `unix_stream` below: https://github.com/Azure/iot-identity-service/commit/fcaade3b329994e2818a958f7f83c3216c5e075f#diff-8bd6e0799ebba2043c7b5d189504fdd1d5166ccc415dd9ef6675f4a0b3aaf76fL70

aziotctl/aziotctl-common/src/config/apply.rs
```diff
@@ -267,6 +267,10 @@ pub fn run(
 
         homedir: super::AZIOT_IDENTITYD_HOMEDIR_PATH.into(),
 
+        cloud_timeout_sec: aziot_identityd_config::Settings::default_cloud_timeout(),
```
arsing, line 270: You haven't added them to the super-config ?

identity/aziot-hub-client-async/src/lib.rs
```diff
@@ -141,328 +154,270 @@ impl Client {
...
+                        &self.cert_client,
+                        self.proxy_uri.clone(),
+                    )
+                    .await?
+                }
+            };
+
+            let client: hyper::Client<_, hyper::Body> = hyper::Client::builder().build(connector);
+
+            let err = match tokio::time::timeout(self.req_timeout, client.request(req)).await {
```
arsing, line 267: Can you put this in `http-common` to reuse it? It's already duplicated between hub-client and dps-client, and in the future certd might need it too.

arsing (changes requested):

http-common/src/connector.rs
gordonwang0: Fixed

aziotctl/aziotctl-common/src/config/apply.rs
gordonwang0: Fixed

identity/aziot-hub-client-async/src/lib.rs
gordonwang0: I tried merging them at first, but these functions each do some Hub or DPS-specific things and have some HubClient/DpsClient-specific parameters (such as self). It would be more rework to merge these with the common code.

arsing (approved):

dylanbronson (approved):
Tested locally on my machine and offline startup works

PR #218, opened 2021-05-13
Update tpm2-pkcs11.md

ksaye:
Without this change, you get the error:  Specified path is not a directory, got: /home/<username>/.tpm2_pkcs11


arsing:
Just remove the `/opt/tpm2-pkcs11` stuff entirely, including the env var. (The env var isn't being used anyway, because `sudo` won't inherit it.)

The default path for the sqlite3 db is `~/.tpm2-pkcs11/tpm2_pkcs11.sqlite3` which puts it under `/var/lib/aziot/keyd/`, which is as good a place as any.

arsing:
https://github.com/Azure/iot-identity-service/pull/219

PR #222, opened 2021-05-15
Ignore duplicate cert error when adding certs to trusted store in openssl 1.0

arsing:
This was originally found with IoT Edge 1.1 because it used an old version of
native-tls that also forwarded this error. IoT Edge fixed the issue by updating
its use of native-tls to a version that ignores this error
(and all other errors).

### DIFF
--- a/http-common/src/proxy.rs
+++ b/http-common/src/proxy.rs
@@ -143,7 +143,40 @@ fn identity_to_tls_connector(
 
     let cert_store = tls_connector.cert_store_mut();
     for trusted_cert in trusted_certs {
-        cert_store.add_cert(trusted_cert.clone())?;
+        if let Err(err) = cert_store.add_cert(trusted_cert.clone()) {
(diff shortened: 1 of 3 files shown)


http-common/src/proxy.rs
```diff
@@ -143,7 +143,40 @@ fn identity_to_tls_connector(
 
     let cert_store = tls_connector.cert_store_mut();
     for trusted_cert in trusted_certs {
-        cert_store.add_cert(trusted_cert.clone())?;
+        if let Err(err) = cert_store.add_cert(trusted_cert.clone()) {
```
darobs, line 146: You are getting rid of this code because you don't expect to support openssl 1.0?

arsing: I'm adding code because we support openssl 1.0

darobs: Geez - I shouldn't read reviews late on aFriday night apparently.

gordonwang0 (approved):

darobs (approved):

PR #224, opened 2021-05-20
Update ex_data.rs for Rust > 1.51

gordonwang0:
Removes workaround for edgelet that was using 1.47.

This reverts d539159c8d23719eeb303392f1062622971b8708


arsing:
Should we hold off on this until iotedge updates to 1.52.0, ie after the tokio 1.0 refactor? Or I guess based on your iotedge PR i-i-s already requires 1.52, so it doesn't matter?

gordonwang0:
Yeah, we made some changed in i-i-s for 1.52, so it shouldn't matter if we merge this now.

arsing (approved):

PR #227, opened 2021-05-25
Reject DPS results with status failed

gordonwang0:
Fail DPS provisioning if the response has status `failed`.

### DIFF
--- a/identity/aziot-identityd/src/identity.rs
+++ b/identity/aziot-identityd/src/identity.rs
@@ -715,26 +715,28 @@ impl IdentityManager {
             .await
             .map_err(Error::DpsClient)?;
 
-        let mut state = operation.registration_state.ok_or(Error::DeviceNotFound)?;
+        // DPS client registration won't return if the status is "assigning".
+        assert!(!operation.status.eq_ignore_ascii_case("assigning"));
 
-        if state.status.as_deref() == Some("failed") {
+        let state = operation.registration_state.ok_or(Error::DeviceNotFound)?;
+
+        if state.assigned_hub.is_none() || state.device_id.is_none() {
             return Err(Error::DpsClient(std::io::Error::new(
                 std::io::ErrorKind::Other,
                 state.error_message.unwrap_or_default(),
             )));
         }
 
-        let iothub_hostname = state.assigned_hub.get_or_insert("".into());
-        let device_id = state.device_id.get_or_insert("".into());
-        let device = aziot_identity_common::IoTHubDevice {
+        let iothub_hostname = state.assigned_hub.expect("assigned_hub should not be none");
+        let device_id = state.device_id.expect("device_id should not be none");
+
+        Ok(aziot_identity_common::IoTHubDevice {
             local_gateway_hostname: local_gateway_hostname
                 .unwrap_or_else(|| iothub_hostname.clone()),
-            iothub_hostname: iothub_hostname.clone(),
-            device_id: device_id.clone(),
+            iothub_hostname,
+            device_id,
             credentials,
-        };
-
-        Ok(device)
+        })
     }
 
     fn get_backup_provisioning_info(
--- a/identity/aziot-identityd/src/identity.rs
+++ b/identity/aziot-identityd/src/identity.rs
@@ -719,6 +719,14 @@ impl IdentityManager {
         assert!(!status.eq_ignore_ascii_case("assigning"));
 
         let mut state = operation.registration_state.ok_or(Error::DeviceNotFound)?;
+
+        if state.status == Some("failed".to_string()) {
+            return Err(Error::DpsClient(std::io::Error::new(
+                std::io::ErrorKind::Other,
+                state.error_message.unwrap_or_default(),
+            )));
+        }
+
         let iothub_hostname = state.assigned_hub.get_or_insert("".into());
         let device_id = state.device_id.get_or_insert("".into());
         let device = aziot_identity_common::IoTHubDevice {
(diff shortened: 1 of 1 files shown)


identity/aziot-identityd/src/identity.rs
```diff
@@ -719,6 +719,14 @@ impl IdentityManager {
         assert!(!status.eq_ignore_ascii_case("assigning"));
```
arsing, line 719: This should also be an `if` + `return Err` rather than a crashing `assert!`

identity/aziot-identityd/src/identity.rs
```diff
@@ -719,6 +719,14 @@ impl IdentityManager {
...
 
         let mut state = operation.registration_state.ok_or(Error::DeviceNotFound)?;
+
+        if state.status == Some("failed".to_string()) {
```
arsing, line 723: `state.status.as_deref() == Some("failed")`, to avoid allocating a String just for the comparison.

arsing, line 723: [1.1's code](https://github.com/Azure/iotedge/blob/release/1.1/edgelet/dps/src/registration.rs#L188) seems to ignore this field and just assume that the presence of `.registration_state` means registration succeeded. In this code that would be handled by line 721 above. Are you seeing a case where registration fails but `registration_state` is still set to a valid device registration result?

arsing (changes requested):

identity/aziot-identityd/src/identity.rs
gordonwang0: I think Rajeev made it an assert because `dps_client.register` above doesn't return unless the status is not `assigning`. So it's expected to never be `assigning`. I'll remove the check though, it doesn't do much.

identity/aziot-identityd/src/identity.rs
gordonwang0: Fixed

gordonwang0: Yes, I was testing something else when I noticed if you change the IoT Hub access key to something invalid, you'll get a valid DeviceRegistrationResult:

```
DeviceRegistrationResult {
    tpm: None,
    x509: None,
    symmetric_key: None,
    registration_id: Some(
        "gowa-virtualbox-vm-1",
    ),
    created_date_time_utc: Some(
        "2021-05-25T17:50:50.1952288Z",
    ),
    assigned_hub: None,
    device_id: None,
    status: Some(
        "failed",
    ),
    substatus: None,
    error_code: Some(
        401002,
    ),
    error_message: Some(
        "{\"Message\":\"{\\\"errorCode\\\":401002,\\\"trackingId\\\":\\\"24821dbe7957463a853eefa84f067e0d-G:2-TimeStamp:05/25/2021 17:50:50\\\",\\\"message\\\":\\\"Unauthorized\\\",\\\"timestampUtc\\\":\\\"2021-05-25T17:50:50.1928721Z\\\"}\",\"ExceptionMessage\":\"\"}",
    ),
    last_updated_date_time_utc: Some(
        "2021-05-25T17:50:50.5209447Z",
    ),
    etag: Some(
        "Ijc0MDUyNTZhLTAwMDAtMDgwMC0wMDAwLTYwYWQzOGZhMDAwMCI=",
    ),
}
```

arsing: Okay. So given that `assigned_hub` and `device_id` are still `None`, I would've thought provisioning would fail for *that* reason. But I see that lines 727-728 default to `""` in that case. Is that supposed to be done?

identity/aziot-identityd/src/identity.rs
arsing: Ah, in that case I'd rather keep the assert actually, and also add a comment with that explanation.

identity/aziot-identityd/src/identity.rs
gordonwang0: Yes, provisioning will fail later on because of the device and hub being None.

But you'll get some misleading log messages in the meantime. Because this function thinks it succeeded, it'll print "Successfully reprovisioned." when it really hasn't. So I think it's better to catch the provisioning error here.

I'll change it to not default to "" though. It probably doesn't make sense to continue without knowing the device and hub names.

arsing: >I'll change it to not default to "" though. It probably doesn't make sense to continue without knowing the device and hub names.

Yes, that's my point. If they're `.ok_or(...)?` instead then you still get the desired outcome of having the function fail, and it'd be "more correct" because those fields shouldn't be `null` in any case.

identity/aziot-identityd/src/identity.rs
gordonwang0: Ok, fixed.

identity/aziot-identityd/src/identity.rs
gordonwang0: Fixed

identity/aziot-identityd/src/identity.rs
```diff
@@ -715,26 +715,28 @@ impl IdentityManager {
...
             return Err(Error::DpsClient(std::io::Error::new(
                 std::io::ErrorKind::Other,
                 state.error_message.unwrap_or_default(),
             )));
         }
 
-        let iothub_hostname = state.assigned_hub.get_or_insert("".into());
-        let device_id = state.device_id.get_or_insert("".into());
-        let device = aziot_identity_common::IoTHubDevice {
+        let iothub_hostname = state.assigned_hub.expect("assigned_hub should not be none");
```
arsing, line 730:
```rust
let (iothub_hostname, device_id) = match (state.assigned_hub, state.device_id) {
    (Some(iothub_hostname, Some(device_id)) => (iothub_hostname, device_id),
    _ => return Err(...),
};
```

gordonwang0: Thanks, fixed.

arsing (approved):

Issue #56, opened 2020-10-24
/identities/modules response is missing the "auth" section


micahl:
Per the [docs for the /identity/modules request](https://github.com/Azure/iot-identity-service/blob/main/docs/identity-service.md#response-sas-case-1) each module identity in the response should include the below, but it isn't. 

**Note:** I do see it in the response from "/identities/identity" and "/identities/modules/{module-id}"

```json
        "auth": {
            "type": "sas",
            "keyHandle": "string"
        }
```

arsing:
It should definitely not include the `keyHandle` because it would be inappropriate. Rajeev can decide whether `type` and `certId` are worth including. (They're not private information but IMO it's still not worth including it, so it would be fine to remove it from the docs to match the implementation.)

micahl:
Would it depend on whether the idtype was set to device or module.  For example, if it's device then the info is included, but if module then not?

arsing:
Would what depend? The `keyHandle` should not be in list responses period, if that's what you're asking.

PR #232, opened 2021-05-29
Update IS api docs

micahl:
Drop the API to update module identity
Change docs to resolve issue #56


gordonwang0:
Are we removing support for this API?

docs/api/identity-service.md
```diff
@@ -162,10 +162,6 @@ The `type` query parameter specifies the identity type to return. Accepted value
         "deviceId": "device01",
         "moduleId": "module01",
         "genId": "12345",
-        "auth": {
```
gordonwang0, line 165: AFAIK, this key is still in the response.

micahl: It's absent in the response I received when testing this API and AFAIK it has never been included per this [old issue](https://github.com/Azure/iot-identity-service/issues/56).  I don't believe it's required so to resolve the issue I'm updating the docs to align with the implementation.

micahl:
Unless we have a clear use case that requires the update module API I don't think we should advertise it and consider whether the code should remain.

gordonwang0:
Do we know why that API exists in the first place?

I'd like to keep the docs and code in sync, so we should remove the docs with the code.

docs/api/identity-service.md
gordonwang0: The code to include it exists though. I think it will be present in certain cases (though I haven't tried it).

micahl:
> Do we know why that API exists in the first place?
> 
> I'd like to keep the docs and code in sync, so we should remove the docs with the code.

OK, I'll add it back and then open a tracking issue on the repo as a reminder to figure out why the code exists or to clean up the doc.

docs/api/identity-service.md
```diff
@@ -363,6 +354,11 @@ The `type` query parameter specifies the identity type to return. Accepted value
 The `type` query parameter specifies the identity type to return. Accepted values are:
 - `aziot`: Module identity.
 
+#### Request
+```json
```
gordonwang0, line 358: I don't think we need to send a request body for this API.

docs/api/identity-service.md
```diff
@@ -162,10 +162,6 @@ The `type` query parameter specifies the identity type to return. Accepted value
         "deviceId": "device01",
         "moduleId": "module01",
         "genId": "12345",
-        "auth": {
-            "type": "sas",
-            "keyHandle": "string"
```
varunpuranik, line 167: So list does not include auth info, but get does? Why is that?

docs/api/identity-service.md
micahl: I included it because I was getting errors when I tried to test the API without it.  Perhaps I'm doing something wrong, but even when I tried to pass no body I'd get `{"message":"EOF while parsing a value at line 1 column 0"}`.  
```bash
curl -X PUT "http://localhost:8901/identities/modules/module01?api-version=2020-09-01&type=aziot" -H "Content-Type: application/json" -H "Content-Length: 0"
```
Whereas it worked fine when with an empty body explicitly given:
```bash
curl -X PUT "http://localhost:8901/identities/modules/module01?api-version=2020-09-01&type=aziot" -H "Content-Type: application/json" --data "{}"
```

docs/api/identity-service.md
micahl: Yeah, that was at the root of my question when I opened #56.  @arsing?

docs/api/identity-service.md
arsing: `Route::put` takes `Self::PutBody`, so a JSON body is required. The route uses `IgnoredAny` so the body is ignored, but it's required.

We can make `Route::put` take `Option<Self::PutBody>` instead just like `post` and `delete` do, and update `make_service!` accordingly, to support optional bodies.

arsing:
>Do we know why that API exists in the first place?

@gordonwang Was this ever resolved?

arsing:
>Do we know why that API exists in the first place?

@gordonwang0 Was this ever resolved?

gordonwang0:
Yeah, I think we can close this.

gordonwang0:
We figured out that we do need the PUT API. So just check that the doc for it is correct, then we can merge.

arsing:
@gordonwang0

>We figured out that we do need the PUT API. So just check that the doc for it is correct, then we can merge.

Well, it looks correct to me, so if looks correct to you, can you approve and merge it?

arsing:
RHEL 8 build failure will be resolved by https://github.com/Azure/iot-identity-service/pull/416

PR #234, opened 2021-06-03
Add all IS APIs to client

gordonwang0:
Previously, aziot-identity-client-async only had the reprovision API because that was the only one used.

This PR adds the rest of the IS APIs to the client since they'll all be needed in edgelet.

### DIFF
--- a/identity/aziot-identity-client-async/src/lib.rs
+++ b/identity/aziot-identity-client-async/src/lib.rs
@@ -161,11 +161,7 @@ impl Client {
         let res: get_module_identities::Response = http_common::request::<(), _>(
             &self.inner,
             http::Method::GET,
-            make_uri!(
-                "/identities/modules",
-                self.api_version,
-                ID_TYPE_AZIOT
-            ),
+            make_uri!("/identities/modules", self.api_version, ID_TYPE_AZIOT),
             None,
         )
         .await?;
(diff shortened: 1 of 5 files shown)


identity/aziot-identity-client-async/src/lib.rs
```diff
@@ -4,37 +4,212 @@
...
+        )
+    };
+    ($path:literal, $api_version:expr, $type:expr) => {
+        &format!(
+            "http://identityd.sock{}?api-version={}&type={}",
+            $path, $api_version, $type
+        )
+    };
+    ($path:literal, $api_version:expr, $type:expr, $name:ident) => {
+        &format!(
```
arsing, line 27: You need to URL-encode the components, at least the `$name` since it's arbitrary.

arsing (changes requested):

identity/aziot-identity-client-async/src/lib.rs
gordonwang0: Good catch. Fixed.

identity/aziot-identity-client-async/src/lib.rs
```diff
@@ -4,37 +4,218 @@
...
+        &self,
+        module_name: &str,
+    ) -> Result<Identity, std::io::Error> {
+        let res: update_module_identity::Response = http_common::request::<(), _>(
+            &self.inner,
+            http::Method::PUT,
+            make_uri!(
+                "/identities/modules",
+                self.api_version,
+                ID_TYPE_AZIOT.to_string(),
```
arsing, line 150: This `.to_string()` isn't needed. Same for the next three fns.

arsing (changes requested):

identity/aziot-identity-client-async/src/lib.rs
gordonwang0: Fixed

arsing (approved):

PR #235, opened 2021-06-04
Fix broken plantuml files

micahl:


micahl:
@eustacea fyi

arsing:
I took the diagram out because nobody could make sense of it and kept asking us questions about it, and it's not important to users of the service.

micahl:
@arsing what questions were asked?

arsing:
I don't think I have the emails any more, but I remember "What is the HLMClient and where can I find it?" "Does newclient mean I have to write a client?" "What is MR and why is it a legacy service?"

The diagram goes into too much detail that is only needed for dev docs. For users the only they need to know about is that IS, KS and CS exist. Also, the edged, EA and EH parts don't belong in this repo's docs.

(I've already removed TPM service's API docs from the web docs and am thinking of downplaying / removing it even further, because people keep thinking that to use TPMs they need to talk to the TPMS instead of KS.)

micahl:
Dropping the reference to the diagram.  Submitting this PR just to fix the old plantuml files which for some reason appear to be missing the opening/ending tags.

arsing:
The move of docs-dev/img/new-component-overview-simple.svg to docs/img/new-component-overview-simple.svg also needs to be reverted.

arsing (approved):

micahl:
@arsing / @gordonwang0 is the failure with Mariner 2 something that is already known and any reason I should wait for that to be resolved before merging this?

arsing:
@josephknierman Do you know what hostname `pkgworker` is trying to resolve, and what we can do about it?

josephknierman:
@arsing  I think that is due to an issue Mariner 2.0 has in the preview PMC. This line https://github.com/Azure/iot-identity-service/blob/777d313844b821bd9e31935d74162f9cc4c11dff/ci/package.sh#L147 needs to be changed to =n instead of =y I just did a similar fix to a PR in iotedge which seems to have fixes the name resolution issue.

PR #237, opened 2021-06-09
Enable sockets in CentOS package

gordonwang0:
Enables service sockets during postinstall.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -270,7 +270,7 @@ rpm:
 	sed -e 's/@version@/$(PACKAGE_VERSION)/g; s/@release@/$(PACKAGE_RELEASE)/g' contrib/centos/aziot-identity-service.spec >$(RPMBUILDDIR)/SPECS/aziot-identity-service.spec
 
 	# Copy preset file to be included in the package
-	cp contrib/centos/00-aziot.preset $(RPMBUILDDIR)
+	cp contrib/centos/00-aziot.preset $(RPMBUILDDIR)/SOURCES
 
 	# Build package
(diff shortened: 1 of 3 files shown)


arsing:
You don't want to do it with the preset file thing?

gordonwang0:
I figured adding the enable to an existing file was easier. If you think the preset file is better, I'll change it.

arsing:
The preset file looks to be the same amount of work (four lines in `/usr/lib/systemd/system-preset/50-aziot-identity-service.preset` ) and appears to be "technically" more correct. My RPM-using distro doesn't ship any preset files and all services need to be enabled manually, but based on Fedora's docs it seems to be the preferred method. Dropping files is nicer that runtime commands and all that.

Makefile
```diff
@@ -269,6 +269,9 @@ rpm:
 	mkdir -p $(RPMBUILDDIR)/SPECS
 	sed -e 's/@version@/$(PACKAGE_VERSION)/g; s/@release@/$(PACKAGE_RELEASE)/g' contrib/centos/aziot-identity-service.spec >$(RPMBUILDDIR)/SPECS/aziot-identity-service.spec
 
+	# Copy preset file to be included in the package
+	cp contrib/centos/00-aziot.preset $(RPMBUILDDIR)
```
gordonwang0, line 273: @arsing I wasn't exactly sure where to put this; if you know a more suitable place, please suggest.

arsing: Extra source files like patches or this go in SOURCES/, next to the regular tarball.

gordonwang0: Ok, moved it.

arsing (approved):

varunpuranik (approved):

PR #238, opened 2021-06-10
Enable sockets in CentOS package (#237)

gordonwang0:
Enables service sockets during postinstall.


gordonwang0:
Didn't realize adding the ready-to-merge label would cause kodiak to automerge...

This is just a cherry-pick of the commit from main, no changes.

PR #240, opened 2021-06-17
Revert build to use Rust 1.47

gordonwang0:
This is a temporary reversion so that the code in this repo is compatible with iotedge.

This reverts:
- f733226 Update ex_data.rs for Rust > 1.51 (#224)
- d9895b5 Fix build for new Rust stable 1.52.0 (#214)

### DIFF
--- a/rust-toolchain
+++ b/rust-toolchain
@@ -1,1 +1,3 @@
-1.47.0
+[toolchain]
+channel = "1.47"
+components = ["clippy", "rustfmt"]
--- a/rust-toolchain
+++ b/rust-toolchain
@@ -1,3 +1,1 @@
-[toolchain]
-channel = "stable"
-components = ["clippy", "rustfmt"]
+1.47.0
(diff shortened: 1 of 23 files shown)


rust-toolchain
```diff
@@ -1,3 +1 @@
-[toolchain]
-channel = "stable"
-components = ["clippy", "rustfmt"]
+1.47
```
arsing, line 1: Keep the new format. Just change the `channel` line.

aziotctl/aziotctl-common/src/config/apply.rs
```diff
@@ -117,7 +117,7 @@ pub fn run(
                                         &identity_cert.method
                                     {
                                         set_est_auth(
-                                            auth.as_ref(),
+                                            &auth,
```
arsing, line 120: Only revert the `str::split_once` part of this commit. The rest is okay.

arsing (changes requested):
Please list all commit hashes in the PR description so that they make it into the commit message, like https://github.com/Azure/iot-identity-service/commit/24abe8711a9274098d59d03cf9128d5541de94a8 did.

rust-toolchain
gordonwang0: Fixed

aziotctl/aziotctl-common/src/config/apply.rs
gordonwang0: Fixed. Had to also remove some clippy changes.

arsing (approved):

PR #246, opened 2021-06-28
Bump missed version number for e2e tests

lfitchett:
The e2e tests are looking for the current version 1.2.1, but the github actions is still tagging the newest builds as 1.2.0. This causes the tests to break

### DIFF
--- a/.github/workflows/packages.yaml
+++ b/.github/workflows/packages.yaml
@@ -45,7 +45,7 @@ jobs:
           '/src/ci/package.sh'
       env:
         ARCH: "${{ matrix.arch }}"
-        PACKAGE_VERSION: '1.2.0'
+        PACKAGE_VERSION: '1.2.1'
         PACKAGE_RELEASE: '1'
     - name: 'Generate artifact properties'
       id: 'generate-artifact-properties'


gordonwang0:
Don't forget to also:
- Update https://github.com/Azure/iotedge/blob/a5436419f90b0918f5b1303fb7bc30b687eace10/edgelet/contrib/centos/aziot-edge.spec#L22 to 1.2.1
- Update https://github.com/Azure/iotedge/blob/a5436419f90b0918f5b1303fb7bc30b687eace10/edgelet/contrib/debian/control#L12 to 1.2.1
- Bump the cargo dependency in iotedge/edgelet

gordonwang0 (approved):

lfitchett:
> Don't forget to also:
> 
> * Update https://github.com/Azure/iotedge/blob/a5436419f90b0918f5b1303fb7bc30b687eace10/edgelet/contrib/centos/aziot-edge.spec#L22 to 1.2.1
> * Update https://github.com/Azure/iotedge/blob/a5436419f90b0918f5b1303fb7bc30b687eace10/edgelet/contrib/debian/control#L12 to 1.2.1
> * Bump the cargo dependency in iotedge/edgelet

Ok, I'll change that in my pr that fixes the privileged flag. Release/1.2 already has the version correct

PR #247, opened 2021-06-29
Update packages versions to 1.3.0-dev

gordonwang0:


arsing:
Why? We can't update master for every new release from release/1.2. If anything, this should've been set to 1.3.0 when release/1.2 was branched, so might as well change it to that. And also the `PACKAGE_VERSION` in `.github/workflows/packages.yaml`

varunpuranik:
> Why? We can't update master for every new release from release/1.2. If anything, this should've been set to 1.3.0 when release/1.2 was branched, so might as well change it to that. And also the `PACKAGE_VERSION` in `.github/workflows/packages.yaml`

+1 to that.. master should be version 1.3.0-dev

.github/workflows/packages.yaml
```diff
@@ -45,8 +45,8 @@ jobs:
           '/src/ci/package.sh'
       env:
         ARCH: "${{ matrix.arch }}"
-        PACKAGE_VERSION: '1.2.1'
-        PACKAGE_RELEASE: '1'
+        PACKAGE_VERSION: '1.3.0'
+        PACKAGE_RELEASE: 'dev'
```
arsing, line 49: https://github.com/Azure/iot-identity-service/commit/997ab230b78c6951bbe28b45f3ff08445666a1ee :)

(Maybe add a YAML comment so that we don't forget in the future.)

arsing (changes requested):

.github/workflows/packages.yaml
gordonwang0: Fixed

gordonwang0: Added comment.

arsing (approved):

varunpuranik (approved):

PR #249, opened 2021-06-30
Allow shutting down http_common server

gordonwang0:
Passes in a shutdown oneshot to `http_common::Incoming:serve`

Posting a message to the corresponding `oneshot::Sender` will cause the server to gracefully shut down by awaiting all unfinished server tasks.

This is currently not used by any of the aziotd services, but edged will use it for when it has to shut down and restart its servers for reprovisioning.

### DIFF
--- a/http-common/Cargo.toml
+++ b/http-common/Cargo.toml
@@ -30,3 +30,4 @@ serde_json = "1"
 
 [features]
 tokio1 = ["headers", "hyper", "hyper-openssl", "hyper-proxy", "openssl", "openssl-sys", "tokio"]
+serve_with_shutdown = ["tokio/time"]
--- a/http-common/Cargo.toml
+++ b/http-common/Cargo.toml
@@ -21,7 +21,7 @@ openssl-sys = { version = "0.9", optional = true }
 percent-encoding = "2"
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
-tokio = { version = "1", features = ["net", "rt-multi-thread", "sync"], optional = true }
+tokio = { version = "1", features = ["net", "rt-multi-thread", "sync", "time"], optional = true }
 tracing = { version = "0.1", features = ["log"] }
 url = { version = "2", features = ["serde"] }
 
@@ -30,4 +30,3 @@ serde_json = "1"
 
 [features]
 tokio1 = ["headers", "hyper", "hyper-openssl", "hyper-proxy", "openssl", "openssl-sys", "tokio"]
-serve_with_shutdown = ["tokio/time"]
--- a/http-common/src/connector.rs
+++ b/http-common/src/connector.rs
@@ -49,8 +49,6 @@ impl Incoming {
             + 'static,
         <H as hyper::service::Service<hyper::Request<hyper::Body>>>::Future: Send,
     {
-        const MAX_REQUESTS_PER_USER: usize = 10;
-
         match self {
             Incoming::Tcp { listener } => loop {
                 let (tcp_stream, _) = listener.accept().await?;
@@ -58,12 +56,7 @@ impl Incoming {
                 // TCP is available in test builds only (not production). Assume current user is root.
                 let server = crate::uid::UidService::new(None, 0, server.clone());
                 tokio::spawn(async move {
-                    if let Err(http_err) = hyper::server::conn::Http::new()
-                        .serve_connection(tcp_stream, server)
-                        .await
-                    {
-                        log::info!("Error while serving HTTP connection: {}", http_err);
-                    }
+                    serve_tcp(server, tcp_stream).await;
                 });
             },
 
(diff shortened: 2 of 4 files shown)


http-common/src/connector.rs
```diff
@@ -72,36 +68,174 @@ impl Incoming {
...
         }
     }
+
+    #[cfg(feature = "serve_with_shutdown")]
+    pub async fn serve_with_shutdown<H>(
+        &mut self,
+        server: H,
+        shutdown: tokio::sync::oneshot::Receiver<()>,
+    ) -> std::io::Result<()>
+    where
```
gordonwang0, line 87: @arsing Is there a way to collapse these `where` clauses so I don't have to attach it to every function that takes H? Just curious, but I didn't find any.

http-common/src/connector.rs
```diff
@@ -72,36 +68,174 @@ impl Incoming {
...
+                        shutdown_loop = shutdown;
+                    }
+                };
+            },
+        }
+
+        // Wait for all running server tasks to finish before returning.
+        let poll_ms = std::time::Duration::from_millis(100);
+
+        while tasks.load(std::sync::atomic::Ordering::Relaxed) != 0 {
```
gordonwang0, line 161: @arsing Is there a better way to wait for an atomic to have a certain value? Though the spin-wait isn't too bad.

arsing: You could replace the counter with a monotonic counter + a map of current counter value -> `JoinHandle` returned by `tokio::spawn`. The counter value is the key that allows you to remove the `JoinHandle`. Then this loop is just `futures_util::future::join_all(map.into_iter().map(|(_, join_handle)| join_handle))`

On the other hand, not only isn't sleeping loop simpler, it also has the advantage that you can log "Waiting for {} more requests to complete..." before you sleep, so I think I prefer that.

http-common/src/connector.rs
```diff
@@ -72,36 +68,174 @@ impl Incoming {
...
+                match future::select(shutdown_loop, accept).await {
+                    future::Either::Left((_, _)) => break,
+                    future::Either::Right((unix_stream, shutdown)) => {
+                        let unix_stream = unix_stream?.0;
+
+                        let (user_state, ucred) = get_user_auth(&unix_stream, user_state)?;
+                        let server =
+                            crate::uid::UidService::new(ucred.pid(), ucred.uid(), server.clone());
+
+                        tasks.fetch_add(1, std::sync::atomic::Ordering::Relaxed);
```
arsing, line 145: I don't think `Relaxed` is appropriate for any of the three. The two `fetch_`s should be `AcqRel` and the `load` should be `Acquire`. `Relaxed` allows them to move around, so for example the two `fetch_`s would not necessarily happen before and after the `serve_unix` call.

http-common/Cargo.toml
```diff
@@ -30,3 +30,4 @@ serde_json = "1"
 
 [features]
 tokio1 = ["headers", "hyper", "hyper-openssl", "hyper-proxy", "openssl", "openssl-sys", "tokio"]
+serve_with_shutdown = ["tokio/time"]
```
arsing, line 33: Why do we need it to be optional? Might as well convert `aziotd` to use it too? (In which case, keep the old function name `serve`.)

http-common/src/connector.rs
gordonwang0: I considered keeping the handles, but you're right the counter was so much simpler. I'll keep as-is and add that log message.

arsing (changes requested):

http-common/src/connector.rs
arsing: Until we get trait aliases (`trait Foo = Bar + Baz`, RFC opened in 2017 so it'll probably get stabilized by 3017 AD), the only way is to make a new trait that inherits the bounds and provide a blanket impl for existing types: `trait Foo: Bar + Baz { } impl<T> Foo for T where T: Bar + Baz { }` and then use `: Foo` as the bound.

http-common/src/connector.rs
gordonwang0: Fixed.

http-common/Cargo.toml
gordonwang0: Do we need graceful shutdown in aziotd? AFAIK, we don't handle CTRL+C here or have any APIs that would cause any service to restart.

arsing: Well, it would simplify this http-common code. aziotd's main can just keep the `Sender` around and not use it. Just make sure not to drop it, otherwise the `Receiver` will resolve with `Err(_)`

gordonwang0: Ok, makes sense. Fixed.

http-common/src/connector.rs
```diff
@@ -51,56 +57,108 @@ impl Incoming {
...
+                futures_util::pin_mut!(accept);
+
+                // Await either the next established connection or the shutdown signal.
+                match future::select(shutdown_loop, accept).await {
+                    future::Either::Left((_, _)) => break,
+                    future::Either::Right((unix_stream, shutdown)) => {
+                        let unix_stream = unix_stream?.0;
+
+                        let ucred = unix_stream.peer_cred()?;
+                        let user_state = user_state
```
arsing, line 111: Since you've "changed" this code anyway (by indenting it), can I convince you to get rid of this semaphore and replace it with another `AtomicUsize` ?

gordonwang0: Ok, did that.

http-common/src/connector.rs
```diff
@@ -108,41 +110,49 @@ impl Incoming {
...
+                            let available = servers_available
+                                .fetch_update(
+                                    atomic::Ordering::AcqRel,
+                                    atomic::Ordering::Acquire,
+                                    |current| {
+                                        if current == 0 {
+                                            None
+                                        } else {
+                                            Some(current - 1)
+                                        }
```
arsing, line 134: `current.checked_sub(1)`

arsing (changes requested):

http-common/src/connector.rs
gordonwang0: fixed

arsing (approved):

arsing:
(I updated the PR title and description to match the new implementation, for the commit message's sake.)

PR #250, opened 2021-07-07
Log to see what the URI is that is failing

nyanzebra:


arsing:
You should be getting that with debug / trace logs already (from hyper)?

PR #252, opened 2021-07-22
Remove identityd "get trust bundle" API.

arsing:
This was an undocumented and unimplemented stub; it always returned no certs.

Since it was undocumented and unimplemented, we don't consider this
a breaking change that necessitates a new API version.


varunpuranik (changes requested):
I would like to check with Shiproom once if this is okay.

huguesBouvier (approved):
LGTM. One failed test to check, real or flaky.

gordonwang0 (approved):
I wasn't aware that this was a stub and was using it in the edged tokio refactor. But LGTM.

arsing:
@varunpuranik

>I would like to check with Shiproom once if this is okay.

Well, the alternative is we leave it as a stub, then implement the IS-scoped trust bundle you and I talked about in the call, and reuse this stub for that.

---

@gordonwang0

It would be good to avoid functional changes as part of your refactor.

For this specific case, yes edged has its own config for the trust bundle's cert ID (set by `iotedge config apply`) which it uses to get the cert from certd directly.

arsing:
Not doing this right now.

PR #253, opened 2021-08-02
Add support for non-JSON response types in http_common

gordonwang0:
Adds `mod response`, which provides various response types for the HTTP server. Previously, the server only supported JSON response types, but others (zip, chunked) will be needed as part of the edged refactor.

### DIFF
--- a/http-common/src/server.rs
+++ b/http-common/src/server.rs
@@ -128,19 +128,17 @@ macro_rules! make_service {
                                                 }
                                             };
 
-                                            let (status_code, response) = match <$route as http_common::server::Route>::delete(route, body).await {
+                                            match <$route as http_common::server::Route>::delete(route, body).await {
                                                 Ok(result) => result,
                                                 Err(err) => return Ok(err.to_http_response()),
-                                            };
-                                            http_common::server::json_response(status_code, response.as_ref())
+                                            }
                                         },
 
                                         http::Method::GET => {
-                                            let (status_code, response) = match <$route as http_common::server::Route>::get(route).await {
+                                            match <$route as http_common::server::Route>::get(route).await {
                                                 Ok(result) => result,
                                                 Err(err) => return Ok(err.to_http_response()),
-                                            };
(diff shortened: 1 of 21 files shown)


http-common/src/server.rs
```diff
@@ -340,31 +326,54 @@ impl Error {
...
+
+    pub fn zip(
+        status_code: hyper::StatusCode,
+        size: usize,
+        body: Vec<u8>,
+    ) -> hyper::Response<hyper::Body> {
+        let res = hyper::Response::builder().status(status_code);
+
+        let res = res
+            .header(hyper::header::CONTENT_ENCODING, "zip")
```
gordonwang0, line 369: This header is probably wrong, but it's what the current edged uses. So I left it.

arsing: Yes, it's wrong. It should be `deflate`.

It should be fine to change it - it was already a bad value so no self-respecting client library would be depending on it anyway.

arsing (changes requested):

http-common/src/server.rs
gordonwang0: Ok, fixed it.

http-common/src/server.rs
```diff
@@ -340,31 +326,54 @@ impl Error {
...
-    body: Option<&impl serde::Serialize>,
-) -> hyper::Response<hyper::Body> {
-    let res = hyper::Response::builder().status(status_code);
-    // `res` is consumed by both branches, so this cannot be replaced with `Option::map_or_else`
-    //
-    // Ref: https://github.com/rust-lang/rust-clippy/issues/5822
-    #[allow(clippy::option_if_let_else)]
-    let res = if let Some(body) = body {
+pub mod response {
+    pub fn empty() -> hyper::Response<hyper::Body> {
```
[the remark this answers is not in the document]
gordonwang0: Fixed.

arsing (approved):

PR #254, opened 2021-08-03
Add chunked server response

gordonwang0:
Adds support for chunked text server responses. This is used by edged's module logs API

### DIFF
--- a/http-common/src/server.rs
+++ b/http-common/src/server.rs
@@ -342,6 +342,22 @@ pub mod response {
         res
     }
 
+    pub fn chunked<S, O, E>(status_code: hyper::StatusCode, body: S) -> hyper::Response<hyper::Body>
+    where
+        S: futures_util::stream::Stream<Item = Result<O, E>> + Send + 'static,
+        O: Into<hyper::body::Bytes> + 'static,
+        E: Into<Box<dyn std::error::Error + Send + Sync>> + 'static,
+    {
+        let body = hyper::Body::wrap_stream(body);
+
+        let res = hyper::Response::builder()
+            .status(status_code)
+            .body(body);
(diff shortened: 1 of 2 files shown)


http-common/src/server.rs
```diff
@@ -342,6 +342,20 @@ pub mod response {
...
 
+    pub fn chunked<S, O, E>(status_code: hyper::StatusCode, body: S) -> hyper::Response<hyper::Body>
+    where
+        S: futures_util::stream::Stream<Item = Result<O, E>> + Send + 'static,
+        O: Into<hyper::body::Bytes> + 'static,
+        E: Into<Box<dyn std::error::Error + Send + Sync>> + 'static,
+    {
+        let body = hyper::Body::wrap_stream(body);
+
+        let res = hyper::Response::builder().status(status_code).body(body);
```
arsing, line 353: Should also set a `content-type`

gordonwang0: Done.

http-common/src/server.rs
```diff
@@ -342,6 +342,23 @@ pub mod response {
         res
     }
 
+    pub fn chunked<S, O, E>(status_code: hyper::StatusCode, body: S) -> hyper::Response<hyper::Body>
```
arsing, line 345: Perhaps either rename thing to `chunked_text` to make it clear it's specific to `text/plain`, or take the `&'static str` of the content type as a parameter. Either approach is fine; I'll sign off to not block.

arsing (approved):

http-common/src/server.rs
gordonwang0: Ok, added the parameter.

PR #256, opened 2021-08-18
Add header support to request helper function

lfitchett:


http-common/src/request.rs
```diff
@@ -1,16 +1,35 @@
...
 
-pub async fn request<TRequest, TResponse>(
+use std::str::FromStr;
+
+pub async fn request<TUri, THeader, TRequest, TResponse>(
     client: &hyper::Client<super::Connector, hyper::Body>,
     method: http::Method,
-    uri: &str,
+    uri: TUri,
+    headers: Option<&[(THeader, THeader)]>,
```
gordonwang0, line 9: Do the header types need to be generic? I can't think of a scenario where headers would need to be anything other than String/str.

lfitchett: Is there a reason not to? I originally did it generic because I was calling it with a header enum that impl'd the AsRef<&str>. I'm not now, but I didn't see a reason to make it more specific since its a helper function

gordonwang0 (approved):

PR #258, opened 2021-08-20
Using unix sockets in priority

huguesBouvier:
Workload manager creates on unix socket per module.
The current code will try to use a systemd socket for each one.

### DIFF
--- a/http-common/src/connector.rs
+++ b/http-common/src/connector.rs
@@ -279,7 +279,21 @@ impl Connector {
                 fd_to_listener(fd)
             }
 
-            // Prefer use of systemd sockets.
+            // Prefer use of unix sockets.
+            (_, Connector::Unix { socket_path }) => {
+                match std::fs::remove_file(&*socket_path) {
+                    Ok(()) => (),
+                    Err(err) if err.kind() == std::io::ErrorKind::NotFound => (),
+                    Err(err) => return Err(err),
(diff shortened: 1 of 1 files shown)


arsing:
Preferring systemd sockets is how it's supposed to be. Why are you making this change?

huguesBouvier:
> Preferring systemd sockets is how it's supposed to be. Why are you making this change?

This is to for the workload manager. Worload manager is creating unix socket, so with the current code, when it creates a socket for a new module, it tries to match it with systemd and we gets the error message:"multiple systemd sockets found".

huguesBouvier:
> > Preferring systemd sockets is how it's supposed to be. Why are you making this change?
> 
> This is to for the workload manager. Worload manager is creating unix socket, so with the current code, when it creates a socket for a new module, it tries to match it with systemd and we gets the error message:"multiple systemd sockets found".

This is now used by edged which requires more than one socket

arsing:
Sure. But edged still uses systemd sockets. and the four services in this repo still use systemd sockets.

huguesBouvier:
> Sure. But edged still uses systemd sockets. and the four services in this repo still use systemd sockets.

Right now it doesn't work for unix socket. It tries to use systemd because it is there in priority. Let's talk about that on teams instead, it will be faster than posting.

PR #259, opened 2021-08-21
Support for multiple socket names in Connector

huguesBouvier:
Support for edged in connector.

### DIFF
--- a/http-common/src/connector.rs
+++ b/http-common/src/connector.rs
@@ -261,48 +261,17 @@ impl Connector {
     }
 
     #[cfg(feature = "tokio1")]
-    pub async fn incoming(self) -> std::io::Result<Incoming> {
+    pub async fn incoming(self, socket_name: Option<String>) -> std::io::Result<Incoming> {
         // Check for systemd sockets.
-        let systemd_sockets = get_num_systemd_sockets()
+        let systemd_socket = get_systemd_sockets(socket_name)
             .map_err(|err| std::io::Error::new(std::io::ErrorKind::Other, err))?;
+        //Can return
 
-        match (systemd_sockets, self) {
-            (Some(sockets), Connector::Fd { fd }) => {
-                // Check that the fd specified in the connector URI is in the valid range of systemd sockets.
-                if fd < SD_LISTEN_FDS_START || fd >= SD_LISTEN_FDS_START + sockets {
-                    return Err(std::io::Error::new(
-                        std::io::ErrorKind::Other,
-                        format!("fd {} not valid", fd),
-                    ));
-                }
-
-                fd_to_listener(fd)
-            }
+        match (systemd_socket, self) {
+            (_, Connector::Fd { fd }) => fd_to_listener(fd),
 
             // Prefer use of systemd sockets.
-            (Some(sockets), _) => {
-                // If more than 1 systemd socket is found, we don't know which one to use.
-                if sockets > 1 {
-                    return Err(std::io::Error::new(
-                        std::io::ErrorKind::Other,
-                        "multiple systemd sockets found",
-                    ));
-                }
-
-                fd_to_listener(SD_LISTEN_FDS_START)
-            }
-
-            (None, Connector::Tcp { host, port }) => {
-                if cfg!(debug_assertions) {
-                    let listener = tokio::net::TcpListener::bind((&*host, port)).await?;
-                    Ok(Incoming::Tcp { listener })
-                } else {
-                    Err(std::io::Error::new(
-                        std::io::ErrorKind::Other,
-                        "servers can only use `unix://` connectors, not `http://` connectors",
-                    ))
-                }
-            }
+            (Some(fd), _) => fd_to_listener(fd),
 
             (None, Connector::Unix { socket_path }) => {
                 match std::fs::remove_file(&*socket_path) {
@@ -318,10 +287,17 @@ impl Connector {
                 })
             }
 
-            (None, Connector::Fd { fd: _ }) => Err(std::io::Error::new(
-                std::io::ErrorKind::Other,
-                "fd URI specified, but systemd sockets not found",
-            )),
+            (None, Connector::Tcp { host, port }) => {
+                if cfg!(debug_assertions) {
+                    let listener = tokio::net::TcpListener::bind((&*host, port)).await?;
+                    Ok(Incoming::Tcp { listener })
+                } else {
+                    Err(std::io::Error::new(
+                        std::io::ErrorKind::Other,
+                        "servers can only use `unix://` connectors, not `http://` connectors",
+                    ))
+                }
+            }
         }
     }
 
@@ -696,8 +672,10 @@ fn socket_name_to_fd(name: &str) -> Result<std::os::unix::io::RawFd, String> {
             None => return Err(format!("socket {} not found", name)),
         };
 
+    let fd = index + SD_LISTEN_FDS_START - 1;
+
     // The index in LISTEN_FDNAMES is an offset from SD_LISTEN_FDS_START.
-    Ok(index + SD_LISTEN_FDS_START)
+    Ok(fd)
 }
 
 #[cfg(feature = "tokio1")]
@@ -724,7 +702,9 @@ fn fd_to_listener(fd: std::os::unix::io::RawFd) -> std::io::Result<Incoming> {
 ///
 /// This mimics `sd_listen_fds` from libsystemd, then returns the number of systemd socket fds.
 #[cfg(feature = "tokio1")]
-fn get_num_systemd_sockets() -> Result<Option<std::os::unix::io::RawFd>, String> {
+fn get_systemd_sockets(
+    socket_name: Option<String>,
+) -> Result<Option<std::os::unix::io::RawFd>, String> {
     // Ref: <https://www.freedesktop.org/software/systemd/man/sd_listen_fds.html>
     //
     // >sd_listen_fds parses the number passed in the $LISTEN_FDS environment variable, then sets the FD_CLOEXEC flag
@@ -742,6 +722,13 @@ fn get_num_systemd_sockets() -> Result<Option<std::os::unix::io::RawFd>, String>
     // both /run/aziot and /var/lib/aziot to connect to a service, or the services continue to bind sockets under /run/aziot but have to create
     // /run/aziot themselves on startup with ACLs for all three users and all three groups.
 
+    //We consider 4 cases:
+    // 1. When there is only 1 socket. In this case, we can ignore the socket name. It means,
+    // the call is made by identity service which uses only one socket. So matching is simple
+    // 2. There are 2 sockets and a socket name is provided. It means edged is telling use to match this socket.
+    // 3. There are 2 sockets and a socket name is provided but no LISTEN_FDNAMES. We can't match.
+    // 4. There are 2 sockets but not socket name is provided. In there case it means there is no corresponding systemd socket we should match
+
     let listen_fds: std::os::unix::io::RawFd = match get_env("LISTEN_FDS")? {
         Some(listen_fds) => listen_fds
             .parse()
@@ -750,9 +737,21 @@ fn get_num_systemd_sockets() -> Result<Option<std::os::unix::io::RawFd>, String>
         None => return Ok(None),
     };
 
+    // If there is  no socket available, not match is possible.
     if listen_fds == 0 {
         return Ok(None);
     }
+
+    //If there is only one socket, we know this is the identity service which uses only one socket, so we have a match:
+    if listen_fds == 1 {
+        return Ok(Some(SD_LISTEN_FDS_START as i32));
+    }
+
+    //if there is more than 1 and we don't have a socket name to match, this is edged telling us that there is no systemd socket we can match.
+    let socket_name = match socket_name {
+        Some(socket_name) => socket_name,
+        None => return Ok(None),
+    };
 
     // fcntl(CLOEXEC) all the fds so that they aren't inherited by the child processes.
     // Note that we want to do this for all the fds, not just the one we're looking for.
@@ -768,8 +767,37 @@ fn get_num_systemd_sockets() -> Result<Option<std::os::unix::io::RawFd>, String>
         }
     }
 
-    // Return the number of systemd sockets.
-    Ok(Some(listen_fds))
+    // If there is more than one socket, this is edged. We can attempt to match the socket name to systemd.
+    // This happens when a unix Uri is provided in the config.toml. Systemd sockets get created noneless so we still prefer to use them.
+    // If no socket name is provided, this mean edged is telling us there is not matching systemd socket, we return early. See above "((listen_fds > 1)&&(socket_name == None))"
+    // If a socket name is provided but we don't see the env variable LISTEN_FDNAMES, it means we are probably on an older OS, we can't match either.
+    let listen_fdnames = match get_env("LISTEN_FDNAMES")? {
+        Some(listen_fdnames) => listen_fdnames,
+        None => return Ok(None),
+    };
+    let listen_fdnames: Vec<&str> = listen_fdnames.split(':').collect();
+
+    if listen_fds != listen_fdnames.len() as i32 {
+        return Err(format!(
+            "Mismatch, there are {} fds, and {} names",
+            listen_fds,
+            listen_fdnames.len()
+        ));
+    }
+
+    let mut fd_index = SD_LISTEN_FDS_START;
+    for name in listen_fdnames.into_iter() {
+        if name.eq(&socket_name) {
+            return Ok(Some(fd_index));
+        }
+
+        fd_index += 1;
+    }
+
+    Err(format!(
+        "Could not find a match for {} in the fd list",
+        socket_name
+    ))
 }
 
 #[cfg(test)]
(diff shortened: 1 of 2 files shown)


http-common/src/connector.rs
```diff
@@ -724,7 +702,9 @@ fn fd_to_listener(fd: std::os::unix::io::RawFd) -> std::io::Result<Incoming> {
 ///
 /// This mimics `sd_listen_fds` from libsystemd, then returns the number of systemd socket fds.
 #[cfg(feature = "tokio1")]
-fn get_num_systemd_sockets() -> Result<Option<std::os::unix::io::RawFd>, String> {
+fn get_systemd_sockets(
```
arsing, line 705: `get_systemd_socket` (singular)

http-common/src/connector.rs
```diff
@@ -750,10 +737,22 @@ fn get_num_systemd_sockets() -> Result<Option<std::os::unix::io::RawFd>, String>
...
+
+    //if there is more than 1 and we don't have a socket name to match, this is edged telling us that there is no systemd socket we can match.
+    let socket_name = match socket_name {
+        Some(socket_name) => socket_name,
+        None => return Ok(None),
+    };
+
     // fcntl(CLOEXEC) all the fds so that they aren't inherited by the child processes.
     // Note that we want to do this for all the fds, not just the one we're looking for.
     for fd in SD_LISTEN_FDS_START..(SD_LISTEN_FDS_START + listen_fds) {
```
arsing, line 758: This needs to be done in any situation where there's an fd. So this needs to be done before the early return in the `listen_fds == 1` case. Just move it up before your line 745.

http-common/src/connector.rs
```diff
@@ -768,8 +767,37 @@ fn get_num_systemd_sockets() -> Result<Option<std::os::unix::io::RawFd>, String>
...
+
+    if listen_fds != listen_fdnames.len() as i32 {
+        return Err(format!(
+            "Mismatch, there are {} fds, and {} names",
+            listen_fds,
+            listen_fdnames.len()
+        ));
+    }
+
+    let mut fd_index = SD_LISTEN_FDS_START;
```
arsing, line 788: This can be just:

```rust
if let Some(index) = listen_fdnames.iter().position(|fdname| fdname == socket_name) {
    Ok(SD_LISTEN_FDS_START + index as _)
}
else {
    Err(format!("could not find a match for...")
}
```

(The `as _` might make clippy complain about `cast_sign_loss`. In that case just `#[allow]` it on this line. You can search for `cast_sign_loss` in this repo for examples.)

arsing (changes requested):

http-common/src/connector.rs
huguesBouvier: You're right, I fixed both.

http-common/src/connector.rs
huguesBouvier: Good catch. I dragged and drop some code and forgot about it.

http-common/src/connector.rs
huguesBouvier: It looks better, thanks a lot

arsing (approved):
Please get signoff from Gordon too.

huguesBouvier:
> Please get signoff from Gordon too.

Ok I will wait his input

http-common/src/connector.rs
```diff
@@ -720,23 +696,32 @@ fn fd_to_listener(fd: std::os::unix::io::RawFd) -> std::io::Result<Incoming> {
     }
 }
 
-/// Finds the number of available systemd sockets. Checks if this process has been socket-activated.
+/// Return a matching systemd sockets. Checks if this process has been socket-activated.
```
gordonwang0, line 699:
```suggestion
/// Return a matching systemd socket. Checks if this process has been socket-activated.
```

http-common/src/connector.rs
```diff
@@ -720,23 +696,32 @@ fn fd_to_listener(fd: std::os::unix::io::RawFd) -> std::io::Result<Incoming> {
...
-/// This mimics `sd_listen_fds` from libsystemd, then returns the number of systemd socket fds.
+/// This mimics `sd_listen_fds` from libsystemd, then returns the fd of systemd socket.
 #[cfg(feature = "tokio1")]
-fn get_num_systemd_sockets() -> Result<Option<std::os::unix::io::RawFd>, String> {
+fn get_systemd_socket(
+    socket_name: Option<String>,
+) -> Result<Option<std::os::unix::io::RawFd>, String> {
     // Ref: <https://www.freedesktop.org/software/systemd/man/sd_listen_fds.html>
     //
+    // Trie to find a systemd socket to match when non "fd" path has been provided.
```
gordonwang0, line 708:
```suggestion
    // Try to find a systemd socket to match when non "fd" path has been provided.
```

http-common/src/connector.rs
```diff
@@ -720,23 +696,32 @@ fn fd_to_listener(fd: std::os::unix::io::RawFd) -> std::io::Result<Incoming> {
...
+    // We consider 4 cases:
+    // 1. When there is only 1 socket. In this case, we can ignore the socket name. It means,
```
gordonwang0, line 710:
```suggestion
    // 1. When there is only 1 socket. In this case, we can ignore the socket name. It means
```

http-common/src/connector.rs
```diff
@@ -720,23 +696,32 @@ fn fd_to_listener(fd: std::os::unix::io::RawFd) -> std::io::Result<Incoming> {
...
+    // the call is made by identity service which uses only one systemd socket. So matching is simple
+    // 2. There are > 1 systemd sockets and a socket name is provided. It means edged is telling use to match an fd with the provided socket name.
```
gordonwang0, line 712:
```suggestion
    // 2. There are > 1 systemd sockets and a socket name is provided. It means edged is telling us to match an fd with the provided socket name.
```

http-common/src/connector.rs
```diff
@@ -720,23 +696,32 @@ fn fd_to_listener(fd: std::os::unix::io::RawFd) -> std::io::Result<Incoming> {
...
+    // 3. There are > 1 systemd sockets and a socket name is provided but no LISTEN_FDNAMES. We can't match.
+    // 4. There are > 1 systemd sockets but not socket name is provided. In there case it means there is no corresponding systemd socket we should match
```
gordonwang0, line 714:
```suggestion
    // 4. There are > 1 systemd sockets but no socket name is provided. In this case it means there is no corresponding systemd socket we should match
```

http-common/src/connector.rs
```diff
@@ -720,23 +696,32 @@ fn fd_to_listener(fd: std::os::unix::io::RawFd) -> std::io::Result<Incoming> {
...
+    //
     // >sd_listen_fds parses the number passed in the $LISTEN_FDS environment variable, then sets the FD_CLOEXEC flag
     // >for the parsed number of file descriptors starting from SD_LISTEN_FDS_START. Finally, it returns the parsed number.
     //
     // Note that it's not possible to distinguish between fd numbers if a process requires more than one socket.
+    // That is why in edged case we use the systemd socket name to know which fd the function should return
```
gordonwang0, line 720:
```suggestion
    // That is why in edged's case we use the systemd socket name to know which fd the function should return
```

http-common/src/connector.rs
```diff
@@ -750,6 +735,7 @@ fn get_num_systemd_sockets() -> Result<Option<std::os::unix::io::RawFd>, String>
         None => return Ok(None),
     };
 
+    // If there is  no socket available, not match is possible.
```
gordonwang0, line 738:
```suggestion
    // If there is no socket available, no match is possible.
```

http-common/src/connector.rs
```diff
@@ -768,8 +754,56 @@ fn get_num_systemd_sockets() -> Result<Option<std::os::unix::io::RawFd>, String>
         }
     }
 
-    // Return the number of systemd sockets.
-    Ok(Some(listen_fds))
+    //If there is only one socket, we know this is the identity service which uses only one socket, so we have a match:
```
gordonwang0, line 757:
```suggestion
    // If there is only one socket, we know this is the identity service which uses only one socket, so we have a match:
```

http-common/src/connector.rs
```diff
@@ -768,8 +754,56 @@ fn get_num_systemd_sockets() -> Result<Option<std::os::unix::io::RawFd>, String>
...
+    if listen_fds == 1 {
+        return Ok(Some(SD_LISTEN_FDS_START as i32));
```
gordonwang0, line 759: Do you need the `as i32` here? If so, it would be better to have it `as std::os::unix::io::RawFd` to match the return type of this function.

http-common/src/connector.rs
```diff
@@ -768,8 +754,56 @@ fn get_num_systemd_sockets() -> Result<Option<std::os::unix::io::RawFd>, String>
...
+    }
+
+    //if there is more than 1 and we don't have a socket name to match, this is edged telling us that there is no systemd socket we can match.
```
gordonwang0, line 762:
```suggestion
    // If there is more than 1 socket and we don't have a socket name to match, this is edged telling us that there is no systemd socket we can match.
```

http-common/src/connector.rs
```diff
@@ -768,8 +754,56 @@ fn get_num_systemd_sockets() -> Result<Option<std::os::unix::io::RawFd>, String>
...
+    let socket_name = match socket_name {
+        Some(socket_name) => socket_name,
+        None => return Ok(None),
+    };
+
+    // If there is more than one socket, this is edged. We can attempt to match the socket name to systemd.
+    // This happens when a unix Uri is provided in the config.toml. Systemd sockets get created noneless so we still prefer to use them.
```
gordonwang0, line 769:
```suggestion
    // This happens when a unix Uri is provided in the config.toml. Systemd sockets get created nonetheless, so we still prefer to use them.
```

http-common/src/connector.rs
```diff
@@ -768,8 +754,56 @@ fn get_num_systemd_sockets() -> Result<Option<std::os::unix::io::RawFd>, String>
...
+    // If no socket name is provided, this mean edged is telling us there is not matching systemd socket, we return early. See above "((listen_fds > 1)&&(socket_name == None))"
```
gordonwang0, line 770: I'm not sure what this comment is referring to. I don't see the code mentioned.

http-common/src/connector.rs
```diff
@@ -768,8 +754,56 @@ fn get_num_systemd_sockets() -> Result<Option<std::os::unix::io::RawFd>, String>
...
+    // If a socket name is provided but we don't see the env variable LISTEN_FDNAMES, it means we are probably on an older OS, we can't match either.
```
gordonwang0, line 771:
```suggestion
    // If a socket name is provided but we don't see the env variable LISTEN_FDNAMES, it means we are probably on an older OS, and we can't match either.
```

http-common/src/connector.rs
```diff
@@ -768,8 +754,56 @@ fn get_num_systemd_sockets() -> Result<Option<std::os::unix::io::RawFd>, String>
...
+        None => return Ok(None),
+    };
+    let listen_fdnames: Vec<&str> = listen_fdnames.split(':').collect();
+
+    #[allow(
+        clippy::cast_possible_truncation,
+        clippy::cast_sign_loss,
+        clippy::cast_possible_wrap
+    )]
+    if listen_fds != listen_fdnames.len() as i32 {
```
gordonwang0, line 783: Would be better to use `try_into()` instead of a cast and remove the clippy allows.

http-common/src/connector.rs
```diff
@@ -768,8 +754,56 @@ fn get_num_systemd_sockets() -> Result<Option<std::os::unix::io::RawFd>, String>
...
+    if let Some(index) = listen_fdnames
+        .iter()
+        .position(|fdname| (*fdname).eq(&socket_name))
+    {
+        #[allow(
+            clippy::cast_possible_truncation,
+            clippy::cast_sign_loss,
+            clippy::cast_possible_wrap
+        )]
+        Ok(Some(SD_LISTEN_FDS_START + index as i32))
```
gordonwang0, line 800: `try_into()` instead of cast.

http-common/src/connector.rs
```diff
@@ -261,48 +261,15 @@ impl Connector {
     }
 
     #[cfg(feature = "tokio1")]
-    pub async fn incoming(self) -> std::io::Result<Incoming> {
+    pub async fn incoming(self, socket_name: Option<String>) -> std::io::Result<Incoming> {
         // Check for systemd sockets.
-        let systemd_sockets = get_num_systemd_sockets()
+        let systemd_socket = get_systemd_socket(socket_name)
             .map_err(|err| std::io::Error::new(std::io::ErrorKind::Other, err))?;
+        //Can return
```
gordonwang0, line 268:
```suggestion

```

http-common/src/connector.rs
```diff
@@ -696,8 +670,10 @@ fn socket_name_to_fd(name: &str) -> Result<std::os::unix::io::RawFd, String> {
             None => return Err(format!("socket {} not found", name)),
         };
 
+    let fd = index + SD_LISTEN_FDS_START - 1;
+
     // The index in LISTEN_FDNAMES is an offset from SD_LISTEN_FDS_START.
```
gordonwang0, line 675: Move comment to above 673.

gordonwang0 (changes requested):

http-common/src/connector.rs
huguesBouvier: You are right

http-common/src/connector.rs
huguesBouvier: Good catch, I updated the code and forgot to remove that comment.

http-common/src/connector.rs
huguesBouvier: hum, it is not happy. But I will use "std::os::unix::io::RawFd" similar to above.

huguesBouvier: found a way

http-common/src/connector.rs
```diff
@@ -768,8 +753,59 @@ fn get_num_systemd_sockets() -> Result<Option<std::os::unix::io::RawFd>, String>
...
+    // If there is more than one socket, this is edged. We can attempt to match the socket name to systemd.
+    // This happens when a unix Uri is provided in the config.toml. Systemd sockets get created nonetheless, so we still prefer to use them.
+    // If a socket name is provided but we don't see the env variable LISTEN_FDNAMES, it means we are probably on an older OS, and we can't match either.
+    let listen_fdnames = match get_env("LISTEN_FDNAMES")? {
+        Some(listen_fdnames) => listen_fdnames,
+        None => return Ok(None),
+    };
+    let listen_fdnames: Vec<&str> = listen_fdnames.split(':').collect();
+
+    #[allow(
```
gordonwang0, line 776: You can remove these allows now.

http-common/src/connector.rs
```diff
@@ -768,8 +753,59 @@ fn get_num_systemd_sockets() -> Result<Option<std::os::unix::io::RawFd>, String>
...
+            listen_fds,
+            listen_fdnames.len()
+        ));
+    }
+
+    if let Some(index) = listen_fdnames
+        .iter()
+        .position(|fdname| (*fdname).eq(&socket_name))
+    {
+        #[allow(
```
gordonwang0, line 795: You can remove these allows now.

http-common/src/connector.rs
```diff
@@ -768,8 +753,59 @@ fn get_num_systemd_sockets() -> Result<Option<std::os::unix::io::RawFd>, String>
...
+    };
+    let listen_fdnames: Vec<&str> = listen_fdnames.split(':').collect();
+
+    #[allow(
+        clippy::cast_possible_truncation,
+        clippy::cast_sign_loss,
+        clippy::cast_possible_wrap
+    )]
+    let len: std::os::unix::io::RawFd = std::convert::TryInto::try_into(listen_fdnames.len())
+        .expect("usize -> std::os::unix::io::RawFd");
```
gordonwang0, line 782: `map_err(|_| "invalid number of sockets".to_string())?` instead of `expect`.

http-common/src/connector.rs
```diff
@@ -768,8 +753,59 @@ fn get_num_systemd_sockets() -> Result<Option<std::os::unix::io::RawFd>, String>
...
+        .iter()
+        .position(|fdname| (*fdname).eq(&socket_name))
+    {
+        #[allow(
+            clippy::cast_possible_truncation,
+            clippy::cast_sign_loss,
+            clippy::cast_possible_wrap
+        )]
+        let index: std::os::unix::io::RawFd =
+            std::convert::TryInto::try_into(index).expect("usize -> std::os::unix::io::RawFd");
```
gordonwang0, line 801: `map_err(|_| "invalid number of sockets".to_string())?` instead of `expect`.

gordonwang0 (changes requested):

http-common/src/connector.rs
huguesBouvier: Done

http-common/src/connector.rs
huguesBouvier: Ok, I was just copying code I found elsewhere:
let digest_len =
                        std::convert::TryInto::try_into(digest.len()).expect("usize -> c_ulong");

http-common/src/connector.rs
huguesBouvier: Ok, I was just copying code I found elsewhere:
let digest_len =
                        std::convert::TryInto::try_into(digest.len()).expect("usize -> c_ulong");

gordonwang0 (approved):

PR #260, opened 2021-08-24
Add headers option to request_no_content

lfitchett:
This pr adds a headers function for request_no_content. It also slightly changes the behavior or request_no_content, allowing responses other than 204 to return Ok. This is because sometimes when docker-rs calls the no_content function, the http response has content, we just don't care about the content and want to drop it. 

This pr also pulls the shared code of both request functions in to shared functions to reduce copied code and make them easier to read.

### DIFF
--- a/http-common/src/lib.rs
+++ b/http-common/src/lib.rs
@@ -29,7 +29,9 @@ pub use proxy::{get_proxy_uri, MaybeProxyConnector};
 #[cfg(feature = "tokio1")]
 mod request;
 #[cfg(feature = "tokio1")]
-pub use request::{request, request_no_content, request_with_headers};
+pub use request::{
+    request, request_no_content, request_with_headers, request_with_headers_no_content,
+};
 
 pub mod server;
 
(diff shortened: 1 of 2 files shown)


gordonwang0 (approved):

arsing:
>This is because sometimes when docker-rs calls the no_content function, the http response has content, we just don't care about the content and want to drop it.

Why not make docker-rs use `request` then? It can ignore the response body by deserializing it as a `serde::de::IgnoredAny`

lfitchett:
> > This is because sometimes when docker-rs calls the no_content function, the http response has content, we just don't care about the content and want to drop it.
> 
> Why not make docker-rs use `request` then? It can ignore the response body by deserializing it as a `serde::de::IgnoredAny`

The main difference between he 2 functions (besides deserializing the output) is that request always checks for ContentType: application/json while request_no_content only does so when it has an error it expects to be able to parse.

gordonwang0 (approved):

PR #268, opened 2021-08-31
Update to Rust 1.54

gordonwang0:


arsing:
Also revert https://github.com/Azure/iot-identity-service/commit/5a7bab601056eb3333b76d81b599df622946ce73 since https://github.com/rust-lang/rust/issues/82006 was fixed in 1.52.0

arsing (approved):

gordonwang0:
@arsing Regarding the test failures: How do I silence the compiler warning for the generated test code?

arsing:
To fix the test build failure, add this after [this list:](https://github.com/Azure/iot-identity-service/blob/5e0ec2561cffbea78da0cef196c19017653c6eed/key/aziot-keyd/src/keys.rs#L1169-L1176)

```rust
// The tests generated by bindgen deref nullptr to test for size and alignment of the structs and fields.
// While this is indeed UB, this is only done for tests, so allow it.
//
// Ref: https://github.com/rust-lang/rust-bindgen/issues/1651
#![cfg_attr(test, allow(deref_nullptr))]
```

arsing:
BTW, we want to do 1.54 specifically rather than just `stable` ?

gordonwang0:
Edged is using 1.54 in its rust-toolchain, so I figured we should keep them in sync.

PR #272, opened 2021-09-02
Don't require content-type: application/json in server

gordonwang0:
While we *should* require JSON requests to specify the correct content type, this requirement wasn't present in pre-tokio-refactor edged. Certain versions of edgeHub don't set the content-type header in their request and crash when this server rejects edgeHub's request.

### DIFF
--- a/http-common/src/server.rs
+++ b/http-common/src/server.rs
@@ -182,33 +182,38 @@ macro_rules! make_service {
 
                                         http::Method::PUT => {
                                             let content_type = headers.get(hyper::header::CONTENT_TYPE).and_then(|value| value.to_str().ok());
-                                            if content_type.as_deref() != Some("application/json") && content_type.as_deref() != None {
-                                                return Ok((http_common::server::Error {
-                                                    status_code: http::StatusCode::UNSUPPORTED_MEDIA_TYPE,
-                                                    message: "request body must be application/json".into(),
-                                                }).to_http_response());
-                                            }
+                                            let body = match content_type.as_deref() {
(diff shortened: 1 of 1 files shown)


gordonwang0:
@arsing Let me know if you think there's a better way. I'm making this change to retain compatibility with certain builds of edgeHub that don't send content-type in their request.

arsing:
Keep the check. Just add a `.unwrap_or("json")` so that a missing header causes the body is treated as JSON too.

ie

```diff,rust
-if content_type.as_deref() == Some("application/json") {
+match content_type.as_deref() { Some("application/json") | None => {
```

That way clients that do set a non-JSON content type will get an error response.

gordonwang0:
Ok, fixed.

http-common/src/server.rs
```diff
@@ -182,7 +182,7 @@ macro_rules! make_service {
 
                                         http::Method::PUT => {
                                             let content_type = headers.get(hyper::header::CONTENT_TYPE).and_then(|value| value.to_str().ok());
-                                            if content_type.as_deref() != Some("application/json") {
+                                            if content_type.as_deref() != Some("application/json") && content_type.as_deref() != None {
```
arsing, line 185: Better to make this be a `match` too instead of comparing twice. Same as the other methods, except the `_ =>` arm has `return Ok(Error { ... })` instead of `None`

gordonwang0: Fixed.

arsing (approved):

PR #273, opened 2021-09-04
Expand build targets to Debian 11 and RHEL 8

micahl:
Adding Debian11 and RHEL8-compatible options to the possible build targets.

Removed reference to the the db-systemd package since it doesn't appear to exist on Bullseye.  Per Debian [bug 822670](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=822670) it was merged into debhelper.  It persisted in Debian 10 to provide a transition window

Validated changes to the e2e-tests.sh script by manually triggering it for 'manual-symmetric-key' on both Debian 11 and RHEL 8 (AlmaLinux).  For the AlmaLinux VM I had to first accept the terms of the VM before subsequent test runs would work (i.e. `az vm image terms accept --urn 'almalinux:almalinux:8_4-gen2:8.4.20210729'`)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -267,12 +267,10 @@ rpm:
 
 	# Copy spec file to rpmbuild specs directory
 	mkdir -p $(RPMBUILDDIR)/SPECS
-	sed -e 's/@version@/$(PACKAGE_VERSION)/g; s/@release@/$(PACKAGE_RELEASE)/g' contrib/centos/aziot-identity-service.spec >$(RPMBUILDDIR)/SPECS/aziot-identity-service.spec
-
-	which openssl # Assert that openssl exists and the empty output of `openssl version -e` is not because it never even ran
-	ENGINESDIR="$$(openssl version -e | sed 's/^ENGINESDIR: "\(.*\)"$$/\1/' | sed 's/\//\\\//g')"; \
-	ENGINESDIR="$${ENGINESDIR:-$(libdir)/openssl/engines}"; \
-	sed -ie "s/@enginesdir@/$$ENGINESDIR/g" $(RPMBUILDDIR)/SPECS/aziot-identity-service.spec
+	ENGINESDIR='/usr/lib64/openssl/engines'; \
+	ENGINESDIR=$$(if [ ! -d $$ENGINESDIR ]; then echo $$(openssl version -e | sed 's/^ENGINESDIR: "\(.*\)"$$/\1/'); else echo $$ENGINESDIR; fi); \
+    ENGINESDIR="$$(echo $$ENGINESDIR | sed -e 's/\//\\\//g')"; \
+	sed -e "s/@version@/$(PACKAGE_VERSION)/g; s/@release@/$(PACKAGE_RELEASE)/g; s/@enginesdir@/$$ENGINESDIR/g" contrib/centos/aziot-identity-service.spec >$(RPMBUILDDIR)/SPECS/aziot-identity-service.spec
 
 	# Copy preset file to be included in the package
 	cp contrib/centos/00-aziot.preset $(RPMBUILDDIR)/SOURCES
--- a/Makefile
+++ b/Makefile
@@ -269,6 +269,11 @@ rpm:
 	mkdir -p $(RPMBUILDDIR)/SPECS
 	sed -e 's/@version@/$(PACKAGE_VERSION)/g; s/@release@/$(PACKAGE_RELEASE)/g' contrib/centos/aziot-identity-service.spec >$(RPMBUILDDIR)/SPECS/aziot-identity-service.spec
 
+	which openssl # Assert that openssl exists and the empty output of `openssl version -e` is not because it never even ran
+	ENGINESDIR="$$(openssl version -e | sed 's/^ENGINESDIR: "\(.*\)"$$/\1/' | sed 's/\//\\\//g')"; \
+	ENGINESDIR="$${ENGINESDIR:-$(libdir)/openssl/engines}"; \
+	sed -ie "s/@enginesdir@/$$ENGINESDIR/g" $(RPMBUILDDIR)/SPECS/aziot-identity-service.spec
+
 	# Copy preset file to be included in the package
 	cp contrib/centos/00-aziot.preset $(RPMBUILDDIR)/SOURCES
 
--- a/Makefile
+++ b/Makefile
@@ -269,7 +269,7 @@ rpm:
 	mkdir -p $(RPMBUILDDIR)/SPECS
 	ENGINESDIR='/usr/lib64/openssl/engines'; \
 	ENGINESDIR=$$(if [ ! -d $$ENGINESDIR ]; then echo $$(openssl version -e | sed 's/^ENGINESDIR: "\(.*\)"$$/\1/'); else echo $$ENGINESDIR; fi); \
-    ENGINESDIR="$$(echo $$ENGINESDIR | sed -e 's/\//\\\//g')"; \
+	ENGINESDIR="$$(echo $$ENGINESDIR | sed -e 's/\//\\\//g')"; \
 	sed -e "s/@version@/$(PACKAGE_VERSION)/g; s/@release@/$(PACKAGE_RELEASE)/g; s/@enginesdir@/$$ENGINESDIR/g" contrib/centos/aziot-identity-service.spec >$(RPMBUILDDIR)/SPECS/aziot-identity-service.spec
 
 	# Copy preset file to be included in the package
--- a/Makefile
+++ b/Makefile
@@ -267,10 +267,12 @@ rpm:
 
 	# Copy spec file to rpmbuild specs directory
 	mkdir -p $(RPMBUILDDIR)/SPECS
-	ENGINESDIR='/usr/lib64/openssl/engines'; \
-	ENGINESDIR=$$(if [ ! -d $$ENGINESDIR ]; then echo $$(openssl version -e | sed 's/^ENGINESDIR: "\(.*\)"$$/\1/'); else echo $$ENGINESDIR; fi); \
-	ENGINESDIR="$$(echo $$ENGINESDIR | sed -e 's/\//\\\//g')"; \
-	sed -e "s/@version@/$(PACKAGE_VERSION)/g; s/@release@/$(PACKAGE_RELEASE)/g; s/@enginesdir@/$$ENGINESDIR/g" contrib/centos/aziot-identity-service.spec >$(RPMBUILDDIR)/SPECS/aziot-identity-service.spec
+
+	# Ask openssl for its engine dir and on versions <1.1 fallback to a hardcoded path.
+	which openssl # Assert that openssl exists and that the empty output of `openssl version -e` is not because it never even ran
+	ENGINESDIR="$$(openssl version -e 2> /dev/null | sed 's/^ENGINESDIR: "\(.*\)"$$/\1/')"; \
+	ENGINESDIR="$${ENGINESDIR:-%\{_libdir\}/openssl/engines}"; \
+	sed -e "s/@version@/$(PACKAGE_VERSION)/g; s/@release@/$(PACKAGE_RELEASE)/g; s|@enginesdir@|$$ENGINESDIR|g" contrib/centos/aziot-identity-service.spec >$(RPMBUILDDIR)/SPECS/aziot-identity-service.spec
 
 	# Copy preset file to be included in the package
 	cp contrib/centos/00-aziot.preset $(RPMBUILDDIR)/SOURCES
--- a/ci/e2e-tests.sh
+++ b/ci/e2e-tests.sh
@@ -60,10 +60,6 @@ get_package() {
             artifact_name='centos-7'
             ;;
 
-        'platform:el8')
-            artifact_name='el-8'
-            ;;
-
         'debian:9')
             artifact_name='debian-9-slim'
             ;;
@@ -74,6 +70,10 @@ get_package() {
 
         'debian:11')
             artifact_name='debian-11-slim'
+            ;;
+
+        'platform:el8')
+            artifact_name='el-8'
             ;;
 
         'ubuntu:18.04')
@@ -121,12 +121,7 @@ get_package() {
             printf '%s/%s\n' "$PWD" aziot-identity-service-*.x86_64.rpm
             ;;
 
-        'platform:el8')
-            unzip -j package.zip 'el8/amd64/aziot-identity-service-*.x86_64.rpm' -x '*-debuginfo-*.rpm' '*-devel-*.rpm' >&2
-            printf '%s/%s\n' "$PWD" aziot-identity-service-*.x86_64.rpm
-            ;;
-
-         'debian:9')
+        'debian:9')
             unzip -j package.zip 'debian9/amd64/aziot-identity-service_*_amd64.deb' >&2
             printf '%s/%s\n' "$PWD" aziot-identity-service_*_amd64.deb
             ;;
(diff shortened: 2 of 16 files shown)


ci/e2e-tests.sh
```diff
@@ -113,7 +121,12 @@ get_package() {
             printf '%s/%s\n' "$PWD" aziot-identity-service-*.x86_64.rpm
             ;;
 
-        'debian:9')
+        'platform:el8')
+            unzip -j package.zip 'el8/amd64/aziot-identity-service-*.x86_64.rpm' -x '*-debuginfo-*.rpm' '*-devel-*.rpm' >&2
+            printf '%s/%s\n' "$PWD" aziot-identity-service-*.x86_64.rpm
+            ;;
+
+         'debian:9')
```
arsing, line 129: Indent is wrong. 9 spaces should be 8.

ci/e2e-tests.sh
```diff
@@ -113,7 +121,12 @@ get_package() {
             printf '%s/%s\n' "$PWD" aziot-identity-service-*.x86_64.rpm
             ;;
 
-        'debian:9')
+        'platform:el8')
```
arsing, line 124: Sort these arms alphabetically, in this and every other file.

ci/e2e-tests.sh
```diff
@@ -218,7 +236,14 @@ case "$OS" in
...
 
-    'debian:9')
+    'platform:el8')
+        # az vm image list --all \
+        #     --publisher 'almalinux' --offer 'almalinux' --sku '8_4-gen2' \
+        #     --query "[?publisher == 'almalinux' && offer == 'almalinux'].{ sku: sku, version: version }" --output table
+        vm_image='almalinux:almalinux:8_4-gen2:8.4.20210729'
+        ;;
+
+     'debian:9')
```
arsing, line 246: Indent is wrong. 5 spaces should be 4.

ci/install-build-deps.sh
```diff
@@ -5,26 +5,24 @@
 # WARNING: This script is destructive to your machine's environment and globally-installed files. For example, the Ubuntu-specific parts of the script
 # modify the contents of /etc/apt. The script is intended to be run inside a container of the corresponding OS, not directly on your machine.
 
-
-OS="$(. /etc/os-release; echo "$ID:$VERSION_ID")"
-
+OS="$(. /etc/os-release; if [ -z ${PLATFORM_ID+x} ]; then echo "$ID:$VERSION_ID"; else echo "$PLATFORM_ID"; fi)"
```
arsing, line 8:
```sh
OS="$(. /etc/os-release; echo "${$PLATFORM_ID:-$ID:$VERSION_ID}")"
```

ci/package.sh
```diff
@@ -11,32 +11,42 @@ mkdir -p packages
 
 
 case "$OS" in
-    'centos:7')
+    'centos:7'|'platform:el8')
         case "$ARCH" in
             'arm32v7'|'aarch64')
-                echo 'Cross-compilation on CentOS 7 is not supported' >&2
+                echo "Cross-compilation on $OS is not supported" >&2
```
arsing, line 17: It's not supported on EL8 ?

contrib/debian/control
```diff
@@ -2,8 +2,7 @@ Source: aziot-identity-service
 Section: admin
 Priority: optional
 Maintainer: Azure IoT Edge Devs
-Build-Depends: debhelper (>=9),
-               dh-systemd
+Build-Depends: debhelper (>=9.20160709)
```
arsing, line 5: Why that specific version?

docs-dev/packages.md
```diff
@@ -16,14 +16,22 @@ This repository contains three services - `aziot-certd`, `aziot-identityd` and `
 <td><code>centos:7</code></td>
 </tr>
 <tr>
-<td>Debian 9 / Raspbian 9</td>
+<td>RHEL 8 compatible</td>
+<td><code>almalinux:8.4</code></td>
```
arsing, line 20: Is there a reason to not use `centos:8` ?

arsing (changes requested):

arsing:
>Removed reference to the the db-systemd package since it doesn't appear to exist on Bullseye. Per Debian [bug 822670](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=822670) it was merged into debhelper. It persisted in Debian 10 to provide a transition window

Yes, this looks okay. The deb file has postinst etc filled out just as it did before with `dh_systemd`.

ci/package.sh
micahl: I haven't done an exhaustive search, but didn't readily see packages for EL8 to set up cross-compile.  I saw this [toolchain from ARM](https://developer.arm.com/tools-and-software/open-source-software/developer-tools/gnu-toolchain/gnu-a/downloads), but am punting on deeper investigation due to personal time constraints.

contrib/debian/control
micahl: It's the one mentioned in the bug I linked in the description.

docs-dev/packages.md
micahl: Only b/c CentOS 8 hits EOL in a few months so I figured it would be better to use something expected to have a longer lifetime.

ci/install-build-deps.sh
micahl: This?  `OS="$(. /etc/os-release; echo "${PLATFORM_ID:=$ID:$VERSION_ID}")"`

I was getting an error otherwise.

arsing: Without that `$` at the front, yes. With `=` instead of `-`, no. Fixed my comment.

ci/package.sh
arsing: If you haven't investigated, then https://github.com/Azure/iot-identity-service/pull/273/files#diff-bff6c5a7caec352e17657fc581f0a2c86581122c44e2501d2bb8278d8c5a9c98R133 needs to be changed.

docs-dev/packages.md
arsing: Hmm, I thought they'd update the tag for Stream, but it looks like they publish a separate `stream8` tag and that too only on quay.io ( https://wiki.centos.org/FAQ/CentOSStream#What_artifacts_are_built.3F -> https://quay.io/repository/centos/centos?tab=tags vs https://hub.docker.com/_/centos?tab=tags&page=1&ordering=last_updated)

Also, I guess for the periods of time where Stream is ahead of RHEL we could have a situation where we link to newer libs / symbols than RHEL and so our packages become uninstallable / unrunnable.

This is okay then.

Do you have any reason to pick AlmaLinux over RockyLinux? (IIRC we're even sponsoring that one.)

ci/package.sh
micahl: Reverted to the prior language with an explanation of the current state of investigation for RHEL 8.

ci/e2e-tests.sh
micahl: Done

ci/e2e-tests.sh
micahl: Fixed

ci/e2e-tests.sh
micahl: Fixed

ci/install-build-deps.sh
micahl: Fixed

ci/package.sh
micahl: Ok, I've reverted to the prior language with another line describing the current state of investigation with RHEL.

docs-dev/packages.md
micahl: > Do you have any reason to pick AlmaLinux over RockyLinux? (IIRC we're even sponsoring that one.)

No. But, now that I've learned of [RedHat's UBI images](https://developers.redhat.com/products/rhel/ubi) I'm recommending one of those instead (at least for the container to use in building). 

Note this PR still suggests using an AlmaLinux VM for the tests - only b/c it doesn't incur any additional cost on top of Azure infrastructure costs.  There isn't an equivalent "free" one for Rocky Linux.

Fwiw, I believe Microsoft is doing some sponsoring of both projects.

gauravIoTEdge (commented):
Would you need to update the pipeline file? 

Add distros here: https://github.com/Azure/iot-identity-service/blob/main/.github/workflows/packages.yaml#L15   ?

and possibly here: https://github.com/Azure/iot-identity-service/blob/main/ci/package.sh#L44 ?

micahl:
> Would you need to update the pipeline file?
> 
> Add distros here: https://github.com/Azure/iot-identity-service/blob/main/.github/workflows/packages.yaml#L15 ?
> 
> and possibly here: https://github.com/Azure/iot-identity-service/blob/main/ci/package.sh#L44 ?

Yeah, I have that change already.  I didn't include it in this PR and don't remember why.  Maybe wasn't sure whether it would be better to have it in a subsequent PR.  I'll just add it.

gauravIoTEdge:
> > Would you need to update the pipeline file?
> > Add distros here: https://github.com/Azure/iot-identity-service/blob/main/.github/workflows/packages.yaml#L15 ?
> > and possibly here: https://github.com/Azure/iot-identity-service/blob/main/ci/package.sh#L44 ?
> 
> Yeah, I have that change already. I didn't include it in this PR and don't remember why. Maybe wasn't sure whether it would be better to have it in a subsequent PR. I'll just add it.

And I see you do have the other changes. 

Sorry for the confusion; I'm not used to the GitHub layout. I just missed the fact that you do have them.

micahl:
Added changes to the workflow files.  Although... it looks like there isn't a check for `aziot-key-openssl-engine-shared` for either Debian 11 or RHEL 8.  Should there have been?  @arsing could probably spot what I missed faster than I.

arsing:
>test / basic (redhat/ub8:latest, amd64) (pull_request) Failing after 24s — basic (redhat/ub8:latest, amd64) 

`redhat/ub8:latest` should be `redhat/ubi8:latest` ?

>Although... it looks like there isn't a check for `aziot-key-openssl-engine-shared` for either Debian 11 or RHEL 8.

The other ones are stuck in "Waiting for status to be reported" so Github is probably having a wibble. Fix the container name so that the checks re-run and we can see.

arsing:
Yeah, the missing openssl engine tests are there now. So workflow changes are fine.

Test failure is legitimate. The new `$OS` calculation in `install-build-deps.sh` needs to be done in `install-runtime-deps.sh` too.

micahl:
> Test failure is legitimate. The new `$OS` calculation in `install-build-deps.sh` needs to be done in `install-runtime-deps.sh` too.

Thanks, I'll add that later tonight

micahl:
@arsing RHEL appears to have put speed bumps in the way of adding the EPEL repo without a subscription on their UBI images. It didn't encounter this on Rocky or Alma Linux. I tried to add a targeted check in the `install-runtime-deps.sh` to only use the workaround if it's RHEL.

arsing:
I saw. You prefer that vs going back to almalinux? (I don't have an opinion either way.)

micahl:
@arsing Preference is to align as closely as possible to the "official" bits, but I have to admit the limitations are rather annoying.  Any more and I'd probably ditch it and use something more straightforward.  Looks like for EL8 platforms the `/usr/lib64/openssl` folder is no more and we'd need to use the ENGINESDIR similar to Debian.  New spec file needed for the EL8 rpm?  Looks like a larger change than I initially thought.

@gauravIoTEdge can you pick up from here on this PR?  I probably wouldn't have time to continue it in the near term.

arsing:
>Looks like for EL8 platforms the `/usr/lib64/openssl` folder is no more and we'd need to use the ENGINESDIR similar to Debian.

Indeed.

```sh
$ openssl version -e

ENGINESDIR: "/usr/lib64/engines-1.1"
```

---

>New spec file needed for the EL8 rpm?

We can make the enginesdir a template variable and have `make rpm` fill it out based on the output of `openssl version -e`, just like it does for `@version@` and `@release@` https://github.com/Azure/iot-identity-service/blob/f6294bdd027284a27a03d83b4c5dc736d98e0a32/contrib/centos/aziot-identity-service.spec#L6-L7 https://github.com/Azure/iot-identity-service/blob/f6294bdd027284a27a03d83b4c5dc736d98e0a32/Makefile#L270

Something like:

```diff,spec
-- <openssl engines directory>/openssl/engines/libaziot_keys.so - An openssl engine that can be used to work with asymmetric keys managed by the Azure IoT Keys Service.
+- @enginesdir@/libaziot_keys.so - An openssl engine that can be used to work with asymmetric keys managed by the Azure IoT Keys Service.

[...]

-%{_libdir}/openssl/engines/libaziot_keys.so
+@enginesdir@/libaziot_keys.so
```

```diff,makefile
     sed -e 's/@version@/$(PACKAGE_VERSION)/g; s/@release@/$(PACKAGE_RELEASE)/g' contrib/centos/aziot-identity-service.spec >$(RPMBUILDDIR)/SPECS/aziot-identity-service.spec
+
+    which openssl # Assert that openssl exists and the empty output of `openssl version -e` is not because it never even ran
+    ENGINESDIR="$$(openssl version -e | sed 's/^ENGINESDIR: "\(.*\)"$/\1/')"; \
+    ENGINESDIR='$${ENGINESDIR:-$(libdir)/openssl/engines}"; \
+    sed -ie 's/@enginesdir@/$$ENGINESDIR/g' $(RPMBUILDDIR)/SPECS/aziot-identity-service.spec
```

It won't work if we have to support arm32 or aarch64 so we'll have to revisit it when / if we care about them.

Makefile
```diff
@@ -267,7 +267,10 @@ rpm:
 
 	# Copy spec file to rpmbuild specs directory
 	mkdir -p $(RPMBUILDDIR)/SPECS
-	sed -e 's/@version@/$(PACKAGE_VERSION)/g; s/@release@/$(PACKAGE_RELEASE)/g' contrib/centos/aziot-identity-service.spec >$(RPMBUILDDIR)/SPECS/aziot-identity-service.spec
+	ENGINESDIR='/usr/lib64/openssl/engines'; \
+	ENGINESDIR=$$(if [ ! -d $$ENGINESDIR ]; then echo $$(openssl version -e | sed 's/^ENGINESDIR: "\(.*\)"$$/\1/'); else echo $$ENGINESDIR; fi); \
```
arsing, line 271: Running `openssl` should have higher preference than checking the directory, as in the snippet I'd given, because ultimately what `openssl` says is the enginesdir is what matters.

Makefile
```diff
@@ -267,7 +267,10 @@ rpm:
...
+	ENGINESDIR="$$(echo $$ENGINESDIR | sed -e 's/\//\\\//g')"; \
```
arsing, line 272: Instead of doing this, change the sed pattern to use `|` instead of `/`.

arsing (changes requested):

Makefile
micahl: `openssl` on CentOS 7 (v1.02) doesn't have a `version -e` option to respond with the enginesdir.

arsing: Of course. That's why we had to hard-code the directory in the first place, and why the snippet I gave specifically accounts for that by checking if the output is empty.

Makefile
micahl: Nice.  Didn't know it was possible to switch the delimiter so easily.

Makefile
micahl: In your snippet was the `which openssl` needed for something?

arsing: Yes. The comment on the same line explains its purpose.

arsing (approved):
Looks good to me, @gordonwang0 or someone should review too.

gordonwang0 (approved):

PR #274, opened 2021-09-08
Add Mariner RPM to Build Pipeline

josephknierman:
Added to allow build pipeline to make RPM files for mariner.


.github/workflows/packages.yaml
```diff
@@ -65,3 +69,33 @@ jobs:
       with:
         name: "${{ steps.generate-artifact-properties.outputs.artifact-name }}"
         path: 'packages'
+  BuildMarinerRelease:
```
gordonwang0, line 72: This job should be added to the OS/arch matrix above. Does that not work?

Makefile
```diff
@@ -406,3 +406,14 @@ install-rpm: install-common
 
 	# devel
 	$(INSTALL_DATA) -D key/aziot-keys/aziot-keys.h $(DESTDIR)$(includedir)/aziot-identity-service/aziot-keys.h
+
+install-rpm-mariner: install-common
```
arsing, line 410: The only difference from `install-rpm` seems to be not installing the unit preset, so the specfile should just use `install-rpm` and then delete the `install`ed preset.

ci/package-mariner.sh
```diff
@@ -0,0 +1,100 @@
+#!/bin/bash
```
arsing, line 1: This should be merged into `package.sh` instead of being a separate file.

.github/workflows/packages.yaml
```diff
@@ -65,3 +69,33 @@ jobs:
...
+    - name: 'Generate artifact properties'
+      id: 'generate-artifact-properties'
+      run: |
+        echo "::set-output name=artifact-name::packages_mariner_amd64"
+    - name: 'Upload'
+      uses: 'actions/upload-artifact@v1'
+      with:
+        name: "${{ steps.generate-artifact-properties.outputs.artifact-name }}"
+        path: 'packages'
+
```
arsing, line 101: Remove this trailing whitespace.

contrib/mariner/aziot-identity-service.spec
```diff
@@ -0,0 +1,227 @@
...
+    docdir=%{_docdir}/%{name} \
+    includedir=%{_includedir} \
+    libdir=%{_libdir} \
+    libexecdir=%{_libexecdir} \
+    localstatedir=%{_localstatedir} \
+    sysconfdir=%{_sysconfdir} \
+    unitdir=%{_unitdir} \
+    OPENSSL_ENGINES_DIR=%{_enginesdir} \
+    RELEASE=1 \
+    V=1 \
```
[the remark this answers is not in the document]
arsing: They're for "verbose" and "non-debug", not for package versions.

ci/package-mariner.sh
```diff
@@ -0,0 +1,100 @@
...
+MarinerSourceDir=$MarinerSpecsDir/SOURCES
+
+# Extract built toolkit in building direectory
+mkdir -p $MarinerRPMBUILDDIR
+pushd $MarinerRPMBUILDDIR
+mv /src/CBL-Mariner/toolkit.tar.gz toolkit.tar.gz
+tar -xzvf toolkit.tar.gz
+popd
+
+# move tarballed IIS source to building directory
```
arsing, line 41: s/IIS/iot-identity-service/

arsing (changes requested):

.github/workflows/packages.yaml
josephknierman: This doesn't work mainly because we are not building the mariner version of this on a mariner system. we need to build the Mainer version on Ubuntu-18.04 docker image, and the current matirx above did not support different builds from the container os. If we want I think i can added a build-for variable to each entry in the matirx, but these would have to be hard coded and would not be a generated combination like its current state.

arsing: You don't need to add a new variable for every matrix entry, just for mariner with [`include`.](https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#example-including-additional-values-into-combinations) Then map that variable to the `OS` env var, and change the scripts that fill in `$OS` based on `/etc/os-release` to only do it if `OS` isn't already set.

Makefile
josephknierman: The main issue with that is that 00-aziot.preset is not a file that is copied over for Mariner at install time. this will result in the make command failing. If you think it is still worth it we can add 00-aziot.preset to be copied over and use the same install-rpm make call.

.github/workflows/packages.yaml
josephknierman: changed to use the same matrix section and removed duplicate task

ci/package-mariner.sh
josephknierman: merged into package.sh

ci/package-mariner.sh
josephknierman: I am not quite sure what you wanted changed here. Is there something at the path s/IIS/iot-identity-service/

.github/workflows/packages.yaml
josephknierman: trailing space should be removed now

ci/package-mariner.sh
arsing: I'm asking you to replace "IIS" with "iot-identity-service"

josephknierman: changed IIS to iot-identity-service

Makefile
josephknierman: Removed install-rpm-mariner. there is now a flag that will determine if the preset gets installed or not

.github/workflows/packages.yaml
arsing: You've also deleted the newline at the end that shouldn't have been deleted. (Notice the diff in the GitHub UI.)

Makefile
arsing: You don't need a flag. Just `rm` after you call `make install` in the new specfile.

josephknierman: The problem is getting the preset file to the correct location because it will fail on finding it without any change. We would have to copy it to wherever ../../Sources is on the mariner tooIkit before calling make, personally feel like that is not an ideal solution but will implement that as well.

arsing: Right, you would have to include the preset in the source too. It's true that it's odd to do that. I'm okay with the flag.

ci/package.sh
```diff
@@ -4,7 +4,7 @@ set -euo pipefail
 
 cd /src
 
-. ./ci/install-build-deps.sh
+. ./ci/install-build-deps.sh $OS
```
arsing, line 7: You don't need to pass it in. If it's defined in the environment (by the caller), use it. Otherwise derive it from `/etc/os-release`.

ci/package.sh
```diff
@@ -106,6 +106,101 @@ case "$OS" in
...
+        popd
+
+        # move tarballed iot-identity-service source to building directory
+        mkdir -p $MarinerSourceDir
+        mv /tmp/aziot-identity-service-$PACKAGE_VERSION.tar.gz $MarinerSourceDir/aziot-identity-service-$PACKAGE_VERSION.tar.gz
+
+        mkdir /temp
+        rustup install stable
+        rustup default stable
+        wget https://github.com/eqrion/cbindgen/archive/refs/tags/v0.18.0.tar.gz -O /temp/cbindgen-0.18.0.tar.gz
```
arsing, line 149: `install-build-deps.sh` uses `^0.15`, so it'll install the latest 0.15, not 0.18. Same for bindgen below, ^0.54 vs 0.57

It's fine to update to the latest deps but then you need to update `install-build-deps.sh` too.

ci/package.sh
josephknierman: removed passing $OS

.github/workflows/packages.yaml
josephknierman: fixed line ending

.github/workflows/packages.yaml
```diff
@@ -46,8 +53,10 @@ jobs:
         docker run --rm \
           -v "$GITHUB_WORKSPACE:/src" \
           -e "ARCH=$ARCH" \
+          -e "OS=$OS" \
           -e "PACKAGE_VERSION=$PACKAGE_VERSION" \
           -e "PACKAGE_RELEASE=$PACKAGE_RELEASE" \
+          --privileged \
```
arsing, line 59: Why does it need `--privileged` ?

josephknierman:
Looks like the CentOS and redhat builds are failing now. likely related to changes in the make file. tracking this down now

arsing (changes requested):

.github/workflows/packages.yaml
josephknierman: Its required for the mariner toolkit to build here. have tried it without privileged an cannot get that to work in this environment.

arsing: Do you know which command requires privilege and why? It's not a blocker. I'm just curious.

josephknierman: I believe it is when we make the RPM file so it would be the "make build-packages" command in packages.sh. it has been a couple months since testing that specifically so that might not be right.

ci/package.sh
josephknierman: all references to 0.18 should not be gone and replace with 0.15

arsing: You haven't changed `install-build-deps.sh`

Also you haven't changed bindgen.

josephknierman: I might be wrong here but i dont think the bingen requirements from intsall-deps.sh will effect us here. Part of the mariner toolkit is that it works fully offline when building RPMS so it will only have access to the bindgen and cbindgen packages that we provided in the form of tarballs before kicking off the RPM build process. These are being downloaded in packages.sh. and then extracted by the spec file in the prep stage.

arsing: The point is to use the same versions of dependencies regardless of whether it's a Mariner build or not. It has nothing to do with Mariner build being hermetic or not.

In fact, for the same reason, the `rustup install stable` and `rustup default stable` commands above this need to use the version defined in the `rust-toolchain` file.

josephknierman: Ok i think i understand. so would we just need to have condition that checks if the build is for mariner and then instead of installing "^0.15" it would get exactly "0.15.0". and then the same thing for bingen.

arsing: No, because that would still install different versions of dependencies, eg 0.15.0 on Mariner and 0.15.2 on non-Mariner.

You'd need to install 0.15.0 always. ie change `^0.15` to `=0.15.0` for everything.

josephknierman:
Just checked and the identify service binaries are still working on a mariner env after all the PR changes.

ci/package.sh
josephknierman: should be resolved now only using exact versions

.github/workflows/packages.yaml
arsing: I see a bunch of chrooting going on, so it's probably that.

arsing:
Please take the latest commit in https://github.com/arsing/iot-identity-service/commits/add-mariner-rpm . I put what I changed in the description. It worked for me up to the point where it failed because of the issue fixed by https://github.com/Azure/iot-identity-service/pull/314 , but if you see anything that I broke we can talk about it.

josephknierman:
@arsing just checked the commit here https://github.com/arsing/iot-identity-service/commits/add-mariner-rpm. I did notice that it was failing the Mariner build here https://github.com/arsing/iot-identity-service/runs/4184371732?check_suite_focus=true, and I confirmed that this was failing on a local builds as well. The fix for this is to move the RPM build directory for the Mariner toolkit our of /tmp. Will add those changes to this PR shortly after verifying the resulting mariner RPM.

arsing:
I thought it might be that, but it doesn't fail for me. Do you know why it fails for you and for the GH Actions VM? /tmp is entirely within the container so I don't see why it would be different.

josephknierman:
Not entirely sure why i am seeing the failure in both GH workflows and on my local machine when building the RPM in /tmp. just found that moving it back to /src resulted in it working locally, will see if it works for GH actions soon https://github.com/josephknierman/iot-identity-service/actions/runs/1454525969.  Just sent a question to the Mariner folks to see if there is some limitation on not being able to build packages in /tmp.

arsing:
Looks like it worked, so yes I guess we can move it back to /src. It'd be nice to have the source dir clean but it's not a big deal.

arsing (approved):
#314 should be in in a few hours, and then the bot can rebase this on top automatically.

josephknierman:
Just merged changes from release/1.2 into this branch to pull in fixes for redhat package build.

yophilav (approved):

gordonwang0 (approved):

arsing:
@josephknierman As we had discussed before, please also cherry-pick to `main` and send a PR.

PR #275, opened 2021-09-14
Temporarily increase MAX_REQUESTS_PER_USER to 30

nlcamp:


arsing:
Why?

nlcamp:
> Why?

Investigating iotedge connectivity test errors where aziot-edged repeatedly reports  "Max simultaneous connections reached for user..." and edgeAgent repeatedly reports "Unable to process module [XYZ] add or update as the module identity could not be obtained".

This is not intended to be merged. I'm just using this to generate packages that can be used in an iotedge connectivity test pipeline to further investigate these errors.

PR #278, opened 2021-09-14
CN fallback for X.509 registration (#212)

gordonwang0:
In 1.1, iotedge had a fallback mechanism for DPS X.509 attestation. In 1.2, however, new options for obtaining device identity certificate became available. In the new world, CS could be configured to issue identity certificate in one of these configurations:

1. receive a CSR with any subject name. This subject would be IS's config.toml registration_id value.
2. receive a CSR and common name is configured within CS config.toml. This will override any common name specified in the CSR.

With this fix, the behavior is changed to bring back parity with 1.1 behavior (i.e. if registration_id is missing, fallback to identity cert's subject name as registration_id). The CSR issued by IS to renew the identity cert will contain subject name set as one of the following values (in order of precedence):
- First, the IS config.toml X.509 attestation registration_id is used in the CSR to renew the identity cert
- If registration_id is missing, the current identity certificate's subject name is used in the CSR to renew the identity cert
- When the identity certificate is returned, the latest valid/renewed cert's subject is used, if the registration_id is missing.

Note that if registration_id is set to a value different from the identity certificate subject name returned by CS, DPS will reject the registration request (as of DPS API version 2019-04-15). So, it is upto the user to avoid this scenario currently.

### DIFF
--- a/identity/aziot-identityd/src/identity.rs
+++ b/identity/aziot-identityd/src/identity.rs
@@ -853,14 +853,6 @@ impl IdentityManager {
 
         // Create new certificate if needed.
         if device_id_cert.is_none() {
-            if let Ok(key_handle) = self.key_client.load_key_pair(identity_pk).await {
-                self.key_client
-                    .delete_key_pair(&key_handle)
-                    .await
-                    .map_err(|err| {
-                        Error::Internal(InternalError::CreateCertificate(Box::new(err)))
-                    })?;
-            }
             let new_cert_subject = match subject {
                 Some(subject) => subject.to_string(),
                 None => old_cert_subject_name.map_err(|err| {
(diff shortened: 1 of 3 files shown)


identity/aziot-identityd/src/identity.rs
```diff
@@ -853,14 +853,6 @@ impl IdentityManager {
 
         // Create new certificate if needed.
         if device_id_cert.is_none() {
-            if let Ok(key_handle) = self.key_client.load_key_pair(identity_pk).await {
```
arsing: What fix is this? main still has this code.

gordonwang0: Keyd in this repo doesn't have delete_key_pair, so can't do this.

I think it's fine, it just means that the old key won't be rotated. That's what already happens on release/1.2, so it's equivalent.

arsing: Ah, makes sense.

arsing (approved):
Signing off assuming you did a sanity check to test it. (Looks like Rajeev didn't add a test for the config parts.)

PR #279, opened 2021-09-15
Add documentation for configuring EST in certd

gordonwang0:
Adds a guide and sample on how to configure an arbitrary certificate to be issued over EST in certd.

### DIFF
--- a/docs/_includes/nav.html
+++ b/docs/_includes/nav.html
@@ -36,5 +36,7 @@
             <li><a href="/iot-identity-service/api/keys-service.html">Keys Service</a></li>
             <li><a href="/iot-identity-service/api/certificates-service.html">Certificates Service</a></li>
         </ul>
+
+        <li><a href="/iot-identity-service/using-est-with-certd.html">Using EST with Certificates Service</a></li>
     </ul>
 </div>
--- a/docs/_includes/nav.html
+++ b/docs/_includes/nav.html
@@ -37,6 +37,6 @@
             <li><a href="/iot-identity-service/api/certificates-service.html">Certificates Service</a></li>
         </ul>
 
-        <li><a href="/iot-identity-service/using-est-with-certd.html">Issuing Arbitrary Certificates over EST (Advanced)</a></li>
+        <li><a href="/iot-identity-service/using-est-with-certd.html">Issuing Certificates over EST (Advanced)</a></li>
     </ul>
 </div>
--- a/docs/_includes/nav.html
+++ b/docs/_includes/nav.html
@@ -37,6 +37,6 @@
             <li><a href="/iot-identity-service/api/certificates-service.html">Certificates Service</a></li>
         </ul>
 
-        <li><a href="/iot-identity-service/using-est-with-certd.html">Using EST with Certificates Service</a></li>
+        <li><a href="/iot-identity-service/using-est-with-certd.html">Issuing Arbitrary Certificates over EST (Advanced)</a></li>
     </ul>
 </div>
--- a/docs/using-est-with-certd.md
+++ b/docs/using-est-with-certd.md
@@ -29,6 +29,26 @@ trusted_certs = ["cert-id"]
 [cert_issuance.est.urls]
 default = "https://est.example.com/.well-known/est"
 
+# Below are options for authenticating with the EST server. The required options will depend on the EST
+# server's configuration. These global settings apply to all certificates that don't configure auth separately.
+[cert_issuance.est.auth]
+
+# Authentication with TLS client certificate. Provide the cert ID of the client cert and its corresponding
+# private key. Note that the aziotcs user must be authorized to access `identity_pk` in Keys Service.
+identity_cert = "identity-cert-id"
+identity_pk = "identity-cert-pk-id"
+
+# Authentication with a TLS client certificate which will be used once to create the initial certificate.
+# After the first certificate issuance, an identity_cert and identity_pk will be automatically created and
+# used. Provide the cert ID of the bootstrap client cert and its corresponding private key. Note that the
+# aziotcs user must be authorized to access `bootstrap_identity_pk` in Keys Service.
+bootstrap_identity_cert = "bootstrap-identity-cert-id"
+bootstrap_identity_pk = "bootstrap-identity-pk-id"
+
+# Authentication with username and password.
+username = "username"
+password = "password"
+
 # Sample configuration of a single EST-issued certificate.
 # Replace `name` with the desired certificate name.
 [cert_issuance.name]
@@ -47,22 +67,15 @@ expiry_days = 30
 # provided here and no default exists.
 url = "https://est.example.com/.well-known/est"
 
-# Below are options for authenticating with the EST server. The required options will depend on the EST
-# server's configuration.
-
-# Authentication with TLS client certificate. Provide the cert ID of the client cert and its corresponding
-# private key. Note that the aziotcs user must be authorized to access `identity_pk` in Keys Service.
+# It is also possible to configure auth separately for each certificate. The options are the
+# same as in the global EST configuration and override the global configuration for their corresponding
+# certificate.
 identity_cert = "identity-cert-id"
 identity_pk = "identity-cert-pk-id"
 
-# Authentication with a TLS client certificate which will be used once to create the initial certificate.
-# After the first certificate issuance, an identity_cert and identity_pk will be automatically created and
-# used. Provide the cert ID of the bootstrap client cert and its corresponding private key. Note that the
-# aziotcs user must be authorized to access `bootstrap_identity_pk` in Keys Service.
 bootstrap_identity_cert = "bootstrap-identity-cert-id"
 bootstrap_identity_pk = "bootstrap-identity-pk-id"
 
-# Authentication with username and password.
 username = "username"
 password = "password"
 ```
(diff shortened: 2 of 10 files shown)


docs/using-est-with-certd.md
```diff
@@ -0,0 +1,169 @@
...
+This document covers how to configure Certificates Service to issue certificates over EST.
+
+## Prerequisites
+
+- Certificates Service and Keys Service configured and running
+- A working EST server that processes enrollment requests
+
+## Configuration Options
+
+EST certificate issuance is configured in Certificates Service's config.toml, which by default is `/etc/aziot/certd/config.toml` or a file in `/etc/aziot/certd/config.d/`.
```
arsing, line 12: Why tell people to do it in certd config instead of super-config?

gordonwang0: This doc about issuing an arbitrary certificate over EST. The super config only supports issuing the device ID and Edge CA certs over EST (not any arbitrary cert).

I'll be writing another doc that covers the device ID and Edge CA certs specifically. That will use the super config.

docs/using-est-with-certd.md
```diff
@@ -0,0 +1,169 @@
...
+# if not provided. The URL must be provided either here or in default, i.e. certd will fail if no URL is
+# provided here and no default exists.
+url = "https://est.example.com/.well-known/est"
+
+# Below are options for authenticating with the EST server. The required options will depend on the EST
+# server's configuration.
+
+# Authentication with TLS client certificate. Provide the cert ID of the client cert and its corresponding
+# private key. Note that the aziotcs user must be authorized to access `identity_pk` in Keys Service.
+identity_cert = "identity-cert-id"
```
arsing, line 55: Would it be better to document these in the `[cert_issuance.est]` section instead of the per-cert section? I assume most users will have a single server with the same auth, and the only thing that's per cert is the URL.

docs/using-est-with-certd.md
```diff
@@ -0,0 +1,169 @@
...
+openssl req -new -key key.pem -subj "/CN=test-est-cert" -out req.pem
+
+# Escape newlines in CSR.
+csr=$(awk '{printf "%s\\n", $0}' req.pem)
+
+# Make the request to Certificates Service.
+# The user making this request must be root or in the aziotcs group.
+curl --unix-socket /run/aziot/certd.sock http://localhost/certificates?api-version=2020-09-01 \
+    -H "content-type: application/json" \
+    --data "{\"certId\": \"test-est-cert\", \"csr\": \"$csr\"}"
```
arsing, line 164: Use `jq --arg` to construct the request body, which does the escaping for you.

gordonwang0: Fixed

docs/using-est-with-certd.md
gordonwang0: Makes sense. I moved these under the global section but left a note under cert_issuance that it's possible to configure separately.

docs/using-est-with-certd.md
arsing: Just do it inline instead of saving to a file, eg https://github.com/Azure/iot-identity-service/blob/f6294bdd027284a27a03d83b4c5dc736d98e0a32/ci/iothub-get-twin.sh#L247-L256

gordonwang0: Fixed

arsing (approved):

docs/using-est-with-certd.md
```diff
@@ -0,0 +1,187 @@
...
+
+The certd configuration file configures EST certificate issuance.
+
+`/etc/aziot/certd/config.d/test-est-cert.toml`
+
+```toml
+# Configure the trusted root CA certificate in the global EST options. This section is optional
+# if the EST server's TLS certificate is already trusted by the system's CA certificates.
+[cert_issuance.est]
+trusted_certs = ["est-server-ca"]
```
varunpuranik, line 95: Can this also be specified by inlining the file name like this? https://github.com/Azure/iot-identity-service/blob/main/aziotctl/aziotctl-common/test-files/apply/dps-x509-pkcs11-est-bootstrap/config.toml#L16

docs/using-est-with-certd.md
```diff
@@ -0,0 +1,187 @@
...
+
+# Configure the issuance of the EST test certificate.
+[cert_issuance.test-est-cert]
+method = "est"
+
+# Example with the EST path segment `test-est-cert` in the URL.
+# The path will depend on server configuration.
+url = "https://est.example.com/.well-known/est/test-est-cert"
+
+# This example will use the EST bootstrap certificate and key to authenticate,
```
varunpuranik, line 110: Why are these required? Can't we just come up with a unique keys and "remember" them?

docs/using-est-with-certd.md
```diff
@@ -0,0 +1,187 @@
...
+identity_pk = "test-est-cert-identity-id"
+
+# The credentials to use upon initial authentication with the EST server.
+# After that, certd will automatically create, use, and renew identity_cert and identity_pk.
+bootstrap_identity_cert = "test-est-bootstrap-cert-id"
+bootstrap_identity_pk = "test-est-bootstrap-key-id"
+
+# Load the bootstrap certificate from a file into certd.
+# This file must be readable by aziotcs.
+[preloaded_certs]
```
varunpuranik, line 124: Just to confirm - this preloaded_certs seems to be in multiple places. I suppose it is a top level item, so you have just put it in the logical places?

docs/using-est-with-certd.md
```diff
@@ -0,0 +1,187 @@
...
+This document covers how to configure Certificates Service to issue certificates over EST.
+
+## Prerequisites
+
+- Certificates Service and Keys Service configured and running
+- A working EST server that processes enrollment requests
+
+## Configuration Options
+
+EST certificate issuance is configured in Certificates Service's config.toml, which by default is `/etc/aziot/certd/config.toml` or a file in `/etc/aziot/certd/config.d/`.
```
micahl, line 12: After a default installation I only see a `/etc/aziot/certd/config.toml.default` file - no `config.toml`.  Are you referring to the config.toml.default or would the guidance be that someone can simply create a config.toml and certd will know to pick it up by default?  It's not clear here so we should clarify.

micahl, line 12: Just to confirm... the config files we're telling a customer to modify here would not be overwritten by someone applying their uber config, right?

docs/using-est-with-certd.md
```diff
@@ -0,0 +1,187 @@
+# Using EST with Certificates Service
```
micahl, line 1: Recommend that we re-title this in a way that it is answering the question for a customer "If I read this page, what will it enable me to accomplish?".   So, maybe something more like "Issuing Arbitrary Certificates over EST with Certificates Service"?

docs/using-est-with-certd.md
```diff
@@ -0,0 +1,187 @@
...
+
+This document covers how to configure Certificates Service to issue certificates over EST.
```
micahl, line 3: This would also be a good place to re-emphasize the purpose of this doc and also when to NOT read this document (e.g. This document addresses the more advanced scenario of configuring the Certificates Service to issue arbitrary certificates over EST.  Refer to the [other doc] that covers the common scenario of issuing the device ID and Edge CA certs.)

docs/using-est-with-certd.md
gordonwang0: No, it's not possible for this file. It needs to be specified as a cert ID (with the file then loaded under preloaded_certs).

docs/using-est-with-certd.md
gordonwang0: Yes, it's fine to put it in multiple places.

docs/using-est-with-certd.md
gordonwang0: We could, but if the customer wants to access or change these credentials it might not be obvious what the names are.

docs/_includes/nav.html
```diff
@@ -36,5 +36,7 @@ <h3>Contents</h3>
             <li><a href="/iot-identity-service/api/keys-service.html">Keys Service</a></li>
             <li><a href="/iot-identity-service/api/certificates-service.html">Certificates Service</a></li>
         </ul>
+
+        <li><a href="/iot-identity-service/using-est-with-certd.html">Using EST with Certificates Service</a></li>
```
micahl, line 40: As we're adding some more docs it might be a good time to also introduce some additional structure to our nav.  We could do this as a separate change.  My thought is that we have a "How-to guides" and an Reference section.  For example: 

* Overview
* How-to guides
  * Installing the Identity Service
  * Configuring the Identity Service
  * Managing the Identity Service
  * Protecting Keys via an HSM or TPM
    * tpm2-pkcs11
    * cryptoauthlib
    * softhsm
  * Creating an IoT agent
  * Issuing Arbitrary Certificates over EST (Advanced)
* Reference
  * Identity Service API
  * Keys Service API
  * Certificates Service API

docs/using-est-with-certd.md
varunpuranik: so do we need to change that test config.toml?

docs/using-est-with-certd.md
varunpuranik: but can/should they? What is the use case?

docs/using-est-with-certd.md
varunpuranik: I think you need to rename/copy config.toml.default to config.toml, and edit it as described here. Maybe we should specify that here?

docs/using-est-with-certd.md
varunpuranik: I like that.. maybe drop Arbitrary..?

docs/using-est-with-certd.md
varunpuranik: Yes, I agree.. in fact we need to put a link to the other doc that talks about IoT Edge scenarios..

docs/using-est-with-certd.md
gordonwang0: The expectation is for the user to just create the files. I added a section.

gordonwang0: The uber config only modifies `/etc/aziot/certd/config.d/00-super.toml`, so changes won't be overwritten as long as the customer chooses another file name.

docs/using-est-with-certd.md
gordonwang0: Updated

docs/using-est-with-certd.md
gordonwang0: Updated

docs/_includes/nav.html
gordonwang0: I updated the title, but we should change the nav in another PR.

docs/using-est-with-certd.md
gordonwang0: No, that test file is a super config. The file:// URI format for that file works.

docs/using-est-with-certd.md
gordonwang0: If customers are issuing EST certs from two endpoints that use the same auth (e.g. we have test servers that issue Edge CA and device ID certs on different endpoints but with the same auth), then they can set the same credential IDs for two different cert issuances.

micahl (approved):

PR #286, opened 2021-09-21
http server: don't try to parse empty bodies

gordonwang0:
https://github.com/Azure/iot-identity-service/commit/f6294bdd027284a27a03d83b4c5dc736d98e0a32 allowed the content-type header to be omitted because some Edge modules do not provide it even when content is provided. But that commit introduced a bug where the server would parse empty bodies as JSON, since empty bodies do not provide a content-type header.

This PR fixes the DELETE and POST methods to only parse nonzero-length bodies. PUT always requires a body, so it was left as-is.

### DIFF
--- a/http-common/src/server.rs
+++ b/http-common/src/server.rs
@@ -96,24 +96,27 @@ macro_rules! make_service {
                                 return Box::pin(async move {
                                     let response = match method {
                                         http::Method::DELETE => {
-                                            let body = {
+                                            let body = match tokio::time::timeout(HYPER_REQUEST_TIMEOUT, hyper::body::to_bytes(body)).await {
+                                                Ok(result) => match result {
+                                                    Ok(body) => body,
+                                                    Err(err) => return Ok((http_common::server::Error {
+                                                        status_code: http::StatusCode::BAD_REQUEST,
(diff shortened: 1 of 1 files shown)


http-common/src/server.rs
```diff
@@ -96,24 +96,27 @@ macro_rules! make_service {
                                 return Box::pin(async move {
                                     let response = match method {
                                         http::Method::DELETE => {
-                                            let body = {
+                                            let body = match tokio::time::timeout(HYPER_REQUEST_TIMEOUT, hyper::body::to_bytes(body)).await {
+                                                Ok(result) => match result {
```
arsing, line 100: You could un-nest the `match` to make it slightly more compact:

```rust
let body = match tokio::time::timeout(HYPER_REQUEST_TIMEOUT, hyper::body::to_bytes(body)).await {
    Ok(Ok(body)) => body,
    Ok(Err(err)) => return Ok((http_common::server::Error { ... }).to_http_response(),
    Err(timeout_err) => return Ok((http_common::server::Error { ... }).to_http_response(),
};
```

gordonwang0: Fixed

nimanch (approved):

PR #288, opened 2021-09-23
Update tokio to v0.12.0

gordonwang0:
tokio v0.11.0 carries a fix to prevent `tokio::time::Instant` from panicking
if `std::time::Instant` happens to go backwards for any reason. We've observed
this happening in Linux VMs and possibly even on Linux hosts, so this change
updates to the latest tokio to pull in that fix.


arsing (approved):
(I changed the PR title and description to explain why we're doing it.)

PR #289, opened 2021-09-23
Fix method used in delete_key_pair and delete_key

gordonwang0:


gordonwang0:
delete_key_pair is only called when renewing identity certs, so that's probably why we didn't see this before.

key/aziot-key-client-async/src/lib.rs
```diff
@@ -95,7 +95,7 @@ impl Client {
 
         http_common::request_no_content(
             &self.inner,
-            http::Method::GET,
+            http::Method::DELETE,
```
arsing, line 98: Oops.

`delete_key` is wrong too, so fix that too please.

gordonwang0: Fixed

arsing (approved):

PR #292, opened 2021-09-30
Add retry logic to identity client

lfitchett:
This adds retry logic so that edged will retry some requests. A follow-up pr for the iotedge repo will make edged use the new functionality.


gordonwang0 (approved):

http-common/src/request.rs
```diff
@@ -16,6 +16,41 @@ where
...
+) -> std::io::Result<TResponse>
+where
+    TRequest: serde::Serialize,
+    TResponse: serde::de::DeserializeOwned,
+{
+    let mut retry_num = 0;
+    loop {
+        match request(client, method.clone(), uri, body).await {
+            Ok(response) => return Ok(response),
+            Err(err) => {
```
arsing, line 34: Retry should only happen for throttling due to connection closed.

arsing (changes requested):

http-common/src/request.rs
lfitchett: is there a reason to not just always retry?

arsing: Key does not exist already and needs to be created first - error.

Cert does not exist already and needs to be created first - error.

We had discussed already that we only want to retry the requests to host services in case of throttling. The only cases of retrying for all errors is with network services, ie Hub and DPS.

lfitchett: ok, then in that case should I still call it max_retries and request_with_retry? I'm having trouble thinking of a good name for just retrying on throttling

arsing: Names are fine. Just add a doc comment.

(Ideally Hub and DPS clients would also have used `http_common` instead of `hyper::Client::request` directly. Then `request_*_with_retry` could check what kind of `Connector` it is and behave accordingly. `Connector::Unix` => retry only on connection-closed, `Connector::Tcp` => retry on all errors.)

http-common/src/request.rs
```diff
@@ -179,61 +150,80 @@ async fn make_call<TUri, TRequest>(
...
+                    body,
+                ) = response.into_parts();
+
+                let body = hyper::body::to_bytes(body)
+                    .await
+                    .map_err(|err| std::io::Error::new(std::io::ErrorKind::Other, err))?;
+
+                return Ok((status, headers, body));
+            }
+            Err(err) if err.is_closed() && retry_num < max_retries => {
```
arsing, line 215: Are you sure? https://docs.rs/hyper/0.14.13/hyper/struct.Error.html#method.is_closed says it's for the sender (of the request body), not the underlying connection.

lfitchett: Yep. I wrote code to send 11 (or more) concurrent requests to IS, and looked at the error code.

Console Output:
image

edged generating the error with 11 requests: 
image

request.rs printing the error:
image

http-common/src/request.rs
```diff
@@ -98,73 +138,94 @@ where
...
+                log::warn!(
```
varunpuranik, line 216: Will this flood our logs? Let's check.. if it is not very common, then it is okay, else we can make it debug..

http-common/src/request.rs
varunpuranik: Remind me again - what does IS return when it throttles a request?

aziotctl/src/internal/check/checks/read_certs.rs
```diff
@@ -43,6 +43,7 @@ impl ReadCerts {
         let cert_client = aziot_cert_client_async::Client::new(
             aziot_cert_common_http::ApiVersion::V2020_09_01,
             aziot_certd.clone(),
+            0,
```
varunpuranik, line 46: why 0 retries here?

http-common/src/request.rs
lfitchett: It should be very uncommon.

aziotctl/src/internal/check/checks/read_certs.rs
lfitchett: I thought it made sense to not retry during check. If the system is overwhelmed enough to force a retry, I think it's better to fail the check then to pass. It should be extremally rare.

http-common/src/request.rs
lfitchett: The `Http Error: channel closed` in the logs above is from IS. The code I used to test just sends 11 concurrent requests and prints the error IS causes. 

IS isn't actually sending an error, we don't get far enough into the request handling to actually send one. IS just closes the incoming connection instead.

http-common/src/request.rs
```diff
@@ -184,3 +245,11 @@ fn validate_json(headers: hyper::HeaderMap) -> std::io::Result<()> {
         "malformed HTTP response, expected content type application/json",
     ))
 }
+
+#[cfg(test)]
+mod test {
+    #[test]
+    fn unit_type_deserializes() {
```
vadim-kovalyov, line 252: is this needed? you're just testing `serde_json`, not our code.

arsing (approved):

PR #294, opened 2021-10-02
E2E test script and workflow changes.

arsing:
- Create the IoT Hub and DPS at the suite level so that they are reused
  for all runs in the suite. This is because ARM is buggy and has
  various issues that make creating a new IoT Hub and DPS for every test
  untenable:

  - `az iot hub create` never completes even though the IoT Hub gets created
    just fine, because the async operation remains in "Running" state.

  - `az iot hub device-identity create` internally lists all hubs in the RG
    to discover the hub it's creating a device identity in, but this query
    randomly decides to return no Hubs at all in the RG. Forcing it to retry
    eventually causes it to return the Hub, but only for some tests.
    In other words, it can be simultaneously the case that one test sees
    Hubs in the RG and another doesn't.

  So this change refactors the e2e-tests.sh into multiple scripts for
  suite and test operations.

  To solve the problem of `az iot hub list` returning no Hubs, this change
  changes the `az iot hub device-identity {create,delete}` invocations to use
  `--login $iot_hub_connection_string` instead of `--hub-name $hub_name`.
  `az iot hub device-identity {create,delete}` don't try to list all Hubs
  when `--login` is provided.

- Include the branch name in the common resource name since the scheduled runs
  run against both main and release/1.2, so they need to use
  different resource names to avoid stepping on each other.

- Fix RedHat 8 package makefile target to derive the correct name for
  the aziot-keys OpenSSL engine.

- Fix RedHat 8 test runs to calculate the correct package artifact name,
  ie the one specified in packages.yaml

- Update VM images for E2E tests to the latest. The previous images would
  sometimes not come back after installing updates and rebooting, for
  unknown reason.

- `az iot hub create` now supports `--tags`, so the IoT Hub can be tagged at
  creation time instead of afterwards, removing the risk of leaking it.

- Change the VM image query samples to print the image URN, since that is what
  needs to be copied into the `vm_image` script variable.

- Switch to writing super-config and running `aziotctl config apply`
  instead of writing the individual services' configs.

- Enforce `az vm create` creates a Basic SKU public IP address for the VM.
  This is currently the default but `az vm create` was printing a warning that
  it will default to Standard in the future, which we don't need.

- Enforce time limits on the job steps since the default timeout of
  six hours is unnecessarily long.

- Limit parallelism to 10 for the manual workflow too.


arsing:
Successful run in my repo: https://github.com/arsing/iot-identity-service/actions/runs/1299027654

ci/e2e-tests/suite-setup.sh
```diff
@@ -0,0 +1,87 @@
...
+
+GITHUB_WORKSPACE="${GITHUB_WORKSPACE:-$PWD}"
+
+. "$GITHUB_WORKSPACE/ci/e2e-tests/suite-common.sh"
+
+
+. "$GITHUB_WORKSPACE/ci/e2e-tests/az-login.sh"
+
+
+# For some reason, `az iot hub create` never returns until the run is timed out,
```
varunpuranik, line 23: We should check with the hub folks about this.

varunpuranik (approved):

.github/workflows/e2e-tests-scheduled.yaml
```diff
@@ -59,3 +79,66 @@ jobs:
         AZURE_LOCATION: "${{ secrets.AZURE_LOCATION }}"
 
         GITHUB_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+
+    - name: 'Clean up on failure'
+      run: |
+        timeout 5m ./ci/e2e-tests/test-cleanup.sh ${{ matrix.test_name }}
+        exit 1 # Force the workflow to fail even if this step succeeds
```
vadim-kovalyov, line 86: q: does this mean that Actions report success if one intermediate step failed, but the subsequent is successful? Asking this b/c it is different in ADO, where if any step fails - the whole run marked as failed.

ci/e2e-tests/az-login.sh
```diff
@@ -0,0 +1,21 @@
...
+    --tenant "$AZURE_TENANT_ID" \
+    --username "$AZURE_USERNAME" \
+    "--password=$AZURE_PASSWORD"
+echo 'Logged in to Azure' >&2
+
+if [ -n "${AZURE_LOCATION:-}" ]; then
+    az configure --defaults "location=$AZURE_LOCATION"
+fi
+
+az config set core.collect_telemetry=no
```
vadim-kovalyov, line 21: nit: any specific reason to turn that off?

vadim-kovalyov (approved):

.github/workflows/e2e-tests-scheduled.yaml
arsing: Yes, that's what I saw. A task that runs on `${{ failure() }}` and succeeds marks the whole workflow as succeeded.

To be clear, this is specifically because of `fail-fast: false` and because this is a `${{ failure() }}`-gated task.

ci/e2e-tests/az-login.sh
arsing: It was being noticably slow in my local testing to the point that it got annoying. Also something-something-telemetry-should-be-opt-in-by-default.

ci/e2e-tests/az-login.sh
```diff
@@ -0,0 +1,21 @@
...
+az extension add --name azure-iot
+az extension update --name azure-iot
+echo 'Installed and updated azure-iot extension' >&2
+
+
+echo 'Logging in to Azure...' >&2
+>/dev/null az login --service-principal \
+    --tenant "$AZURE_TENANT_ID" \
+    --username "$AZURE_USERNAME" \
+    "--password=$AZURE_PASSWORD"
```
gordonwang0, line 14: Any reason that "--password arg" is quoted but tenant and username only have the arg quoted?

gordonwang0 (approved):

ci/e2e-tests/az-login.sh
arsing: IIRC there was a reason I had to use `--password=foo` instead of `--password foo`, so the quoting is just a result of that. (This is old code, just moved from the old `e2e-tests.sh`.)

To be precise, bash will allow `--password="$AZURE_PASSWORD"` too, but I find it cleaner to quote entire shell words rather than parts of them.

arsing: Oh, actually, `az login --help` reminded me:

>```
>    Log in with user name and password. This doesn't work with Microsoft accounts or accounts that
>    have two-factor authentication enabled. Use -p=secret if the first character of the password is
>    '-'.
>        az login -u johndoe@contoso.com -p VerySecret
>```

PR #296, opened 2021-10-07
Subject DN configuration

onalante-msft:
This PR adds subject DN configuration options to `aziot_certd_config`, and includes a refactor of `create_cert` in `aziot_certd`.

### DIFF
--- a/cert/aziot-certd-config/src/lib.rs
+++ b/cert/aziot-certd-config/src/lib.rs
@@ -15,8 +15,8 @@ use std::collections::BTreeMap;
 use std::convert::TryFrom;
 use std::path::{Path, PathBuf};
 
-use serde::{Deserialize, Serialize};
 use serde::de::Error as _;
+use serde::{Deserialize, Serialize};
 use serde_with::{skip_serializing_none, with_prefix};
 use url::Url;
 
@@ -98,7 +98,7 @@ pub struct EstAuth {
 
     /// Authentication parameters when using TLS client cert authentication.
     #[serde(flatten)]
-    pub x509: Option<EstAuthX509>
+    pub x509: Option<EstAuthX509>,
 }
 
 #[derive(Deserialize)]
@@ -106,7 +106,7 @@ struct EstAuthInner {
     #[serde(flatten)]
     pub basic: Option<EstAuthBasic>,
     #[serde(flatten)]
-    pub x509: Option<EstAuthX509>
+    pub x509: Option<EstAuthX509>,
 }
 
 impl TryFrom<EstAuthInner> for EstAuth {
@@ -115,9 +115,10 @@ impl TryFrom<EstAuthInner> for EstAuth {
     fn try_from(value: EstAuthInner) -> Result<Self, Self::Error> {
         let EstAuthInner { basic, x509 } = value;
         if basic.is_none() && x509.is_none() {
-            Err(Self::Error::missing_field("empty authentication parameters"))
-        }
-        else {
+            Err(Self::Error::missing_field(
+                "empty authentication parameters",
+            ))
+        } else {
             Ok(Self { basic, x509 })
         }
     }
@@ -181,7 +182,7 @@ pub struct CertIssuanceOptions {
 #[serde(rename_all = "snake_case")]
 pub enum CertSubject {
     CommonName(String),
-    Subject(BTreeMap<String, String>)
+    Subject(BTreeMap<String, String>),
 }
 
 pub fn deserialize_expiry_days<'de, D>(deserializer: D) -> Result<Option<u32>, D::Error>
@@ -332,7 +333,10 @@ certs = ["test"]
                         auth: EstAuth {
                             basic: None,
                             x509: Some(EstAuthX509 {
-                                identity: CertificateWithPrivateKey { cert: "est-id".to_owned(), pk: "est-id".to_owned() },
+                                identity: CertificateWithPrivateKey {
+                                    cert: "est-id".to_owned(),
+                                    pk: "est-id".to_owned()
+                                },
                                 bootstrap_identity: Some(CertificateWithPrivateKey {
                                     cert: "bootstrap".to_owned(),
                                     pk: "bootstrap".to_owned()
@@ -400,7 +404,9 @@ certs = ["test"]
                                         ("L".to_owned(), "AQ".to_owned()),
                                         ("ST".to_owned(), "Antarctica".to_owned()),
                                         ("CN".to_owned(), "test-device".to_owned())
-                                    ].into_iter().collect()
+                                    ]
+                                    .into_iter()
+                                    .collect()
                                 )),
                                 expiry_days: Some(365)
                             }
@@ -430,15 +436,11 @@ certs = ["test"]
                 preloaded_certs: vec![
                     (
                         "bootstrap".to_owned(),
-                        PreloadedCert::Uri(
-                            "file:///var/secrets/bootstrap.cer".parse().unwrap()
-                        )
+                        PreloadedCert::Uri("file:///var/secrets/bootstrap.cer".parse().unwrap())
                     ),
                     (
                         "est-ca".to_owned(),
-                        PreloadedCert::Uri(
-                            "file:///var/secrets/est-ca.cer".parse().unwrap()
-                        )
+                        PreloadedCert::Uri("file:///var/secrets/est-ca.cer".parse().unwrap())
                     ),
                     (
                         "trust-bundle".to_owned(),
@@ -508,31 +510,34 @@ aziot_certd = "unix:///run/aziot/certd.sock"
 
             cert_issuance: CertIssuance {
                 est: Some(Est {
-                    trusted_certs: vec!["est-ca".to_owned(),],
+                    trusted_certs: vec!["est-ca".to_owned()],
                     auth: EstAuth {
                         basic: None,
                         x509: Some(EstAuthX509 {
-                            identity: CertificateWithPrivateKey { cert: "est-id".to_owned(), pk: "est-id".to_owned() },
+                            identity: CertificateWithPrivateKey {
+                                cert: "est-id".to_owned(),
+                                pk: "est-id".to_owned(),
+                            },
                             bootstrap_identity: Some(CertificateWithPrivateKey {
                                 cert: "bootstrap".to_owned(),
-                                pk: "bootstrap".to_owned()
+                                pk: "bootstrap".to_owned(),
                             }),
-                        })
+                        }),
                     },
                     urls: vec![
                         (
                             "default".to_owned(),
-                            "https://estendpoint.com/.well-known/est/".parse().unwrap()
+                            "https://estendpoint.com/.well-known/est/".parse().unwrap(),
                         ),
                         (
                             "device-ca".to_owned(),
                             "https://estendpoint.com/.well-known/est/device-ca/"
                                 .parse()
-                                .unwrap()
+                                .unwrap(),
                         ),
                     ]
                     .into_iter()
-                    .collect()
+                    .collect(),
                 }),
 
                 local_ca: None,
@@ -543,11 +548,11 @@ aziot_certd = "unix:///run/aziot/certd.sock"
                         CertIssuanceOptions {
                             method: CertIssuanceMethod::Est {
                                 url: None,
-                                auth: None
+                                auth: None,
                             },
                             subject: Some(CertSubject::CommonName("custom-name".to_owned())),
                             expiry_days: None,
-                        }
+                        },
                     ),
                     (
                         "device-id",
@@ -556,7 +561,7 @@ aziot_certd = "unix:///run/aziot/certd.sock"
                                 url: Some(
                                     "https://estendpoint.com/.well-known/est/device-id/"
                                         .parse()
-                                        .unwrap()
+                                        .unwrap(),
                                 ),
                                 auth: Some(EstAuth {
                                     basic: Some(EstAuthBasic {
@@ -566,18 +571,18 @@ aziot_certd = "unix:///run/aziot/certd.sock"
                                     x509: Some(EstAuthX509 {
                                         identity: CertificateWithPrivateKey {
                                             cert: "device-id".to_owned(),
-                                            pk: "device-id".to_owned()
+                                            pk: "device-id".to_owned(),
                                         },
                                         bootstrap_identity: Some(CertificateWithPrivateKey {
                                             cert: "bootstrap".to_owned(),
-                                            pk: "bootstrap".to_owned()
+                                            pk: "bootstrap".to_owned(),
                                         }),
-                                    })
-                                })
+                                    }),
+                                }),
                             },
                             subject: Some(CertSubject::CommonName("test-device".to_owned())),
-                            expiry_days: Some(365)
-                        }
+                            expiry_days: Some(365),
+                        },
                     ),
                     (
                         "module-id",
@@ -585,7 +590,7 @@ aziot_certd = "unix:///run/aziot/certd.sock"
                             method: CertIssuanceMethod::SelfSigned,
                             subject: Some(CertSubject::CommonName("custom-name".to_owned())),
                             expiry_days: Some(90),
-                        }
+                        },
                     ),
                     (
                         "module-server",
@@ -593,7 +598,7 @@ aziot_certd = "unix:///run/aziot/certd.sock"
                             method: CertIssuanceMethod::LocalCa,
                             subject: None,
                             expiry_days: None,
-                        }
+                        },
                     ),
                 ]
                 .iter()
@@ -604,19 +609,15 @@ aziot_certd = "unix:///run/aziot/certd.sock"
             preloaded_certs: vec![
                 (
                     "bootstrap".to_owned(),
-                    PreloadedCert::Uri(
-                        "file:///var/secrets/bootstrap.cer".parse().unwrap()
-                    )
+                    PreloadedCert::Uri("file:///var/secrets/bootstrap.cer".parse().unwrap()),
                 ),
                 (
                     "est-ca".to_owned(),
-                    PreloadedCert::Uri(
-                        "file:///var/secrets/est-ca.cer".parse().unwrap()
-                    )
+                    PreloadedCert::Uri("file:///var/secrets/est-ca.cer".parse().unwrap()),
                 ),
                 (
                     "trust-bundle".to_owned(),
-                    PreloadedCert::Ids(vec!["est-ca".to_owned()])
+                    PreloadedCert::Ids(vec!["est-ca".to_owned()]),
                 ),
             ]
             .into_iter()
@@ -624,24 +625,21 @@ aziot_certd = "unix:///run/aziot/certd.sock"
 
             endpoints: Endpoints {
                 aziot_certd: Connector::Unix {
-                    socket_path: Path::new("/run/aziot/certd.sock").into()
+                    socket_path: Path::new("/run/aziot/certd.sock").into(),
                 },
                 aziot_keyd: Connector::Unix {
-                    socket_path: Path::new("/run/aziot/keyd.sock").into()
+                    socket_path: Path::new("/run/aziot/keyd.sock").into(),
                 },
             },
 
             principal: vec![Principal {
                 uid: 1000,
-                certs: vec!["test".to_string()]
+                certs: vec!["test".to_string()],
             }],
         };
 
         let serialized = toml::to_string(&configuration).unwrap();
 
-        assert_eq!(
-            configuration,
-            toml::from_str(&serialized).unwrap()
-        );
+        assert_eq!(configuration, toml::from_str(&serialized).unwrap());
     }
 }
--- a/cert/aziot-certd/src/error.rs
+++ b/cert/aziot-certd/src/error.rs
@@ -53,7 +53,7 @@ pub enum InternalError {
     InvalidProxyUri(Box<dyn std::error::Error + Send + Sync>),
     LoadKeyOpensslEngine(openssl2::Error),
     ReadFile(std::io::Error),
-    WriteFile(std::io::Error)
+    WriteFile(std::io::Error),
 }
 
 impl std::fmt::Display for InternalError {
@@ -69,7 +69,7 @@ impl std::fmt::Display for InternalError {
                 f.write_str("could not load aziot-key-openssl-engine")
             }
             InternalError::ReadFile(_) => f.write_str("could not read cert file"),
-            InternalError::WriteFile(_) => f.write_str("could not write cert file")
+            InternalError::WriteFile(_) => f.write_str("could not write cert file"),
         }
     }
 }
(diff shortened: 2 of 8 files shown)


cert/aziot-certd-config/src/lib.rs
```diff
@@ -45,130 +53,110 @@ pub struct Config {
...
     /// Authentication parameters when using TLS client cert authentication.
+    #[serde(flatten)]
+    pub x509: Option<EstAuthX509>,
+
+    /// Headers to inject into authentication request.
+    pub headers: Option<BTreeMap<String, String>>
+}
+
+#[derive(Deserialize)]
+struct _EstAuth {
```
arsing, line 107: Leading `_` in Rust means "unused". Something like the original `EstAuthInner` or `EstAuth_` would be better.

cert/aziot-certd/src/error.rs
```diff
@@ -48,26 +48,28 @@ impl std::error::Error for Error {
 #[derive(Debug)]
 pub enum InternalError {
     CreateCert(Box<dyn std::error::Error + Send + Sync>),
-    DeleteFile(std::io::Error),
     GetPath(Box<dyn std::error::Error + Send + Sync>),
     InvalidProxyUri(Box<dyn std::error::Error + Send + Sync>),
     LoadKeyOpensslEngine(openssl2::Error),
+    DeleteFile(std::io::Error),
```
arsing, line 54: Variants should be in alphabetical order.

cert/aziot-certd/src/est.rs
```diff
@@ -1,20 +1,29 @@
 // Copyright (c) Microsoft. All rights reserved.
 
+use aziot_certd_config::EstAuthBasic;
```
arsing, line 3: Same comment about order and grouping of imports.

cert/aziot-certd/src/lib.rs
```diff
@@ -11,28 +11,46 @@
...
-use async_trait::async_trait;
-
 mod error;
-use error::{Error, InternalError};
-
 mod est;
-
 mod http;
 
+use error::{Error, InternalError};
```
arsing, line 18: Same comment about order and grouping of imports.

cert/aziot-certd/src/lib.rs
```diff
@@ -195,654 +211,379 @@ impl UpdateConfig for Api {
...
 
-fn create_cert<'a>(
+fn create_cert_inner<'a>(
     api: &'a mut Api,
     id: &'a str,
     csr: &'a [u8],
-    issuer: Option<(&'a str, &'a aziot_key_common::KeyHandle)>,
-) -> std::pin::Pin<Box<dyn std::future::Future<Output = Result<Vec<u8>, Error>> + Send + 'a>> {
+    issuer: Option<(&'a str, &'a KeyHandle)>
+) -> Pin<Box<dyn Future<Output = Result<Vec<u8>, BoxedError>> + Send + 'a>> {
```
arsing, line 239: Why is this returning `Box<dyn std::error::Error>` instead of our `Error` ?

http-common/src/server.rs
```diff
@@ -1,5 +1,6 @@
 // Copyright (c) Microsoft. All rights reserved.
 
+#[cfg(feature = "tokio1")]
```
arsing, line 3: Just get rid of this feature entirely. We don't need it now that Gordon and Lee have completed the tokio v1 refactor of the iotedge repo.

arsing (changes requested):

cert/aziot-certd-config/src/lib.rs
```diff
@@ -45,130 +53,110 @@ pub struct Config {
...
-#[derive(Clone, Debug, PartialEq)]
+#[derive(Clone, Debug, Deserialize, PartialEq, Serialize)]
 pub struct EstAuthX509 {
     /// Cert ID and private key ID for the identity cert.
     ///
     /// If this cert does not exist, it will be requested from the EST server,
     /// with the bootstrap identity cert used as the initial TLS client cert.
-    pub identity: (String, String),
+    #[serde(flatten, with = "prefix_identity")]
+    pub identity: CertAuthority,
```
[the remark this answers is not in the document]
onalante-msft: [Updated](https://github.com/Azure/iot-identity-service/blob/0d3cf64e71eee726f7cb8e1fd4854373030797ce/cert/aziot-certd-config/src/lib.rs#L160)

cert/aziot-certd/src/error.rs
onalante-msft: [Updated](https://github.com/Azure/iot-identity-service/blob/6dc85fd27362b0013551bc39f633711bad41c84f/cert/aziot-certd/src/error.rs#L49-L90)

cert/aziot-certd/src/lib.rs
```diff
@@ -195,654 +211,379 @@ impl UpdateConfig for Api {
...
-                            identity_cert,
-                        ) {
-                            Ok(Some(identity_cert)) => {
-                                match api.key_client.load_key_pair(identity_private_key) {
-                                    Ok(identity_private_key) => {
-                                        Ok((identity_cert, identity_private_key))
-                                    }
-                                    Err(err) => Err(format!(
+                                    acc.extend(X509::stack_from_pem(&pem)?);
+                                    Result::<_, BoxedError>::Ok(acc)
```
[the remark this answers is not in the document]
onalante-msft: [Updated](https://github.com/Azure/iot-identity-service/blob/3e44872269a9a6fe43e0e6e791332784f82eaa2a/cert/aziot-certd/src/lib.rs#L412)

cert/aziot-certd/src/lib.rs
onalante-msft: Most errors that were not previously marked as invalid parameter errors were being wrapped in `Error::Internal(InternalError::CreateCert(Box::new(err)))`, so I made the inner function return a boxed error that we could later wrap in `Api::create_cert` using the same method.

cert/aziot-certd/src/lib.rs
```diff
@@ -195,654 +211,379 @@ impl UpdateConfig for Api {
...
+                .and_then(|opts| opts.subject.as_ref())
+                .map(build_name)
+                .transpose()?;
+
+            let req = X509Req::from_pem(csr)
+                .or_else(|_| X509Req::from_der(csr))?;
+            let pubkey = req.public_key()?;
+
+            if !req.verify(&pubkey)? {
+                Err("CSR failed to be verified with its public key".to_owned())?
```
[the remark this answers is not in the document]
onalante-msft: Parameter validation is now done before entering `create_cert_inner`: [reference](https://github.com/Azure/iot-identity-service/blob/3e44872269a9a6fe43e0e6e791332784f82eaa2a/cert/aziot-certd/src/lib.rs#L126-L146).

cert/aziot-certd/src/lib.rs
```diff
@@ -106,15 +123,46 @@ impl Api {
...
+        let pubkey = req.public_key()
+            .map_err(|err| Error::invalid_parameter("csr", err))?;
+
+        if !req.verify(&pubkey)
+            .map_err(|err| Error::invalid_parameter("csr", err))?
+        {
+            Err(Error::invalid_parameter(
+                "csr",
+                "CSR failed to be verified with its public key".to_owned()
+            ))?
```
arsing, line 141: Use `return Err();` instead of `Err()?`

(I'm surprised https://rust-lang.github.io/rust-clippy/master/index.html#try_err didn't fire. `clippy::all` should've enabled it.)

cert/aziot-certd/src/lib.rs
```diff
@@ -195,654 +242,356 @@ impl UpdateConfig for Api {
...
-                if let Some(c) = &options.common_name {
-                    let mut name_builder = openssl::x509::X509Name::builder()
-                        .map_err(|err| Error::Internal(InternalError::CreateCert(Box::new(err))))?;
-
-                    name_builder
-                        .append_entry_by_text("CN", c)
-                        .map_err(|err| Error::Internal(InternalError::CreateCert(Box::new(err))))?;
-                    common_name = name_builder.build();
-                    subject_name = &common_name;
+    csr: (&'a X509Req, &'a PKey<Public>),
```
arsing, line 269: Since the first thing you do with it is destructure it:

```suggestion
    (req, pubkey): (&'a X509Req, &'a PKeyRef<Public>),
```

cert/aziot-certd/src/lib.rs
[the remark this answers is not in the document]
arsing: >Parameter validation is now done before entering `create_cert_inner`: [reference](https://github.com/Azure/iot-identity-service/blob/6dc85fd27362b0013551bc39f633711bad41c84f/cert/aziot-certd/src/lib.rs#L127-L147).

Not true: https://github.com/Azure/iot-identity-service/pull/296/files#diff-ce1948fdb394a64e8a42f87d92e2057992c63a20520bad0c4c7359b64d0854c2R316 (cert/aziot-certd/src/lib.rs:316 in case it doesn't load.)

This kind of mistake is easy to make again in the future. Please change it back to returning `Error` so that every error return is consciously converted to the appropriate error kind.

cert/aziot-certd/src/lib.rs
```diff
@@ -195,654 +242,356 @@ impl UpdateConfig for Api {
...
+    issuer: Option<(&'a str, &'a CString)>
```
arsing, line 270:
```suggestion
    issuer: Option<(&'a str, &'a CStr)>
```

cert/aziot-certd/src/lib.rs
```diff
@@ -195,654 +242,356 @@ impl UpdateConfig for Api {
...
+) -> Result<Vec<u8>, BoxedError> {
+    let cert_options = api.cert_issuance.certs.get(id);
+
+    let (req, pubkey) = csr;
+
+    if let Some(ref issuer) = issuer {
```
arsing, line 276:
```suggestion
    if let Some((issuer_id, issuer_pk)) = issuer {
```

cert/aziot-certd/src/lib.rs
```diff
@@ -195,654 +242,356 @@ impl UpdateConfig for Api {
...
+                let api = api_.into_inner();
 
-                let mut x509 = x509
-                    .to_pem()
-                    .map_err(|err| Error::Internal(InternalError::CreateCert(Box::new(err))))?;
+                create_cert_inner(api, id, csr, Some((id, &CString::new(pk.0)?)))
+                    .await
+            },
+            CertIssuanceMethod::LocalCa => {
+                let CertificateWithPrivateKey { cert, pk } = api.cert_issuance.local_ca.clone()
```
arsing, line 365: No reason to `clone()` it. Only `pk.0` is moved, so just `clone()` that as part of line 375.

cert/aziot-certd/src/lib.rs
```diff
@@ -195,654 +242,356 @@ impl UpdateConfig for Api {
...
+
+                // NOTE: Hints for borrow checker since closure uses mutable reference.
+                let homedir_path = &api.homedir_path;
+                let preloaded_certs = &api.preloaded_certs;
+                let key_client = &api.key_client;
+                let key_engine = &mut api.key_engine;
+                let id_opt = auth.x509.as_ref()
+                    .map(|x509| &x509.identity)
+                    .and_then(move |CertificateWithPrivateKey { cert, pk }| {
+                        let cert = get_cert_inner(
```
arsing, line 438:
```suggestion
                let id_opt = auth.x509.as_ref()
                    .map(|x509| &x509.identity)
                    .and_then({
                        let homedir_path = &api.homedir_path;
                        let preloaded_certs = &api.preloaded_certs;
                        let key_client = &api.key_client;
                        let key_engine = &mut api.key_engine;
                        let id_opt = auth.x509.as_ref()
                        move |CertificateWithPrivateKey { cert, pk }| {
                            let cert = get_cert_inner(
```

This makes it clear that the bindings are for the sake of the closure, and also defines them in a smaller scope.

Eg https://github.com/Azure/iotedge/blob/cb33102e7ac5139bf8f016e95bcc93e511f1a86f/edgelet/edgelet-http/src/lib.rs#L270-L272

cert/aziot-certd/src/lib.rs
```diff
@@ -195,654 +242,356 @@ impl UpdateConfig for Api {
...
-                x509.set_not_after(&not_after)
-                    .map_err(|err| Error::Internal(InternalError::CreateCert(Box::new(err))))?;
+            let issuer_expiry = issuer_x509.not_after();
 
-                x509.set_issuer_name(subject_name)
-                    .map_err(|err| Error::Internal(InternalError::CreateCert(Box::new(err))))?;
+            ( issuer_x509.subject_name()
+            , if expiry < issuer_expiry { expiry } else { issuer_expiry }
+            , Some(issuer_x509_pem)
+            )
```
arsing, line 327:
```suggestion
            (
                issuer_x509.subject_name(),
                if expiry < issuer_expiry { expiry } else { issuer_expiry },
                Some(issuer_x509_pem),
            )
```

cert/aziot-certd/src/lib.rs
```diff
@@ -106,15 +123,46 @@ impl Api {
...
-            &csr,
-            issuer
-                .as_ref()
-                .map(|(issuer_cert, issuer_private_key)| (&**issuer_cert, issuer_private_key)),
-        )
-        .await?;
+        let req = X509Req::from_pem(&csr)
+            .or_else(|_| -> Result<_, BoxedError> {
+                let bytes = base64::decode(&csr)?;
+                Ok(X509Req::from_der(&bytes)?)
```
arsing, line 129: We do not support PKCS#10 requests in this API.

cert/aziot-certd/src/lib.rs
```diff
@@ -180,11 +228,10 @@ impl UpdateConfig for Api {
         // Don't allow changes to homedir path or endpoints while daemon is running.
         // Only update other fields.
         let Config {
-            homedir_path: _,
             cert_issuance,
             preloaded_certs,
-            endpoints: _,
             principal,
+            ..
```
arsing, line 234: The fields are explicitly listed intentionally, so that any refactoring of the config to add new fields has to be consciously handled here. That's why the aziotctl-config code also uses exhaustive patterns throughout, and why the comment on this line mentions why those fields are being ignored.

So please keep it as it was.

cert/aziot-certd/src/lib.rs
onalante-msft: Fixed in e6688907b6598cb9b4b2b38a4f3fd9ec86567e11.

cert/aziot-certd/src/lib.rs
onalante-msft: Fixed in e2d57edbe5898347cf7633f3a59c841737c9e963.

cert/aziot-certd/src/lib.rs
onalante-msft: Fixed in e2d57edbe5898347cf7633f3a59c841737c9e963.

cert/aziot-certd/src/lib.rs
onalante-msft: Fixed in e2d57edbe5898347cf7633f3a59c841737c9e963.

key/aziot-keys/aziot-keys.h
```diff
@@ -33,7 +33,7 @@ typedef unsigned int AZIOT_KEYS_VERSION;
 /**
  * The base struct of all of function lists.
  */
-typedef struct {
+typedef struct AZIOT_KEYS_FUNCTION_LIST {
```
arsing, line 36: These changes shouldn't be happening (in this PR). Make sure you're using the correct version of cbindgen. https://github.com/Azure/iot-identity-service/blob/9a12cd33ef27f82a5524c0d0d58938ec797fa852/ci/install-build-deps.sh#L157

cert/aziot-certd/src/lib.rs
```diff
@@ -195,654 +258,349 @@ impl UpdateConfig for Api {
...
+                                base64::encode(req.to_der()?).into_bytes(),
+                                url,
+                                auth.basic.as_ref(),
+                                id_opt.as_ref().map(|(cert, pk)| (&**cert, &**pk)),
+                                &trusted_certs,
+                                api.proxy_uri.clone()
+                            )
+                            .await,
+                    Err(err) => {
+                        log::warn!("{}", err);
```
arsing, line 486: This should be bubbling up the error like the old code did so that the context chain is preserved, rather than ad-hoc logging like this.

cert/aziot-certd/src/lib.rs
```diff
@@ -195,654 +258,349 @@ impl UpdateConfig for Api {
...
+                            )
+                            .await?;
+                        write_cert(
+                                &api.homedir_path,
+                                &api.preloaded_certs,
+                                &auth_x509.identity.cert,
+                                &id_init
+                            )?;
+
+                        create_cert_inner(api, id, (req, pubkey), issuer)
```
arsing, line 584: Restore the original comment that explains why this recursive call is happening.

cert/aziot-certd/src/lib.rs
```diff
@@ -195,654 +258,349 @@ impl UpdateConfig for Api {
...
+            )?;
+        let name_override = cert_options
+            .and_then(|opts| opts.subject.as_ref())
+            .map(build_name)
+            .transpose()?;
+
+        let subject_name = name_override.as_deref()
+            .unwrap_or_else(|| req.subject_name());
+
+        let issuer_privkey = api.key_engine
```
arsing, line 304: `s/privkey/pk/g` for consistency.

cert/aziot-certd/src/lib.rs
```diff
@@ -195,654 +258,349 @@ impl UpdateConfig for Api {
...
+
+                        let id_init = est::create_cert(
+                                base64::encode(builder.build().to_der()?).into_bytes(),
+                                url,
+                                auth.basic.as_ref(),
+                                Some((&bcert, &bpk)),
+                                &trusted_certs,
+                                api.proxy_uri.clone()
+                            )
+                            .await?;
```
arsing, line 576: Restore the error message from the original code. It contained both the error from getting the EST ID cert and the error from issuing a new one.

cert/aziot-certd/src/lib.rs
```diff
@@ -106,15 +123,61 @@ impl Api {
...
+
+                Ok((stack, handle))
+            })
+            .transpose()?;
+
+        let x509 = create_cert_inner(
+                &mut *this,
+                &id,
+                (&req, &pubkey),
+                issuer.as_ref().map(|(id, handle)| (&**id, &**handle))
```
arsing, line 170:
```suggestion
                issuer.as_ref().map(|(x509, pk)| (&**x509, &**pk))
```

cert/aziot-certd/src/lib.rs
```diff
@@ -106,15 +123,61 @@ impl Api {
...
+        let issuer = issuer
+            .map(|(id, handle)| -> Result<_, Error> {
+                let pem = get_cert_inner(
+                        &this.homedir_path,
+                        &this.preloaded_certs,
+                        &id
+                    )
+                    .map_err(|err| Error::Internal(InternalError::CreateCert(err.into())))?
+                    .ok_or_else(|| Error::invalid_parameter("issuer.certId", "not found"))?;
+                let stack = X509::stack_from_pem(&pem)
```
arsing, line 149: Same here, please use imperative code. There's no point in doing:

```rust
let stack = X509::stack_from_pem(&pem)
    .map_err(|err| Error::Internal(InternalError::CreateCert(err.into())))
    .and_then(|stack| {
        stack.get(0)
            .ok_or_else(||
                Error::invalid_parameter("issuer.certId", "invalid issuer")
            )?;
        Ok(stack)
    })?;
```

... when:

```rust
let stack = X509::stack_from_pem(&pem)
    .map_err(|err| Error::Internal(InternalError::CreateCert(err.into())))?;
let x509 = stack.get(0)
    .ok_or_else(|| Error::invalid_parameter("issuer.certId", "invalid issuer"))?;
```

... does the same thing with less complexity and less typing.

arsing (changes requested):

cert/aziot-certd/src/lib.rs
onalante-msft: Fixed in 56ac22b39cba8804ead30ca1d3031387e08366b9.

https://github.com/Azure/iot-identity-service/pull/296/commits/56ac22b39cba8804ead30ca1d3031387e08366b9#diff-ce1948fdb394a64e8a42f87d92e2057992c63a20520bad0c4c7359b64d0854c2R570-R578

cert/aziot-certd/src/lib.rs
onalante-msft: Fixed in 56ac22b39cba8804ead30ca1d3031387e08366b9.

key/aziot-key-client/src/lib.rs
```diff
@@ -535,7 +535,7 @@ fn try_parse_response(
         }
     }
 
-    if !is_json {
+    if !is_json && res_status_code != Some(204) && content_length.unwrap_or_default() == 0 {
```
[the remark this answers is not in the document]
onalante-msft: `&& content_length.unwrap_or_default == 0` is a wildly incorrect test, too. Thanks for pointing out the error. Fixed in 56ac22b39cba8804ead30ca1d3031387e08366b9.

cert/aziot-certd/src/lib.rs
```diff
@@ -195,654 +258,349 @@ impl UpdateConfig for Api {
...
-                        ) {
-                            Ok(Some(identity_cert)) => {
-                                match api.key_client.load_key_pair(identity_private_key) {
-                                    Ok(identity_private_key) => {
-                                        Ok((identity_cert, identity_private_key))
-                                    }
-                                    Err(err) => Err(format!(
+                                )?;
+                            let pk = key_client.load_key_pair(&pk)
+                                .map_err::<BoxedError, _>(Into::into)
```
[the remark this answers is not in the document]
onalante-msft: Fixed in 56ac22b39cba8804ead30ca1d3031387e08366b9.

cert/aziot-certd/src/lib.rs
onalante-msft: Addendum: the current code does not fall back to the bootstrap certificate when fetching with the identity fails, so the only kinds of errors that will be propagated will be identity fetching errors. However, this is the same behavior as the original code. Should we fall back to the bootstrap certificate when the identity certificate fails to produce a valid certificate (may be incorrect if bootstrap is single-use) and/or try re-enrolling?

gordonwang0 (approved):

cert/aziot-certd/src/lib.rs
```diff
@@ -195,654 +258,349 @@ impl UpdateConfig for Api {
...
+                                    )?;
+
+                                acc.extend(X509::stack_from_pem(&pem)?);
+                                Ok::<_, BoxedError>(acc)
+                            })
+                    )
+                    .transpose()?
+                    .unwrap_or_default();
+
+                let id_opt = auth.x509.as_ref()
```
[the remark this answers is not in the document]
onalante-msft: Discussed offline. This seems to be largely unavoidable.

Arnavion (approved):

onalante-msft:
Oh, sorry. My mistake.

PR #297, opened 2021-10-09
EST request and response fixes.

arsing:
Cherry-pick from main of d0a103564366e1496c9c08ec73a196b7e99a0f3f
with some rewriting since release/1.2's rustc is older and doesn't support
`<[T]>::strip_{prefix,suffix}`

- Strip PEM header and footer when creating a PKCS#10 request.

- When adding PEM footer to PKCS#7 response, ensure the footer goes on
  a new line.

- Fix est::create_cert to use trusted_certs for the EST server connection
  even when using basic auth.

  aa62e9cc61b907f886c2317a3784bf08187564ea fixed this bug in
  MaybeProxyConnector::new itself, but this code had the bug in the call to
  MaybeProxyConnector::new

### DIFF
--- a/cert/aziot-certd/src/est.rs
+++ b/cert/aziot-certd/src/est.rs
@@ -3,24 +3,26 @@
 use http_common::MaybeProxyConnector;
 
 pub(crate) async fn create_cert(
-    csr: Vec<u8>,
+    mut csr: &[u8],
     url: &url::Url,
     basic_auth: Option<(&str, &str)>,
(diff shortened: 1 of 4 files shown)


cert/aziot-certd/src/est.rs
```diff
@@ -3,24 +3,26 @@
 use http_common::MaybeProxyConnector;
 
 pub(crate) async fn create_cert(
-    csr: Vec<u8>,
+    mut csr: &[u8],
```
varunpuranik, line 6: Can we add some tests for this?

varunpuranik (approved):

PR #298, opened 2021-10-12
Expand build targets to include Debian 11 & RHEL 8

gauravIoTEdge:
Cherry-pick from master of:

- fce14de5cff7463f2dee552508081641831cb97e

  Expand build targets to Debian 11 and RHEL 8 (#273) 

- ef2d8c2e0deaf3352b9470c7086476b4f2fc364a

  E2E test script and workflow changes. (#294)

- a1087ff32e53410a1f9c99ae4c7ec6e05022a00a

   Rename contrib/centos directory to contrib/enterprise-linux since it's also used for EL8 now. (#295)

### DIFF
--- a/.github/workflows/e2e-tests-manual.yaml
+++ b/.github/workflows/e2e-tests-manual.yaml
@@ -64,14 +64,14 @@ jobs:
     steps:
     - uses: 'actions/checkout@v1'
       with:
-        ref: "${{ github.events.input.branch }}"
+        ref: "${{ github.event.inputs.branch }}"
 
     - name: 'Run'
       run: |
         timeout 15m ./ci/e2e-tests/test-run.sh '${{ matrix.test_name }}' &&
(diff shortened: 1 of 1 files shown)


arsing:
>Anyway, I was able to run E2E tests -- the symmetric key tests fail: https://github.com/gauravIoTEdge/iot-identity-service/actions/runs/1335210784

Yeah, because the name is too long, which is fine. There isn't anything about those tests that would be special.

The el8 tests also failed, because you didn't follow [these steps,](https://github.com/Azure/iot-identity-service/pull/298/files#diff-bbd0af96de04bdc05edf3899e5e329e72d2ede477b57317dc7e7b0d37dab38cbR141-R149) but that's also fine. The Debian 11 tests work and there isn't any other el8-specific change.

arsing (approved):

PR #299, opened 2021-10-13
`certd` asynchronous key client

onalante-msft:
`certd` currently uses the synchronous keyd client. This PR modifies it to use the asynchronous keyd client, and only initialize the synchronous client to pass to the key engine.

Depends on #296.


onalante-msft:
Closed in favor of #299.

PR #310, opened 2021-11-09
`certd` asynchronous key client

onalante-msft:
`certd` currently uses the synchronous `keyd` client. This PR modifies it to use the asynchronous `keyd` client, and only initialize the synchronous client to pass to the key engine.


arsing (approved):

arsing:
(You can rerun GH Actions from the web UI by clicking the "Rerun all jobs" button on the Details page of the run. You don't have to push an empty commit.)

PR #315, opened 2021-11-12
Add mock-dps-server and mock-dps-provision test

gordonwang0:
Adds a test program, `mock-dps-server`, that can mock a subset of DPS functionality. Also adds the `mock-dps-provision` test, which provisions using the mocked DPS server. This new test runs as part of the commit CI.


ci/mock-dps-tests/mock-dps-provision.sh
```diff
@@ -0,0 +1,220 @@
...
+openssl genrsa -out device_id_certkey.pem
+openssl req -new -key device_id_certkey.pem -subj "/CN=mock-dps-provision" -out device_id_certreq.pem
+
+openssl x509 -req \
+    -in device_id_certreq.pem \
+    -extfile device_id_cert.conf -extensions device_id_cert \
+    -signkey device_id_certkey.pem -sha256 \
+    -out device_id_cert.pem
+
+>/etc/aziot/keyd/config.toml cat <<-EOF
```
arsing, line 52: Can you do these with `aziotctl config apply` instead? Would be nice to test it too.

ci/mock-dps-tests/mock-dps-provision.sh
```diff
@@ -0,0 +1,220 @@
...
+[aziot_keys]
+homedir_path = "/var/lib/aziot/keyd"
+
+[preloaded_keys]
+device-id-symkey = "file://$(pwd)/device_id_symkey"
```
arsing, line 57: `$(pwd)` -> `$PWD`

identity/mock-dps-server/src/main.rs
```diff
@@ -0,0 +1,76 @@
...
+
+#![deny(rust_2018_idioms)]
+#![warn(clippy::all, clippy::pedantic)]
+
+mod request;
+mod server;
+
+#[tokio::main]
+async fn main() {
+    let matches = clap::App::new("mock-dps-server")
```
arsing, line 11: Use structopt, like we do for the other binaries.

identity/mock-dps-server/src/request.rs
```diff
@@ -0,0 +1,198 @@
...
+
+    let value = percent_encoding::percent_decode_str(value)
+        .decode_utf8()
+        .map_err(|_| Response::bad_request(format!("bad {}", name)))?
+        .to_string();
+
+    Ok(value)
+}
+
+pub(crate) fn process_dps_request(
```
arsing, line 129: Would it make sense to add TLS support to `http_common::Connector` instead of making an ad-hoc server like this? Or is this the only mock server we'll need?

identity/mock-dps-server/src/request.rs
```diff
@@ -0,0 +1,198 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+use crate::server::Response;
+
+trait DpsRequest {
```
arsing, line 5: You always call `.process()` on the concrete type, so it doesn't seem necessary to have a trait.

Even the three concrete types can just be three functions.

key/aziot-key-test-common/Cargo.toml
```diff
@@ -0,0 +1,14 @@
+[package]
+name = "aziot-key-test-common"
```
arsing, line 2: It isn't really aziot-key-specific. You could just call it test-common and put it in the root of the repo.

arsing (changes requested):

key/aziot-key-test-common/Cargo.toml
gordonwang0: Fixed

identity/mock-dps-server/src/request.rs
gordonwang0: Fixed

ci/mock-dps-tests/mock-dps-provision.sh
gordonwang0: Fixed

identity/mock-dps-server/src/main.rs
gordonwang0: Fixed

ci/mock-dps-tests/mock-dps-provision.sh
gordonwang0: I tried that first, but aziotctl expects a package installation (it checks for the aziotks/aziotid/aziotcs users, uses systemctl restart, etc.). I figured this way was simpler.

identity/mock-dps-server/src/request.rs
gordonwang0: I think this is the only mock server we need. We probably don't need to mock IoT Hub since IS can function offline.

We can always add it later if needed.

ci/mock-dps-tests/mock-dps-provision.sh
arsing: Ah, right.

arsing (approved):

PR #316, opened 2021-11-12
Specify API version for az resource tag in E2E

gordonwang0:
A bug in the latest az CLI causes az resource tag to use an API version not supported by the cloud. For now, we fill manually specify the API version.

### DIFF
--- a/ci/e2e-tests/suite-cleanup.sh
+++ b/ci/e2e-tests/suite-cleanup.sh
@@ -57,8 +57,8 @@ while :; do
         fi
     done
 
-    <<< "$dps_ids" timeout 30s xargs az resource delete --api-version 2020-03-01 --ids >/dev/null
-    <<< "$other_ids" timeout 30s xargs az resource delete --ids >/dev/null
+    <<< "$dps_ids" timeout 30s xargs -r az resource delete --api-version 2020-03-01 --ids >/dev/null
+    <<< "$other_ids" timeout 30s xargs -r az resource delete --ids >/dev/null
 
     sleep 1
     echo 'Retrying...' >&2
--- a/ci/e2e-tests/suite-cleanup.sh
+++ b/ci/e2e-tests/suite-cleanup.sh
@@ -46,14 +46,19 @@ while :; do
     # Ref: https://github.com/Azure/azure-cli/issues/20263
     readarray -t ids <<< "$ids"
 
+    dps_ids=""
+    other_ids=""
     for resource in "${ids[@]}"
     do
         if [[ "$resource" == *"/Microsoft.Devices/ProvisioningServices/"* ]]; then
-            timeout 30s az resource delete --api-version 2020-03-01 --ids "$resource" >/dev/null
+            dps_ids+="$resource"$'\n'
         else
-            timeout 30s az resource delete --ids "$resource" >/dev/null
+            other_ids+="$resource"$'\n'
         fi
     done
+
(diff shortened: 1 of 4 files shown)


ci/e2e-tests/suite-setup.sh
```diff
@@ -80,8 +80,13 @@ az iot dps linked-hub create \
 # `az iot dps linked-hub create` deletes the tags on the DPS that
 # were set by `az iot dps create` for some unknown reason, so we
 # need to tag it again.
+#
+# A bug in the latest az CLI causes az resource tag to use an API
+# version not supported by the cloud. For now, we fill manually
+# specify the API version.
```
arsing, line 86: Add the link to the bug too ( https://github.com/Azure/azure-cli/issues/20263 ) so we can check on it later and remove this.

Also, this is the only command that needs this workaround?

arsing (changes requested):

ci/e2e-tests/suite-cleanup.sh
```diff
@@ -39,7 +39,21 @@ while :; do
...
 
-    <<< "$ids" timeout 30s xargs az resource delete --ids >/dev/null
+    # A bug in the latest az CLI causes az resource delete to use an API
+    # version not supported by the cloud for DPS instances. For now, we
+    # will manually specify the API version when deleting DPS instances.
+    #
+    # Ref: https://github.com/Azure/azure-cli/issues/20263
+    readarray -t ids <<< "$ids"
+
+    for resource in "${ids[@]}"
```
arsing, line 49: This will be many times slower because it'll delete resources one at a time.

What I'd suggested was to filter out the DPS ID in one place, run `az resource delete` on all the other resources first, and then on the DPS ID with the API version parameter.

gordonwang0: Fixed

ci/e2e-tests/suite-cleanup.sh
```diff
@@ -46,15 +46,20 @@ while :; do
...
-            timeout 30s az resource delete --api-version 2020-03-01 --ids "$resource" >/dev/null
+            dps_ids+="$resource"$'\n'
         else
-            timeout 30s az resource delete --ids "$resource" >/dev/null
+            other_ids+="$resource"$'\n'
         fi
     done
 
+    <<< "$dps_ids" timeout 30s xargs az resource delete --api-version 2020-03-01 --ids >/dev/null
+    <<< "$other_ids" timeout 30s xargs az resource delete --ids >/dev/null
```
arsing, line 61: `xargs` -> `xargs -r` in both lines, in case the input is empty.

ci/e2e-tests/suite-cleanup.sh
arsing: Thanks. Is the timeout for the suite-cleanup stages sufficient? Or do you need to double it like I'd said on Teams?

arsing (changes requested):

ci/e2e-tests/suite-cleanup.sh
gordonwang0: Fixed

ci/e2e-tests/suite-cleanup.sh
gordonwang0: Yes, it's sufficient. I haven't noticed it timing out.

arsing (approved):

PR #319, opened 2021-11-15
[release/1.2] Emit base64-encoded DER in 64-character chunks

onalante-msft:
Cisco's libest implementation does not accept non-chunked inputs, and
non-chunked base64 encoding. Though libest is not advertised as an EST
server suitable for production, it is still popular enough to warrant
accommodation by our client. Since it seems that EST servers that accept
unchunked input also accept chunked input, it should be safe (with
respect to compatibility with other servers) to make this change.


onalante-msft:
Wrong branch, disregard.

PR #320, opened 2021-11-15
[release/1.2] Emit base64-encoded DER in 64-character chunks

onalante-msft:
*Cf.* #323.


onalante-msft:
@arsing This should be ready to go given that my account is back in working condition.

arsing (approved):

ksaye:
After a little testing, this works with the Cisco EST server.

PR #322, opened 2021-11-16
Add CI tests for DPS trust bundle

gordonwang0:


ci/mock-dps-tests/mock-dps-trust-bundle.sh
```diff
@@ -0,0 +1,126 @@
...
+server_pid="$!"
+sleep 1
+
+# Reprovision and check that the trust bundle was deleted.
+curl -s --unix-socket /run/aziot/identityd.sock "http://localhost/identities/device/reprovision?api-version=2021-12-01" \
+    -H "content-type: application/json" --data '{"type": "aziot"}' &> /dev/null
+
+message=$(curl -s --unix-socket /run/aziot/certd.sock "http://localhost/certificates/aziot-dps-trust-bundle?api-version=2020-09-01" | jq -r .message)
+
+if [ "$(echo $message)" != 'parameter "id" has an invalid value caused by: not found' ]; then
```
gordonwang0, line 103: @arsing I'm not sure why I had to use `echo` here? i.e. `if [ "$message" != 'parameter "id" has an invalid value caused by: not found' ]; ` doesn't work.

arsing: `echo $message` will normalize all whitespace to one space and strip leading / terminating whitespace. I assume the actual `.message` contains a newline between `value caused` instead of a space.

gordonwang0: Got it. I'll leave it as is then.

PR #330, opened 2021-11-22
E2E scheduled run fixes involving EL8.

arsing:
Cherry-pick from main of:

- 412eef3741ec550ba3b164683e16650538cf1685

  Increase timeout for e2e-tests/test-run.sh for the sake of AlmaLinux. (#329)

  AlmaLinux seems to take 15 minutes just to install updates.

- 3f6b402a71757869af81b8abea6ea2362e30add8

  Disable EL8 in E2E tests. (#331)

  412eef3741ec550ba3b164683e16650538cf1685 tried to fix the EL8 failures from it
  taking too long to install packages, but EL8 tests still fail because
  the Azure VM spontaneously lose ssh access after installing updates for
  some reason. Serial console doesn't reveal why.

  So this change disables EL8 runs for the sake of getting E2E tests passing,
  until we get around to investigating the VM issue.

### DIFF
--- a/.github/workflows/e2e-tests-manual.yaml
+++ b/.github/workflows/e2e-tests-manual.yaml
@@ -68,7 +68,7 @@ jobs:
 
     - name: 'Run'
       run: |
-        timeout 15m ./ci/e2e-tests/test-run.sh '${{ matrix.test_name }}' &&
+        timeout 30m ./ci/e2e-tests/test-run.sh '${{ matrix.test_name }}' &&
         timeout 5m ./ci/e2e-tests/test-cleanup.sh '${{ matrix.test_name }}'
       env:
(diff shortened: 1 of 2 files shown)


varunpuranik (approved):
Change looks good.. but let's first confirm it fixes the issue in main.

arsing:
Okay, I can kick off a manual run in main.

arsing:
Bah, now the VM never comes back after it's rebooted. It's the same problem we had (and I've personally had) with CentOS VMs on Azure in the past. Serial console also doesn't come back after the reboot so there's no way to know what's wrong.

I think I'll just remove the `reboot` step for these distros.

arsing:
Updated this PR to have both the timeout bump change and the change to disable EL8 in the scheduled tests workflow.

PR #332, opened 2021-11-30
Add Timeout to TLS Handshake during IoT Hub Connect Check

nimanch:
The reason for doing this that behind a proxy, the IoT Edge Check takes a long time to run if specific ports are blocked. This causes Support Bundle Operation to time out.

### DIFF
--- a/aziotctl/src/internal/check/checks/host_connect_iothub.rs
+++ b/aziotctl/src/internal/check/checks/host_connect_iothub.rs
@@ -2,7 +2,7 @@
 
 use anyhow::{Context, Result};
 use serde::Serialize;
-use tokio::time::timeout;
+use tokio::time::{timeout, Duration};
 
 use crate::internal::check::{CheckResult, Checker, CheckerCache, CheckerMeta, CheckerShared};
 
@@ -143,13 +143,20 @@ impl HostConnectIotHub {
             None
         };
 
+        // TLS Handshake behind Proxy takes a long time if the ports are blocked. This causes support bundle operation to fail. Add a timeout to the call.
         let future = crate::internal::common::resolve_and_tls_handshake(
             iothub_hostname_url,
             iothub_hostname,
             proxy,
         );
 
-        timeout(std::time::Duration::new(30, 0), future).await.map_err(|e| anyhow::Error::from(e))?.map_err( |e| if nested {
+        timeout(Duration::from_secs(identityd_config.cloud_timeout_sec), future).await.map_err(|e| {
+            let context = format!(
+                "Failed to do TLS Handshake, Connection Attempt Timed out in {} Seconds",
+                identityd_config.cloud_timeout_sec,
+            );
+            anyhow::Error::from(e).context(context)
+            })?.map_err( |e| if nested {
             e.context("Make sure the parent device is reachable using 'curl https://parenthostname'. Make sure the the trust bundle has been added to the trusted store: sudo cp <path>/azure-iot-test-only.root.ca.cert.pem /usr/local/share/ca-certificates/azure-iot-test-only.root.ca.cert.pem.crt
             sudo update-ca-certificates")
         } else {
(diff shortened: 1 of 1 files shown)


aziotctl/src/internal/check/checks/host_connect_iothub.rs
```diff
@@ -142,12 +143,13 @@ impl HostConnectIotHub {
...
-        crate::internal::common::resolve_and_tls_handshake(
+        let future = crate::internal::common::resolve_and_tls_handshake(
             iothub_hostname_url,
             iothub_hostname,
             proxy,
-        )
-        .await.map_err( |e| if nested {
+        );
+
+        timeout(std::time::Duration::new(30, 0), future).await.map_err(|e| anyhow::Error::from(e))?.map_err( |e| if nested {
```
arsing, line 152: `Duration::from_secs(30)`

nimanch: Done

aziotctl/src/internal/check/checks/host_connect_iothub.rs
```diff
@@ -159,3 +166,81 @@ impl HostConnectIotHub {
...
+    use aziot_identityd_config::{
+        Endpoints, ManualAuthMethod, Provisioning, ProvisioningType, Settings,
+    };
+    use std::path::PathBuf;
+
+    #[tokio::test]
+    async fn tls_handshake_timesout() {
+        let mut host = HostConnectIotHub::new(
+            "hostConnectIoTHubMQTT",
+            "
```
arsing, line 184: Remove this newline inside the string literal.

aziotctl/src/internal/check/checks/host_connect_iothub.rs
```diff
@@ -159,3 +166,81 @@ impl HostConnectIotHub {
...
+    use crate::internal::check::{CheckerCache, CheckerCfg, CheckerShared, DaemonConfigs};
+
+    use super::HostConnectIotHub;
+    use aziot_identityd_config::{
+        Endpoints, ManualAuthMethod, Provisioning, ProvisioningType, Settings,
+    };
+    use std::path::PathBuf;
+
+    #[tokio::test]
+    async fn tls_handshake_timesout() {
```
arsing, line 181: Please don't make a test that stalls for 10s. It breaks the dev workflow.

arsing (changes requested):

aziotctl/src/internal/check/checks/host_connect_iothub.rs
nimanch: I have set the cloud timeout as 1 second for this particular test

nimanch: Compiling aziotctl v1.3.0 (/workspaces/iot-identity-service/aziotctl)
    Finished test [unoptimized + debuginfo] target(s) in 13.28s
     Running unittests (/workspaces/iot-identity-service/target/debug/deps/aziotctl-833ade0ec64079e5)

running 1 test
test internal::check::checks::host_connect_iothub::tests::tls_handshake_timesout ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.00s

arsing: Ah right. Sorry.

I would also like it to not make a network call and expect it to fail. We do have precedence for it with the mini-sntp tests, but at least that's legitimate traffic to NTP servers, unlike connecting to bing.com on port 8883. In fact, it's lucky that (presumably) Bing happens to black-hole the traffic and thus your tests time out, as opposed to it rejecting the connection which would result in a different error. There's no guarantee it couldn't change in the future.

I suggest spinning up a bespoke TCP server (just a spawned task that does `tokio::TcpListener::bind(("::", 0))` + `listener.accept()` + `tokio::time::sleep(5)`) and connecting to that (`"localhost"` for the hostname and `listener.local_addr().port()` for the port).

nimanch: @arsing  : Thanks for the suggestion!, I ended up using just binding to a Port without having to accept connections to keep things simple. I observed that I still get a timeout without having to do ```listener.accept()```. Let me know what you think?

aziotctl/src/internal/check/checks/host_connect_iothub.rs
nimanch: Done

aziotctl/src/internal/check/checks/host_connect_iothub.rs
arsing: Oh right, that'll work too, and yeah that's much easier than spawning a sleeping task.

aziotctl/src/internal/check/checks/host_connect_iothub.rs
```diff
@@ -157,3 +166,85 @@ impl HostConnectIotHub {
         Ok(CheckResult::Ok)
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use crate::internal::check::{CheckerCache, CheckerCfg, CheckerShared, DaemonConfigs};
```
arsing, line 172: Imports should be in order of std imports, then third-party crate imports (tokio), then first-party crate imports (aziot_identityd_config), then local crate imports (crate, super), with an empty line between each group.

arsing (approved):

gordonwang0 (approved):

aziotctl/src/internal/check/checks/host_connect_iothub.rs
arsing: @nimanch This was not resolved?

nimanch: @arsing : I may have missed this comment. Sorry, I will send out a new PR for this.

PR #335, opened 2021-12-07
Update toolchain to 1.57 and 2021 edition

vadim-kovalyov:
Update toolchain to 1.57. and rust edition to 2021.

### DIFF
--- a/aziotctl/Cargo.toml
+++ b/aziotctl/Cargo.toml
@@ -2,7 +2,7 @@
 name = "aziotctl"
 version = "1.3.0"
 authors = ["Azure IoT Edge Devs"]
-edition = "2018"
+edition = "2021"
 
 
 [dependencies]
--- a/aziotctl/aziotctl-common/Cargo.toml
+++ b/aziotctl/aziotctl-common/Cargo.toml
@@ -2,7 +2,7 @@
 name = "aziotctl-common"
 version = "0.1.0"
 authors = ["Azure IoT Edge Devs"]
-edition = "2018"
+edition = "2021"
 
 
 [dependencies]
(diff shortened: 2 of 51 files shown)


arsing:
If you're going to update it, you might as well update it to latest, including switching to E2021 and making clippy fixes? Why bump the toolchain just to have the *option* of using E2021, instead of actually using it?

vadim-kovalyov:
> If you're going to update it, you might as well update it to latest, including switching to E2021 and making clippy fixes? Why bump the toolchain just to have the _option_ of using E2021, instead of actually using it?

Sure, I tried that and, surprisingly, there were not a lot of errors. Since I don't have much context on this repo - the three errors I saw I fixed with `allow` attributes. @arsing please suggest better fix if needed, and I'll do that.

arsing:
The `large_enum_variant` one is fine. We have a bunch of them already.

Is the `dead_code` one because most of the fields are not used? I guess they made `#[derive(Debug)]` not count as a use any more unless Debug formatting is actually used (which is sensible).

So rather than `#[allow(dead_code)]`'ing it, I think it would be better to do:

```diff,rust
     let packet = Packet::parse(buf, sntp_epoch);
-    #[cfg(test)]
-    let packet = dbg!(packet);
+    log::debug!("Received SNTP packet {:?}", packet);

```

vadim-kovalyov:
> ```diff
> log::debug!("Received SNTP packet {:?}", packet);
> ```

That did not help, unfortunately, unless I manually implement `Debug` (which does not make sense IMHO).

arsing:
Strange. Let's deal with it later then.

arsing (approved):

arsing:
>Strange. Let's deal with it later then.

Reason for the error: https://github.com/rust-lang/rust/issues/88900

An annoying change. So until they give a better way to opt out, `#[allow(dead_code)]` on each affected struct seems fine to me.

PR #336, opened 2021-12-15
Ignore invalid backups when provisioning Identity Service

gordonwang0:
Identity Service keeps a backup of the previous run's device info and settings so that it can maintain state between restarts. Previously, if this backup was using an outdated format, corrupted, or otherwise invalid, Identity Service would fail to start.

This change causes Identity Service to ignore invalid backup files and provision without its previous state.

### DIFF
--- a/identity/aziot-identityd/src/identity.rs
+++ b/identity/aziot-identityd/src/identity.rs
@@ -742,7 +742,7 @@ impl IdentityManager {
         credentials: aziot_identity_common::Credentials,
         local_gateway_hostname: Option<String>,
     ) -> Result<aziot_identity_common::IoTHubDevice, Error> {
-        let backup_device = self.get_backup_provisioning_info(credentials.clone())?;
+        let backup_device = self.get_backup_provisioning_info(credentials.clone());
 
         if skip_if_backup_is_valid && backup_device.is_some() {
             let backup_device = backup_device.expect("backup device cannot be none");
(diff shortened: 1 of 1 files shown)


identity/aziot-identityd/src/identity.rs
```diff
@@ -1002,6 +996,45 @@ impl IdentityManager {
...
+        Err(err) => {
+            log::warn!("Ignoring invalid device info backup: {}", err);
+
+            return std::collections::BTreeSet::default();
+        }
+    };
+
+    // Only consider the previous Hub modules if the current and previous Hub devices match.
+    if prev_hub_device_info != Some(curr_hub_device_info) {
+        return std::collections::BTreeSet::default();
```
varunpuranik, line 1019: Should we clear the prev hub device info?

identity/aziot-identityd/src/identity.rs
```diff
@@ -1002,6 +996,45 @@ impl IdentityManager {
...
+    }
+
+    let prev_settings =
+        match crate::configext::load_file(prev_settings_path).map_err(Error::Internal) {
+            Ok(settings) => settings,
+            Err(err) => {
+                log::warn!("Ignoring invalid device settings backup: {}", err);
```
varunpuranik, line 1026: here as well?

varunpuranik (approved):

identity/aziot-identityd/src/identity.rs
gordonwang0: It's overwritten by the caller of this function (line 986) so no need to clear it here.

identity/aziot-identityd/src/identity.rs
gordonwang0: See above.

identity/aziot-identityd/src/identity.rs
```diff
@@ -1002,6 +996,45 @@ impl IdentityManager {
     }
 }
 
+fn get_prev_modules(
+    prev_settings_path: &std::path::Path,
+    prev_device_info_path: &std::path::Path,
+    curr_hub_device_info: HubDeviceInfo,
+) -> std::collections::BTreeSet<aziot_identity_common::ModuleId> {
+    if !prev_settings_path.exists() || !prev_device_info_path.exists() {
+        return std::collections::BTreeSet::default();
```
arsing, line 1005: `return Default::default();`

arsing (approved):

varunpuranik (approved):

identity/aziot-identityd/src/identity.rs
gordonwang0: Fixed

Issue #233, opened 2021-06-03
Buffering issues in log_msg() of tpm/aziot-tpm-sys/azure-iot-hsm-c


theKidOfArcrania:
Hi. Whenever a message (error/info/dbg) gets [logged in the azure-iot-hsm-c library](https://github.com/Azure/iot-identity-service/blob/main/tpm/aziot-tpm-sys/azure-iot-hsm-c/src/hsm_log.c#L39) it uses the `printf` function, which if stdout is set to fully buffered (which would be the case when stdout/stderr gets piped to syslog), the individual log entries do not get written out until it fills up the buffer (which would not be optimal in the case of logging messages and expecting it to be immediately written to a log file/syslog).

A few possible fixes might include, either using `fprintf(stderr, ...)` instead (which I believe is usually always unbuffered), or invoking `fflush(stdout);` after the printf, in order to ensure the log entry gets written out immediately.

arsing:
Yes, it's been on the back of everyone's minds to fix this since back when this code belonged in the iotedge repo. We realized the problem when we saw that the hsm-c logs were always printed in one big dump at the end when the process exited.

And yes, the fix should be to write them to stderr because that's where logs out to go, not to flush stdout.

arsing:
#338 fixes part of the logs to go to stderr, but logs from within the C SDK still go to stdout and are out of our control.

arsing:
The C SDK is no longer used as of 143cdf79be08e7069a6a0b17b38f33469c291b1e

PR #338, opened 2021-12-21
Fix aziot-tpm-sys's C code to log to stderr instead of stdout.

arsing:
This means logs are no longer buffered until some arbitrary point in the future
(such as when the process exits). Instead they are immediately printed.

Unfortunately this does not account for logs from the C SDK dependencies,
ie utpm and c-shared, since they use their own logging macro which also writes
using `printf`. So, for example, when running tpmd without a reachable TPM
prints this:

    <6>2021-12-21T00:24:30Z [INFO] (/home/arsing/src/iot-identity-service/tpm/aziot-tpm-sys/azure-iot-hsm-c/src/hsm_log.c:log_init:20) Initialized logging
    Error: Time:Mon Dec 20 16:24:30 2021 File:/home/arsing/src/iot-identity-service/tpm/aziot-tpm-sys/azure-iot-hsm-c/deps/utpm/src/tpm_comm_linux.c Func:tpm_usermode_resmgr_connect Line:268 Failure: No user mode TRM found.
    Error: Time:Mon Dec 20 16:24:30 2021 File:/home/arsing/src/iot-identity-service/tpm/aziot-tpm-sys/azure-iot-hsm-c/deps/utpm/src/tpm_comm_linux.c Func:tpm_comm_create Line:335 Failure: connecting to the TPM device
    Error: Time:Mon Dec 20 16:24:30 2021 File:/home/arsing/src/iot-identity-service/tpm/aziot-tpm-sys/azure-iot-hsm-c/deps/utpm/src/tpm_codec.c Func:Initialize_TPM_Codec Line:256 creating tpm_comm object
    <3>2021-12-21T00:24:30Z [ERR!] (/home/arsing/src/iot-identity-service/tpm/aziot-tpm-sys/azure-iot-hsm-c/src/hsm_client_tpm_device.c:initialize_tpm_device:278) Failure initializeing TPM Codec
    <3>2021-12-21T00:24:30Z [ERR!] (/home/arsing/src/iot-identity-service/tpm/aziot-tpm-sys/azure-iot-hsm-c/src/hsm_client_tpm_device.c:hsm_client_tpm_create:311) Failure initializing tpm device.

... the "ERR!" logs are now printed to stderr, but the "Error:" logs are still
printed to stdout. Unfortunately the latter are still needed for fully
diagnosing issues.

Nevertheless, we'll at least know the aziot-tpm-sys part of the problem
more easily.

Other changes:

- Write the output directly instead of building up a string buffer first.

- Write with Unix newlines instead of Windows newlines.

Ref #233

### DIFF
--- a/tpm/aziot-tpm-sys/azure-iot-hsm-c/src/hsm_log.c
+++ b/tpm/aziot-tpm-sys/azure-iot-hsm-c/src/hsm_log.c
@@ -5,8 +5,6 @@
 
 #include "hsm_log.h"
 
-#define MAX_LOG_SIZE 256
-
 static bool g_is_log_initialized = false;
 
 static int log_level = LVL_INFO;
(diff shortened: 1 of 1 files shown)


gordonwang0 (approved):

tpm/aziot-tpm-sys/azure-iot-hsm-c/src/hsm_log.c
```diff
@@ -25,18 +23,20 @@ void log_init(int level) {
 
 void log_msg(int level, const char* file, const char* function, int line, const char* fmt_str, ...) {
     static char levels[3][5] = {"DBUG", "INFO", "ERR!"};
```
gordonwang0, line 25: This 2d array has a 1d initializer and is only ever accessed as a 1d array?

gordonwang0, line 25: Doesn't this only need to be [3][4]?

PR #339, opened 2021-12-27
Code coverage

brotherneeru:
This PR is for adding code coverage to the iot-identity-service,

TODO:
- Add coverage threshold for failure
- automate commenting on PRs the coverage level


Makefile
```diff
@@ -207,6 +207,18 @@ test:
 		done
 
 
+codecov: default
+	mkdir -p coverage
+	
+	$(INSTALL_PROGRAM) -D target/$(CARGO_TARGET)/$(CARGO_PROFILE_DIRECTORY)/libaziot_keys.so $(DESTDIR)$(libdir)/libaziot_keys.so
```
arsing, line 213: Don't do this. You're not running in an install scenario, so DESTDIR is empty, which means you're installing to the real `/usr/lib`.

Add `target/$(CARGO_TARGET)/$(CARGO_PROFILE_DIRECTORY)/` to `LD_LIBRARY_PATH` for the cargo-tarpaulin invocation instead.

```diff
-$(CARGO) tarpaulin --all --verbose \
+LD_LIBRARY_PATH="$$LD_LIBRARY_PATH:$$PWD/target/$(CARGO_TARGET)/$(CARGO_PROFILE_DIRECTORY)" $(CARGO) tarpaulin --all --verbose \
```

arsing (changes requested):

Makefile
brotherneeru: Updated PR to address this

nyanzebra (approved):
Get @arsing approval too, but this should work for what we need as far as I can tell

arsing (approved):

gordonwang0 (approved):

PR #340, opened 2021-12-29
Ensure parent directories are readable by target user when creating directories in `config apply`

arsing:
When `config apply` creates the /var/secrets/aziot/keyd directory to store
the inline symmetric key, it assumed that parent directories would already
be readable by the target user. But this is not a valid assumption since
it depends on root's umask, and we've had a few cases of devices where
root's umask is such that new directories are 0700 or 0750 by default.

This change adds a step of walking through the parents,
checking if they're readable by the target user, and making them
world-readable if they're not.

### DIFF
--- a/aziotctl/src/internal/check/checks/read_key_pairs.rs
+++ b/aziotctl/src/internal/check/checks/read_key_pairs.rs
@@ -62,61 +62,79 @@ impl ReadKeyPairs {
         let mut err_aggregated = vec![];
         let mut warn_aggregated = vec![];
 
-        for id in preloaded_keys.keys() {
-            // We don't know whether `id` is a symmetric or asymmetric key,
-            // and the `load_key_pair` error doesn't tell us whether it failed because it's a symmetric key
-            // or because of some other reason.
-            //
-            // We also can't go behind keyd's back and load the keys ourselves, because the point of this check
-            // is to test keyd's ability to load the keys, but also because only keyd is allowed to load PKCS#11 keys
-            // since PKCS#11 implementations are generally not cross-process-safe.
-            //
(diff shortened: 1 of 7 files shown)


gordonwang0:
Can we also add a check for this in the read_key_pairs check, since it's possible to modify the directories' permissions after creation?

gordonwang0 (approved):

arsing:
>Can we also add a check for this in the read_key_pairs check, since it's possible to modify the directories' permissions after creation?

Done, and for certs checks too.

aziotctl/src/internal/check/checks/read_key_pairs.rs
```diff
@@ -62,61 +62,79 @@ impl ReadKeyPairs {
...
+        let aziotks_user = crate::internal::common::get_system_user("aziotks")?;
+
+        for (id, path) in preloaded_keys {
+            let mut readable = true;
+            if let Ok(aziot_keys_common::PreloadedKeyLocation::Filesystem { path }) = path.parse() {
+                if let Err(err) =
+                    aziotctl_common::config::check_readable(&path, &aziotks_user, false)
+                {
+                    err_aggregated.push(format!("{:?}", err));
+                    readable = false;
```
gordonwang0, line 75: I don't think it's necessary to keep a `readable` flag? You can just `continue` when not readable.

gordonwang0 (approved):

aziotctl/src/internal/check/checks/read_key_pairs.rs
arsing: Ah yes. It was necessary when I wrote it, but then I refactored it a little and now it's not necessary. Removed.

PR #341, opened 2021-12-29
Fix keyd documentation to not suggest /decrypt supports keypair algorithms.

### LINKED ISSUE #311: Decrypt not supported for AZIOT_KEYS_ENCRYPT_MECHANISM_RSA_PKCS1
It seems that [decrypt](https://github.com/Azure/iot-identity-service/blob/febb3e740fa4aef4f84d9b3042ca426a7d2b8cf2/key/aziot-keyd/src/lib.rs#L419) does not support RSA-PKCS1 (AZIOT_KEYS_ENCRYPT_MECHANISM_RSA_PKCS1)

while encrypt does: https://github.com/Azure/iot-identity-service/blob/febb3e740fa4aef4f84d9b3042ca426a7d2b8cf2/key/aziot-keyd/src/lib.rs#L354 

what is the rationale behind this? I am not able to decrypt what I encrypted or do I miss something?

Thanks
Vitaliy

arsing:
Fixes #311

### DIFF
--- a/docs/api/keys-service.md
+++ b/docs/api/keys-service.md
@@ -350,26 +350,6 @@ Only valid for symmetric keys.
 }
 ```
 
-##### RSA-PKCS1
-
-```json
-{
-    "keyHandle": "string",
-    "algorithm": "RSA-PKCS1",
-    "ciphertext": "base64-encoded-string"
-}
-```
-
-##### RSA-NO-PADDING
-
-```json
-{
-    "keyHandle": "string",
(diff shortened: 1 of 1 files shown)


varunpuranik (approved):

varunpuranik:
@arsing - it will be good to have a sample on how to use the APIs for encryption/decryption for diff types of algorithms. Can you pls add an item in our backlog to do that?

gordonwang0 (approved):

PR #342, opened 2022-01-04
Move system information gathering to common crate

onalante-msft:
`edgelet` uses the `OsInfo`, `DmiInfo`, and `parse_shell_line` functions.

### DIFF
--- a/aziotctl/Cargo.toml
+++ b/aziotctl/Cargo.toml
@@ -18,7 +18,7 @@ foreign-types-shared = "0.1"
 hyper = "0.14"
 hyper-openssl = "0.9"
 libc = "0.2"
-nix = "0.18"
+nix = "0.22"
 log = "0.4"
 openssl = "0.10"
 serde = { version = "1", features = ["derive"] }
--- a/aziotctl/Cargo.toml
+++ b/aziotctl/Cargo.toml
@@ -18,7 +18,7 @@ foreign-types-shared = "0.1"
 hyper = "0.14"
 hyper-openssl = "0.9"
 libc = "0.2"
-nix = "0.22"
+nix = "0.23"
 log = "0.4"
 openssl = "0.10"
 serde = { version = "1", features = ["derive"] }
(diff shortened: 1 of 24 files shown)


aziotctl/aziotctl-common/src/host_info.rs
```diff
@@ -0,0 +1,145 @@
...
+
+    Some((key, value))
+}
+
+fn try_read_dmi(entry: &'static str) -> Option<String> {
+    let path = format!("/sys/devices/virtual/dmi/id/{}", entry);
+
+    let bytes = fs::read(path).ok()?;
+
+    Some(String::from_utf8(bytes)
```
arsing, line 141: If you're `to_owned()`'ing at the end anyway, you might as well do `str::from_utf8()` and save on the first copy.

aziotctl/Cargo.toml
```diff
@@ -18,7 +18,7 @@ foreign-types-shared = "0.1"
 hyper = "0.14"
 hyper-openssl = "0.9"
 libc = "0.2"
-nix = "0.18"
+nix = "0.22"
```
arsing, line 21: Why is this needed?

And you might as well update to 0.23 then (the current latest).

arsing (approved):

aziotctl/aziotctl-common/src/host_info.rs
```diff
@@ -0,0 +1,145 @@
...
+pub fn parse_shell_line(line: &str) -> Option<(&str, &str)> {
+    let line = line.trim();
+
+    let (key, value) = line.split_once('=')?;
+
+    // The value is essentially a shell string, so it can be quoted in single or
+    // double quotes, and can have escaped sequences using backslash.
+    // For simplicitly, just trim the quotes instead of implementing a full shell
+    // string grammar.
+    let value = if (value.starts_with('\'') && value.ends_with('\''))
```
gordonwang0, line 125: `trim_matches` could replace this block.

gordonwang0 (approved):

aziotctl/aziotctl-common/src/host_info.rs
arsing: I think I wrote it this way to ensure the quotes are matched up on either end.

gordonwang0: Got it. Yeah, that makes sense.

PR #343, opened 2022-01-05
Add documentation for socket throttling

gordonwang0:
Documents the socket throttling mechanism of the services, which limits all users to 10 simultaneous requests.

### DIFF
--- a/docs/socket-throttling.md
+++ b/docs/socket-throttling.md
@@ -1,3 +1,13 @@
 # Socket Throttling
 
-Each of the services' sockets implement a throttling mechanism to prevent a malicious process from making continuous requests and denying service to other processes.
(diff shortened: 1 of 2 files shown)


docs/socket-throttling.md
```diff
@@ -0,0 +1,13 @@
+# Socket Throttling
+
+Each of the services' sockets implement a throttling mechanism to prevent a malicious process from making continuous requests and denying service to other processes. This throttling mechanism limits each UNIX user to making 10 simultaneous requests per socket.
```
varunpuranik, line 3: Nit: While technically correct, this can be a bit confusing. 
Can we say something like -
"This throttling mechanism limits each caller to make 10 simultaneous requests per socket. Each caller is identified by its UID, so 2 processes sharing the same UNIX user will be throttled together".

varunpuranik (approved):

docs/socket-throttling.md
gordonwang0: Fixed

PR #351, opened 2022-01-12
Generate SBOM (Software BIll of Materials)

nlcamp:
These changes turn the packages workflow into a reusable workflow and adds a new packages-release workflow. The new packages-release workflow calls the packages workflow, downloads its artifacts, and then uses them to generate a software bill of materials (SBOM).

### DIFF
--- a/.github/workflows/packages-release.yaml
+++ b/.github/workflows/packages-release.yaml
@@ -19,11 +19,10 @@ jobs:
       - name: 'Download Artifacts'
         uses: actions/download-artifact@v2
         with:
-          path: $GITHUB_WORKSPACE/release-artifacts
+          path: release-artifacts
       - name: 'Generate SBOM'
         id: 'generate-sbom'
         run: |
-          mkdir $GITHUB_WORKSPACE/release-artifacts
           sudo apt-get install -y dotnet-sdk-3.1
           dotnet new console
           dotnet nuget add source https://1essharedassets.pkgs.visualstudio.com/_packaging/Packaging/nuget/v3/index.json --username IoTEdge1-PAT-1esSharedAssets --password ${{ secrets.IOTEDGE1_1ES_SHARED_ASSETS_PAT }} --store-password-in-clear-text
--- a/.github/workflows/packages-release.yaml
+++ b/.github/workflows/packages-release.yaml
@@ -1,0 +1,32 @@
+name: 'packages-release'
+
+on:
+  workflow_dispatch:
+    branches: [release/1.2]
+
+env:
+  PACKAGE_VERSION: '1.2.0~rc4'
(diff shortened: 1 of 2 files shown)


nlcamp:
Leaving this as a draft PR until I've completed testing. I've already completed testing in my forked repo

arsing:
What is this? And how is it different from the THIRDPARTYNOTICES file that we already bundle?

nlcamp:
> What is this? And how is it different from the THIRDPARTYNOTICES file that we already bundle?

There's a Cyber Executive Order that mandates that we generate an SBOM for all software sold to the federal government. See more at aka.ms/SBOM.

arsing:
Okay, and to answer my second question, it's a specific format so the TPN can't be reused.

Anyway, this needs to be merged into `main` first, then cherry-picked to `release/1.2`

.github/workflows/packages-release.yaml
```diff
@@ -0,0 +1,34 @@
...
+  workflow_dispatch:
+    branches: [release/1.2]
+
+jobs:
+  packages:
+    # TODO: Replace sbom-generation branch w/ release/1.2
+    uses: nlcamp/iot-identity-service/.github/workflows/packages.yaml@sbom-generation-1.2
+  sbom:
+    needs: packages
+    runs-on: ubuntu-latest
```
arsing, line 16: Pin a specific version like the other workflows. Otherwise a new OS release can cause breakage.

.github/workflows/packages-release.yaml
```diff
@@ -0,0 +1,34 @@
+# This workflow is intended to be run manually when cutting an official release.
```
arsing, line 1: Is there a reason not to do this in the packages workflow? ie is there a cost to generating the SBOM for every packages run?

.github/workflows/packages.yaml
```diff
@@ -1,8 +1,14 @@
...
 
 on:
-- 'pull_request'
-- 'push'
+  pull_request:
+  push:
+  workflow_call:
+    outputs:
+      version:
+        value: ${{ jobs.packages.steps.run.env.PACKAGE_VERSION }}
```
arsing, line 9: Enclose `${{}}` strings in `"`.

arsing (changes requested):

.github/workflows/packages-release.yaml
nlcamp: I was originally thinking we'd want to restrict permission to run the SBOM job to only Microsoft devs. However, that's probably not necessary and more work would have to be done to accomplish that goal, so I'll just move this to packages.yaml.

nlcamp:
Closing this PR as it is not required at this time. Will open a new one in the future if necessary.

PR #354, opened 2022-01-19
File exclusions for code coverage

brotherneeru:
Including following exclusions for tarpaulin code coverage:
•	identity/mock-dps-server/* (test code)
•	key/aziot-key-openssl-engine-shared-test/* (test code)
•	openssl-build/* (build code)
•	test-common/* (test code)
•	tpm/aziot-tpm-sys/aziot-iot-hsm-c/deps/* (git submodule,)
•	tpm/aziot-tpm-sys/tests/* (test code)

In addition, deleting /iotedged directory


arsing:
As I'd said in the email thread, we should just delete iotedged. It has served its purpose and I don't think it even works any more.

docs-dev/running/index.md
```diff
@@ -20,43 +20,6 @@
...
-        # The value of `--preloaded-device-id-ca-cert` matches
-        # the name of the key preloaded into keyd and
-        # the name of the cert preloaded into certd.
-        cargo run -p iotedged -- \
-            --hub-id 'example.azure-devices.net' \
-            --device-id 'example-1' \
-            --preloaded-device-id-ca-cert 'device-id-ca'
-        ```
-
-`iotedged` is a test binary that performs some operations that would be done by the Identity Service or Edge Module Runtime. Specifically, it connects to `aziot-keyd` and `aziot-certd` and does the following:
```
arsing, line 58: The list of three points below this is also about the `iotedged` binary, so delete that too.

arsing:
Lockfile still references it. You need to do a build. `make -s test-release` locally should do it.

(We should make `test-release` do a `--locked` build in the future so that such issues are caught...)

Makefile
```diff
@@ -212,6 +209,11 @@ codecov: default
 	
 	+LD_LIBRARY_PATH="$$LD_LIBRARY_PATH:$$PWD/target/$(CARGO_TARGET)/$(CARGO_PROFILE_DIRECTORY)" $(CARGO) tarpaulin --all --verbose \
 		--exclude aziot-key-openssl-engine-shared \
+		--exclude openssl-build --exclude test-common \
+		--exclude-files identity/mock-dps-server/* \
```
gordonwang0, line 213: Why the distinction between `--exclude` for some directories and `--exclude-files` for others? Shouldn't `--exclude identity/mock-dps-server` work?

arsing: `--exclude` takes a crate name, so it would be `--exclude mock-dps-server`, but yes it should be made consistent.

brotherneeru: If thats the case then for tpm/aziot-tpm-sys/azure-iot-hsm-c/deps and the aziot-tpm-sys test code I'll have to keep it as a wildcard to exclude all files in that folder?

arsing: Probably. Try `--exclude aziot-tpm-sys` and see if it's smart enough to also exclude the C files, but I doubt it, since the object files are just registered for being linked in rather than part of the crate itself.

brotherneeru: Yup. It didn't exclude the C files . Updated the PR with a more consistent arrangement

gordonwang0 (approved):

arsing (approved):

PR #357, opened 2022-01-31
Add Hub mocking to mock-iot-server

gordonwang0:
- Renames mock-dps-server to mock-iot-server.
- Adds capability to mock Hub get_modules requests. This allow the full provisioning and identity reconciliation flow to be tested.


identity/mock-iot-server/README.md
```diff
@@ -0,0 +1,19 @@
+# mock-iot-server
+
+mock-iot-server provides a subset of DPS server and IoT Hub functionality for testing. In its current form, it is *not* a substitute for testing with the real IoT.
```
arsing, line 3:
```diff
-with the real IoT.
+with a real DPS or IoT Hub.
```

identity/aziot-identityd/src/lib.rs
```diff
@@ -455,6 +460,29 @@ impl Api {
...
+            return true;
+        }
+
+        let prev_settings = if let Ok(settings) = crate::configext::load_file(&prev_settings_path) {
+            settings
+        } else {
+            return true;
+        };
+
+        if prev_settings.provisioning == self.settings.provisioning {
```
arsing, line 477: This won't catch cases where key ID / cert ID are the same but the key / cert is different now.

We expect people to run `aziotctl system reprovision` after changing provisioning settings, right? Is there a reason to automatically reprovision here, especially since it might be confusing when it happens in some cases but not others?

arsing (changes requested):

identity/mock-iot-server/README.md
gordonwang0: Fixed

identity/aziot-identityd/src/lib.rs
gordonwang0: I was thinking that there's no scenario where someone would change the provisioning settings and not want Identity Service to reprovision. Having to run the `system reprovision` command just seems like an unnecessary extra step. True, this change doesn't catch all cases where the config changes, but I think it's better than nothing.

arsing (approved):

PR #359, opened 2022-02-03
Add Hub mocking to mock-iot-server (#357)

gordonwang0:
- Renames mock-dps-server to mock-iot-server.
- Adds capability to mock Hub get_modules requests. This allow the full provisioning and identity reconciliation flow to be tested.


gordonwang0:
Cherry-pick this change to release/1.2 to keep the test suite consistent between branches.

arsing (approved):

gordonwang0:
@arsing Do you know why BTreeSet requires Ord on Module here? Rust 1.47?

arsing:
Yes, 1.56 removed the requirement for the elements to be `Ord` since `new()` doesn't require that.

arsing:
To be clear, you still can't meaningfully use it as a `BTreeSet` for `get` / `insert` without the elements being `Ord`. I see that you don't actually use it in the `main` code, which is why you didn't notice.

gordonwang0:
Got it, thanks. Changed to use HashMap since the ordering doesn't ever matter. Also made a similar fix to main #360

PR #360, opened 2022-02-03
Minor fixes for mock-iot-server

gordonwang0:
- Fix spacing and permissions on mock-iot-setup.sh
- Change mock-iot-server to use HashSet for Modules, since ordering isn't important.

### DIFF
--- a/identity/mock-iot-server/src/hub.rs
+++ b/identity/mock-iot-server/src/hub.rs
@@ -3,7 +3,7 @@
 use crate::server::Response;
 
 fn list_modules(
-    modules: &std::collections::BTreeSet<aziot_identity_common::hub::Module>,
+    modules: &std::collections::HashSet<aziot_identity_common::hub::Module>,
 ) -> Response {
     let mut response = vec![];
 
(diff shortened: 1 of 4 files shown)


identity/mock-iot-server/src/hub.rs
```diff
@@ -3,7 +3,7 @@
 use crate::server::Response;
 
 fn list_modules(
-    modules: &std::collections::BTreeSet<aziot_identity_common::hub::Module>,
+    modules: &std::collections::HashSet<aziot_identity_common::hub::Module>,
```
nyanzebra, line 6: This is for correctness or optimization?

gordonwang0: Optimization. Ordering doesn't matter for this, so it's fine to use HashSet.

arsing:
`Module` isn't `Eq + Hash` either, so this will still have the same problem when you eventually use the set.

gordonwang0:
True, but since this is test code and we don't have tests that require it, I think it's fine as is. If we add tests in the future that require Modules to be inserted into the set, we can add the derives to the Module struct.

arsing (approved):

nyanzebra (approved):

PR #361, opened 2022-02-04
[main] Update regex and env_logger to latest versions

onalante-msft:
"regex" was not vulnerable to the "thread_local" vulnerability in this version, but since iotedge will be updating to the latest versions of the aforementioned packages, we should update here as well.


arsing (approved):

arsing:
(Test failure fixed in https://github.com/Azure/iot-identity-service/pull/363 )

arsing (approved):

PR #364, opened 2022-02-07
Always provide DPS registration ID in get_provisioning_info

gordonwang0:
The DPS registration ID is always known to Identity Service and should always be provided in the get_provisioning_info response.

### DIFF
--- a/identity/aziot-identityd/src/lib.rs
+++ b/identity/aziot-identityd/src/lib.rs
@@ -220,37 +220,49 @@ impl Api {
                     config::DpsAttestationMethod::Tpm { registration_id } => {
                         ("tpm".to_string(), registration_id.to_string())
                     }
-                    config::DpsAttestationMethod::X509 { identity_cert, .. } => {
-                        // This API call is only available after a successful provision, which means
-                        // that the device ID certificate was used. The registration ID is taken from
-                        // the CN of the device ID certificate.
-                        let identity_cert = self
-                            .cert_client
-                            .get_cert(identity_cert)
-                            .await
-                            .map_err(|err| {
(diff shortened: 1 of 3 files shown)


identity/aziot-identityd/src/lib.rs
```diff
@@ -204,6 +204,83 @@ impl Api {
...
+                    config::DpsAttestationMethod::SymmetricKey {
+                        registration_id, ..
+                    } => ("symmetric_key".to_string(), registration_id.to_string()),
+                    config::DpsAttestationMethod::Tpm { registration_id } => {
+                        ("tpm".to_string(), registration_id.to_string())
+                    }
+                    config::DpsAttestationMethod::X509 { identity_cert, .. } => {
+                        // The identity certificate's common name is used as registration ID and always
+                        // overrides any registration ID set in the config. Therefore, the registration
+                        // ID must always be retrieved from the identity certificate.
```
arsing, line 226: `provision_device` prefers the `registration_id` in the config over the cert CN, no?

gordonwang0: So it's a bit complicated:
- If an identity certificate exists and is not within a day of expiry, IS uses that certificate's CN as registration ID and ignores the registration ID in the config.
- Otherwise, if no cert exists or cert is expired, the priority is:
  - config's registration ID
  - old certificate's CN

For this API, which is only available after a successful provision (i.e. when the device identity certificate has been successfully used and therefore must exist and be valid), the CN of the certificate will always be the registration ID.

gordonwang0: I updated this comment to make it clearer.

arsing (approved):

PR #372, opened 2022-02-23
Update version & changelog

yophilav:
Prepare for release 1.2.6


yophilav:
@damonbarry  / @varunpuranik / @micahl 
What's your thoughts on the autogenerated changelog messages in CHANGELOG.md?

micahl:
lgtm.  if possible i'd have it only mention the headline change for each change with a link to it.  if someone wanted to get more details (e.g. who made the change) then they could follow the link

damonbarry:
Isn't this 1.2.8, not 1.2.6?

damonbarry (approved):

arsing:
>What's your thoughts on the autogenerated changelog messages in CHANGELOG.md?

The branch name prefixed into every bullet would be nice to remove, since it's just noise.

The CI is complaining about the file not containing a trailing newline, so if that can be fixed that would be great, otherwise we'll need to change the makefile rule to skip it.

But most importantly, can we make GH releases from now on, and put this info there instead of in a checked-in file? That way it's available in the release feed instead of requiring people to hunt in the file.

yophilav:
@arsing 
Thanks for the comment! 
1. The 1-liner for the commit description comes from the title of the commit message. I think we make set a standard for the commit title so that it will be changelog-ready for the upcoming release. In the meantime, I can definitely do the manual touch up on those strings.
2. For the CI Newline, I'm not super sure if there is something I can modify from the github side. I'll need to do more digging on the API. I think the easier way to do it is to amend a rule for CI? 
3. That seems to be the standard way that other people are doing it. For our team in /iotedge, we have both the GH release form AND changelog.md. I wonder if we should keep the consistency or if we should no longer maintaining the changelog.md file. @micahl thinks we should have a changelog file.

arsing:
> 1. The 1-liner for the commit description comes from the title of the commit message. I think we make set a standard for the commit title so that it will be changelog-ready for the upcoming release. In the meantime, I can definitely do the manual touch up on those strings.

Ah yes, I see all of Tarik's commits have that prefix in the commit message itself. But yes, please remove them.

> 2. For the CI Newline, I'm not super sure if there is something I can modify from the github side. I'll need to do more digging on the API. I think the easier way to do it is to amend a rule for CI?

Okay.

> 3. That seems to be the standard way that other people are doing it. For our team in /iotedge, we have both the GH release form AND changelog.md. I wonder if we should keep the consistency or if we should no longer maintaining the changelog.md file. @micahl thinks we should have a changelog file.

The one we have in iotedge hasn't been updated since 1.0.8 actually, which is exactly what I was worried about, namely that if it's in two places it's just more work to keep them in sync. Unless there's some Azure requirement that we should have one or the other or both, my preference is releases-only > both changelog-and-releases > changelog-only.

yophilav:
> The 1-liner for the commit description comes from the title of the commit message. I think we make set a standard for the commit title so that it will be changelog-ready for the upcoming release. In the meantime, I can definitely do the manual touch up on those strings.
Ah yes, I see all of Tarik's commits have that prefix in the commit message itself. But yes, please remove them.

Will do

> For the CI Newline, I'm not super sure if there is something I can modify from the github side. I'll need to do more digging on the API. I think the easier way to do it is to amend a rule for CI?
Okay.

I can find the API here (https://docs.github.com/en/rest/reference/releases#generate-release-notes-content-for-a-release) though they didn't have an example of '.github/release.yml' or '.github/release.yaml' for `configuration_file_path` parameter. I don't know what the config may entails.

> That seems to be the standard way that other people are doing it. For our team in /iotedge, we have both the GH release form AND changelog.md. I wonder if we should keep the consistency or if we should no longer maintaining the changelog.md file. @micahl thinks we should have a changelog file.
The one we have in iotedge hasn't been updated since 1.0.8 actually, which is exactly what I was worried about, namely that if it's in two places it's just more work to keep them in sync. Unless there's some Azure requirement that we should have one or the other or both, my preference is releases-only > both changelog-and-releases > changelog-only.

We update the changelog for a particular branch of release in /iotedge (not the /azure-iotedge) repository. We are planning on removing the changelog.md of the /azure-iotedge in this release.

damonbarry:
I'm also in favor of handling the release notes in the simplest way possible, e.g., GitHub Release page only, no changelog. @micahl  what do you think?

micahl:
> I'm also in favor of handling the release notes in the simplest way possible, e.g., GitHub Release page only, no changelog. @micahl Micah Lewis FTE what do you think?

100% agree.  Perhaps I have used the term 'changelog' too loosely.  What we need to ensure is that we have a short summary of what has changed from one release to the next and that it is easily consumable.

PR #376, opened 2022-03-01
Add certificate renewal policy to settings

gordonwang0:
- Adds the cert-renewal crate and its Policy type, a new setting to control certificate renewal.
- Adds tests for Policy


aziotctl/aziotctl-common/test-files/apply/dps-x509-pkcs11-est-bootstrap/config.toml
```diff
@@ -23,5 +23,9 @@ password = "estpwd"
 bootstrap_identity_cert = "file:///var/secrets/est-bootstrap-id.pem"
 bootstrap_identity_pk = "pkcs11:slot-id=0;object=est-bootstrap-id?pin-value=1234"
 
+[cert_issuance.est.identity_auto_renew]
```
varunpuranik, line 26: I thought we were planning to have a auto_renew explicit flag as well..

gordonwang0: I did that at first, but I think this way looks better.
- threshold and retry have no effect if auto_renew is false. This way, those settings don't exist when the [auto_renew] section isn't set.
- This allows the setting to be called something other than just `auto_renew`. So you can distinguish, e.g. `identity_auto_renew`.

If you agree, I'll update the doc for this.

cert/cert-renewal/src/policy.rs
```diff
@@ -0,0 +1,361 @@
...
+            '%' => {
+                if value >= 100 {
+                    return Err(serde::de::Error::custom(
+                        "lifetime percentage must be less than 100",
+                    ));
+                }
+
+                Ok(Policy::Percentage(value))
+            }
+            'm' => Ok(Policy::Time(value * 60)),
```
arsing, line 99: I wonder if people will think "m" means "months" instead of "minutes", given that the other option is "days".

Maybe we should require them in full? "day", "min"

arsing (changes requested):

aziotctl/config/unix/template.toml
```diff
@@ -120,6 +120,18 @@
...
+#
+## Uncomment the section below to enable automatic renewal of the device ID certificate.
+## Auto-renewal requires a known certificate issuance method. This generally means that
+## 'method' is either 'est' or 'local_ca'.
+##
+## IMPORTANT:
+## ONLY ENABLE AUTO-RENEWAL IF THIS DEVICE IS CONFIGURED FOR CA-BASED DPS ENROLLMENT.
+## USING AUTO-RENEWAL FOR AN INDIVIDUAL ENROLLMENT WILL CAUSE THE DEVICE TO BE UNABLE
+## TO REPROVISION.
+# [provisioning.attestation.identity_cert.auto_renewal]
```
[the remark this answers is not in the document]
gordonwang0: Thanks, fixed.

cert/cert-renewal/src/policy.rs
gordonwang0: Changed to "day" and "min".

cert/cert-renewal/src/policy.rs
```diff
@@ -66,12 +66,29 @@ impl<'de> serde::Deserialize<'de> for Policy {
...
         } else {
             return Err(serde::de::Error::custom("policy not specified"));
         };
 
+        let mut unit = last.to_string();
+
+        // Parse for "min" or "day".
+        if last == 'n' || last == 'y' {
+            if let Some(last) = policy.pop() {
+                unit = last.to_string() + &unit;
```
arsing, line 80: `unit.insert(0, last);`

But also, something like `policy.len().checked_sub(3).map(|prefix_len| policy.split_at(prefix_len))` will give you the prefix and suffix in one shot.

arsing (approved):

cert/cert-renewal/src/policy.rs
arsing: Ah, right.

PR #377, opened 2022-03-02
Create Mariner 2.0 Build for Iot-Identity-Service

josephknierman:
Adds an additional build for mariner 2.0. This requires a separate build of the mariner toolkit to be used. As older packages are not being backported to mariner 2.0, this means we would not have access to the correct versions of GCC and Rust. A patch to the mariner spec was added file which prevents warnings from c-shared being propagated as errors when building. The patch also modifies the arguments to the hmacResult function as they were mismatched causing a warning which would result in an error. These warnings are a result of using a gcc version greater that 9 and Mariner 2.0 only provides 11+. 

This also will utilize the rust toolkit provided by the install-build-deps.sh script. this will then be packaged and forwarded to the spec files sources directory.


arsing:
>and also requires the build to use rust version 1.56.1

Non-Mariner builds also need to be compiled with Rust 1.56.1 then. Just as I had said last time, we can't afford to support multiple Rust releases in a single release branch; it's just going to cause problems where devs test against one version and CI against another, or where one version requires something that the other doesn't support.

ie https://github.com/Azure/iot-identity-service/blob/release/1.2/rust-toolchain needs to be changed to 1.56.1

But also, since you can't compile iotedge 1.2 with Rust 1.56.1, as I'd told you over email, what is your plan going to be for that repo? There are crates shared between i-i-s and iotedge repos so you're not going to find it easy to compile them with different Rust versions, for the same reason as my first paragraph.

contrib/mariner/compile.patch
```diff
@@ -0,0 +1,44 @@
+Submodule tpm/aziot-tpm-sys/azure-iot-hsm-c/deps/c-shared contains modified content
```
arsing, line 1: Call it something more descriptive please. `gcc-11.patch` perhaps.

arsing (changes requested):

contrib/mariner/compile.patch
josephknierman: changed name to be gcc-11.patch

josephknierman:
Note currently the build is failing due to a change in mariner 2.0-stable branch and this would not be fixed by changes to this PR. a passing run for 2.0-mariner can be found here before they changes they added last night hopefully this will be resolved soon. https://github.com/josephknierman/iot-identity-service/actions/runs/1919761332

josephknierman:
I see the comment about the rust versioning. I thought that only applied to iotedge and not IIS. I think i know a way to get t to work with both, working on that now

contrib/mariner/aziot-identity-service.spec
```diff
@@ -57,6 +58,15 @@ This package contains development files for the Azure IoT device runtime.
...
 
 %prep
+# include rust toolchain that matches the one from iot-identity-service's pipeline
+pushd ~
+tar xf %{SOURCE3} --no-same-owner --strip-components=1
+popd
+export CARGO_HOME=~/.cargo
+export PATH=$PATH:$CARGO_HOME/bin
+export RUSTUP_HOME=~/.rustup
+export RUSTUP_TOOLCHAIN=~/.rustup/toolchains/1.47-x86_64-unknown-linux-gnu
```
arsing, line 68: Why do you need this? When `cargo` is invoked in the source directory (as part of `make ...`), it will already find the `rust-toolchain` file and know which toolchain to use.

josephknierman: without explicitly setting the toolchain i get the error bellow when using cargo

time="2022-03-23T12:32:34-04:00" level=debug msg="+ cargo install bindgen --path rust-bindgen-0.54.0 --offline"
time="2022-03-23T12:32:34-04:00" level=debug msg="error: no override and no default toolchain set"

arsing: Like I said, make sure to `cd` into the source directory, ie the root of the repository, ie the directory that contains the `rust-toolchain` file, first.

josephknierman: thanks for the info. this should be fixed in the spec file now.

arsing (approved):
Thanks.

Please update the PR title and description to match the latest situation (eg it doesn't use 1.56 any more) since they are used to construct the commit title and message.

Also please cherry-pick it to main, and include the fix from https://github.com/Azure/iot-identity-service/pull/380

gordonwang0 (approved):

PR #388, opened 2022-04-08
Add rotate_key setting and update principals for cert renewal

gordonwang0:
- Adds the `rotate_key` setting for cert renewal. Defaults to `true`.
- Updates authorized principals to allow access to `temp` keys when necessary.

### DIFF
--- a/aziotctl/aziotctl-common/src/config/apply.rs
+++ b/aziotctl/aziotctl-common/src/config/apply.rs
@@ -213,6 +213,16 @@ pub fn run(
 
                                 aziotid_keys.keys.push(super::DEVICE_ID_ID.to_owned());
 
+                                // Identity Service needs authorization to manage a temporary key
+                                // during key rotation.
+                                if let Some(auto_renew) = &auto_renew {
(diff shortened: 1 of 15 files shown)


aziotctl/aziotctl-common/src/config/apply.rs
```diff
@@ -213,6 +213,22 @@ pub fn run(
 
                                 aziotid_keys.keys.push(super::DEVICE_ID_ID.to_owned());
 
+                                // Identity Service needs authorization to manage temporary credentials
+                                // during cert rotation.
+                                if let Some(auto_renew) = &auto_renew {
+                                    let temp = format!("{}-temp", super::DEVICE_ID_ID);
+
+                                    cert_issuance_certs.insert(
```
gordonwang0, line 221: @arsing There's currently no way to specify issuance options in the create_cert API call or tell it to use another cert's issuance options. So we must create a "device-id-temp" cert issuance entry that duplicates the settings for "device-id". Thoughts?

arsing (approved):

PR #389, opened 2022-04-11
[main] Upgrade to latest Rust version

onalante-msft:
This is our monthly toolchain upgrade to the latest stable version. Since the primary artifacts of this repository are binaries, we want to be tracking Rust compiler releases closely in the event a stdlib vulnerability is found. We do not pin to "stable", however, since that has caused pipeline breakage when some code patterns are made illegal (like in the upgrade from 1.47 to 1.48 [^0]) or, more often, clippy lints are added.

- Use `std::ptr::addr_of` and `std::ptr::addr_of_mut` for pointers to places

[^0]: https://github.com/rust-lang/rust/blob/master/RELEASES.md#compatibility-notes-11
  Namely, the point on `mem::uninitialized`.

### DIFF
--- a/http-common/src/connector.rs
+++ b/http-common/src/connector.rs
@@ -881,6 +881,7 @@ mod tests {
             assert_eq!(*expected, deserialized_connector);
         }
 
+        #[allow(clippy::single_element_loop)]
         for input in &[
             // unsupported scheme
             "ftp://127.0.0.1",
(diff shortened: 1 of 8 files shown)


arsing:
The CI failures with softhsm are 0xd1 == `CKR_TEMPLATE_INCONSISTENT`, so it might be a genuine failure due to this change.

onalante-msft:
Closed in favor of #390.

PR #392, opened 2022-04-12
Add renewal of device identity certificate

gordonwang0:
- Adds feature to automatically renew device identity certificate.
- Removes behavior where device identity certificate renews within 1 day of expiry. This is superseded by the auto-renew feature.

### DIFF
--- a/identity/aziot-identityd/src/lib.rs
+++ b/identity/aziot-identityd/src/lib.rs
@@ -738,7 +738,7 @@ impl UpdateConfig for Api {
 
 pub(crate) async fn get_keys(
     key_handle: aziot_key_common::KeyHandle,
-    key_engine: &Arc<futures_util::lock::Mutex<openssl2::FunctionalEngine>>,
+    key_engine: &futures_util::lock::Mutex<openssl2::FunctionalEngine>,
 ) -> Result<
     (
         openssl::pkey::PKey<openssl::pkey::Private>,
--- a/identity/aziot-identityd/src/lib.rs
+++ b/identity/aziot-identityd/src/lib.rs
@@ -28,6 +28,7 @@ pub mod configext;
 pub mod error;
 mod http;
 pub mod identity;
+mod renewal;
 
 use config_common::watcher::UpdateConfig;
 pub use error::{Error, InternalError};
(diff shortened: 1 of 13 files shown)


identity/aziot-identityd/src/renewal.rs
```diff
@@ -0,0 +1,265 @@
...
+                .await
+                .map_err(|_| cert_renewal::Error::fatal_error("failed to restore old cert"))?;
+        }
+
+        // Reprovisioning the device so that the new identity cert is registered. Note that if this
+        // fails, the device will remain in an error state until the next renewal retry as it can no
+        // longer fall back to the old identity cert.
+        let mut api = self.api.lock().await;
+
+        api.reprovision_device(
```
gordonwang0, line 251: @arsing @varunpuranik This is potentially a problem because a reprovisioning failure will put the device in an error state until the next renewal retry (potentially days/or weeks for long-lived certificates). At this point, the device can no longer fall back to old credentials because they were erased by move_key_pair.

Perhaps we should change move_key_pair to swap_key_pair so that we can fall back to the old key if reprovisioning fails?

gordonwang0: Upon further discussion, we'll reprovision first then write the credentials. This will cause the device to go into an error state if the write fails (as the device was provisioned with credentials that won't be saved). But that will only happen if certd and keyd are down, at which point nothing will work.

identity/aziot-identityd/src/lib.rs
```diff
@@ -655,15 +736,30 @@ impl Api {
...
-    type Config = config::Settings;
-    type Error = Error;
-
-    async fn update_config(&mut self, new_config: config::Settings) -> Result<(), Self::Error> {
-        self.update_config_inner(new_config, ReprovisionTrigger::ConfigurationFileUpdate)
-            .await
-    }
+pub(crate) async fn get_keys(
+    key_handle: aziot_key_common::KeyHandle,
+    key_engine: &Arc<futures_util::lock::Mutex<openssl2::FunctionalEngine>>,
```
arsing, line 741:
```suggestion
    key_engine: &futures_util::lock::Mutex<openssl2::FunctionalEngine>,
```

identity/aziot-identityd/src/renewal.rs
```diff
@@ -0,0 +1,301 @@
...
+
+        let new_cert = openssl::x509::X509::from_pem(&new_cert)
+            .map_err(|_| cert_renewal::Error::retryable_error("failed to parse new cert"))?;
+
+        Ok((new_cert, key_id))
+    }
+
+    async fn write_credentials(
+        &mut self,
+        _old_cert: &openssl::x509::X509,
```
arsing, line 217: Remove unused param?

identity/aziot-identityd/src/renewal.rs
```diff
@@ -0,0 +1,301 @@
...
+            );
+        }
+
+        let new_cert = openssl::x509::X509::from_pem(&new_cert)
+            .map_err(|_| cert_renewal::Error::retryable_error("failed to parse new cert"))?;
+
+        Ok((new_cert, key_id))
+    }
+
+    async fn write_credentials(
```
arsing, line 215: Everything that wants to renew certs will need to use this "create new key (maybe) + new cert -> use new key and cert for some operation -> rename to overwrite" algorithm, right? If so, should this be in the cert-renewal crate where the trait fn is invoked, and the trait fn itself only handles the "some operation" part?

arsing (changes requested):

identity/aziot-identityd/src/renewal.rs
gordonwang0: It's used by other implementers of this trait. They revert to using the old cert if something fails. Renewal of the device identity certificate doesn't use it because it has already reprovisioned the device with the new cert, so it would have to provision again to revert back.

identity/aziot-identityd/src/lib.rs
gordonwang0: Fixed.

identity/aziot-identityd/src/renewal.rs
gordonwang0: No, "use new key and cert for some operation" is unique to the device identity cert. Renewing the Edge CA cert requires restarting modules, but that operation doesn't depend on the new key and cert. Renewing EST identity certs has no additional operation.

arsing (approved):

PR #395, opened 2022-04-25
Improve stability of E2E tests

gordonwang0:
- Updates VMs to more recent images
- sudo apt-get update before installing jq. This resolves an issue where the package sometimes isn't found.
- Add retries to GitHub API calls
- Disable el8 in manual runs, as it's disabled in scheduled runs.

### DIFF
--- a/ci/e2e-tests/test-run.sh
+++ b/ci/e2e-tests/test-run.sh
@@ -745,8 +745,6 @@ case "$OS" in
         scp -i "$PWD/vm-ssh-key" "$package" "aziot@$vm_public_ip:/home/aziot/aziot-identity-service.deb"
 
         ssh -i "$PWD/vm-ssh-key" "aziot@$vm_public_ip" '
-            set -euxo pipefail
-
             for retry in {0..3}; do
                 if [ "$retry" != "0" ]; then
                     sleep 10
@@ -758,6 +756,8 @@ case "$OS" in
                     break
                 fi
             done
+
+            set -euxo pipefail
 
             sudo DEBIAN_FRONTEND=noninteractive apt-get install -y bc curl jq perl
             sudo DEBIAN_FRONTEND=noninteractive apt-get install -y /home/aziot/aziot-identity-service.deb
--- a/ci/e2e-tests/test-run.sh
+++ b/ci/e2e-tests/test-run.sh
@@ -54,12 +54,27 @@ get_package() {
     #
     # Ref: https://docs.github.com/en/rest/reference/actions#list-workflow-runs
     echo "Getting latest workflow run's artifacts URL..." >&2
-    artifacts_url="$(
-        github_curl -L \
-            -H 'accept: application/vnd.github.v3+json' \
-            "$GITHUB_API_URL/repos/$GITHUB_REPOSITORY/actions/workflows/packages.yaml/runs?branch=${BRANCH//\//%2f}&event=push&status=success" |
-            jq -r '.workflow_runs[0].artifacts_url'
-    )"
+    artifacts_url='null'
+
+    # GitHub API calls may fail if too many other tests make the same call concurrently.
+    # Allow a few retries before failing.
+    for retry in {0..3}; do
+        if [ "$retry" != '0' ]; then
+            sleep 10
+        fi
+
+        artifacts_url="$(
+            github_curl -L \
+                -H 'accept: application/vnd.github.v3+json' \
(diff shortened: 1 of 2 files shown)


ci/e2e-tests/test-run.sh
```diff
@@ -54,12 +54,27 @@ get_package() {
     #
     # Ref: https://docs.github.com/en/rest/reference/actions#list-workflow-runs
     echo "Getting latest workflow run's artifacts URL..." >&2
-    artifacts_url="$(
```
arsing, line 57: You haven't changed the `set -euo pipefail` at the start of the script, so any command that fails will still abort the script entirely.

You should still leave the `set` since it catches general programming errors, but it means that any command that you want to retry needs to suppress that, either by appending `|| :` or doing `set +e` for that command.

ci/e2e-tests/test-run.sh
```diff
@@ -103,25 +118,48 @@ get_package() {
...
+    for retry in {0..3}; do
+        if [ "$retry" != '0' ]; then
+            sleep 10
+        fi
+
+        github_curl -L \
+            -o package.zip \
+            "$artifact_download_url"
+
+        if [ -f package.zip ]; then
```
arsing, line 159: This will pass if the download aborts in between and you're left with an incomplete file. You should check the exit code of `github_curl` instead.

ci/e2e-tests/test-run.sh
```diff
@@ -689,8 +727,9 @@ case "$OS" in
         ssh -i "$PWD/vm-ssh-key" "aziot@$vm_public_ip" '
```
arsing, line 727: Didn't we need a retry around this whole too? Because the package repo would sometimes be borked and not find jq?

arsing (changes requested):

ci/e2e-tests/test-run.sh
arsing: Oh, just saw the commit message.

ci/e2e-tests/test-run.sh
gordonwang0: I think the command still returns 0 (success), but the result is HTTP 429 and the jquery parse is 'null'. So it should be fine if e is still set.

ci/e2e-tests/test-run.sh
gordonwang0: But it will fail to unzip an incomplete file immediately after this. And with set e, the script will abort if curl fails, so we can assume the exit code was 0 by this line, right?

ci/e2e-tests/test-run.sh
arsing: And if curl does fail and returns a non-zero exit code? The point of this change was to retry when these external operations fail, right?

ci/e2e-tests/test-run.sh
arsing: If curl succeeds and unzip fails, that just means even the unzip needs to be inside the retry.

ci/e2e-tests/test-run.sh
gordonwang0: Ah right. Fixed.

ci/e2e-tests/test-run.sh
gordonwang0: Fixed

ci/e2e-tests/test-run.sh
arsing: This is still broken if curl succeeds but outputs something that isn't JSON, because jq will choke on that.

I think the cleanest solution, just like the other place, is to do a `set +e` inside the subshell used to run `github_curl | jq` and then check `$artifacts_url` for both `''` and `'null'`

gordonwang0: Oh right. Updated.

ci/e2e-tests/test-run.sh
```diff
@@ -103,25 +120,55 @@ get_package() {
...
+            github_curl -L \
+                -H 'accept: application/vnd.github.v3+json' \
+                "$artifacts_url" |
+                jq \
+                    --arg artifact_name "$artifact_name" \
+                    -r \
+                    '.artifacts[] | select(.name == $artifact_name) | .archive_download_url'
+        )"
+
+        if [ "$artifact_download_url" != 'null' ] && [ ! -z "$artifact_download_url" ]; then
```
arsing, line 141: It would be clearer as:

```sh
if [ -n "$artifact_download_url" ] && [ "$artifact_download_url" != 'null' ]; then
```

arsing (approved):

ci/e2e-tests/test-run.sh
gordonwang0: Fixed

PR #406, opened 2022-05-03
Add SKID and AKID to certd-issued certs

gordonwang0:
Adds the Subject Key ID extension and the Authority Key ID extension to all certificates issued by Certificates Service. This is necessary to distinguish renewed CAs which all have the same CN.


cert/aziot-certd/src/lib.rs
```diff
@@ -384,27 +384,74 @@ async fn create_cert_inner<'a>(
...
+                {
+                    if basic_constraints.ca {
+                        let subj_key_id = openssl::x509::extension::SubjectKeyIdentifier::new();
+                        let context = builder.x509v3_context(issuer_ref, None);
+                        let subj_key_id = subj_key_id.build(&context)?;
+                        builder.append_extension(subj_key_id)?;
+                    }
+                }
+
                 builder.append_extension(ext)?;
```
gordonwang0, line 419: @arsing If a CSR already sets the SKID extension (not sure they would be able to set AKID as they probably don't have the issuer key), this line would append the SKID extension twice. Should we check to make sure SKID is only appended once? Right now it doesn't matter because none of the other services generate CSRs with SKID.

gordonwang0:
@arsing This change is necessary for renewal of the Edge CA cert (really, any CA) to distinguish two CAs with the same CN. This will cause other services to change cert chain validation from CN-based to SKID/AKID-based. Will this have any unintended side effects? I can't think of any.

arsing:
>Will this have any unintended side effects? I can't think of any.

Me neither.

cert/aziot-certd/src/lib.rs
arsing: Yes, we should only add our own SKID extension if there isn't one already in `req`.

cert/aziot-certd/src/lib.rs
```diff
@@ -393,18 +393,58 @@ async fn create_cert_inner<'a>(
...
-            (
-                x509.subject_name(),
-                if expiry < issuer_expiry {
-                    expiry
-                } else {
-                    issuer_expiry
-                },
-            )
-        });
+        builder.append_extension(auth_key_id)?;
```
arsing, line 447: As I mentioned on Teams, AKID should only be added when issuer is a separate cert. ie even self-signed certs shouldn't have an AKID, only a SKID. This matches up with the self-signed certs I see in ca-certificates.

If this is the source of the problem that necessitated the hack, then you can remove the hack too.

cert/aziot-certd/src/lib.rs
```diff
@@ -393,18 +393,58 @@ async fn create_cert_inner<'a>(
...
+            let issuer_key_id = issuer_key_id.build(&issuer_context)?;
+            issuer_builder.append_extension(issuer_key_id)?;
+
+            let issuer = issuer_builder.build();
+
+            let context = builder.x509v3_context(Some(&issuer), None);
+            auth_key_id.build(&context)?
+        } else {
+            let context = builder.x509v3_context(issuer_ref, None);
+            auth_key_id.build(&context)?
```
arsing, line 444: What does openssl do if the issuer doesn't have a SKID? Does it no-op the AKID extension or does it fail? If it fails, we should not fail the cert creation and instead just skip the AKID extension.

arsing (changes requested):

cert/aziot-certd/src/lib.rs
gordonwang0: Fixed.

cert/aziot-certd/src/lib.rs
gordonwang0: Fixed.

cert/aziot-certd/src/lib.rs
gordonwang0: Fixed.

arsing (approved):

PR #407, opened 2022-05-06
Drop debian 9 from main/1.3 since it's reaching EOL

gauravIoTEdge:
Debian 9 is reaching end of life. This PR removes Debian 9 from all tests/pipelines/builds.

**Testing Notes:**

image


arsing:
(I've removed the Debian 9 jobs from `main`'s branch policy, so they won't block this PR or future ones.)

.github/workflows/e2e-tests-scheduled.yaml
```diff
@@ -63,7 +63,6 @@ jobs:
         - 'release/1.2'
         os:
         - 'centos:7'
-        - 'debian:9'
```
arsing, line 66: It would be useful to continue testing `release/1.2` on it, so see if https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategymatrixinclude works, ie

```yaml
include:
- branch: 'release/1.2'
  os: 'debian-9'
```

Docs aren't clear on whether the extra job will also have all combinations of`test_name` or it'll have none of them. If it has none of them, then using `exclude` to exclude `branch: main / os: 'debian:9'` should work instead.

arsing (changes requested):

.github/workflows/e2e-tests-scheduled.yaml
damonbarry: +1. Debian 9 support isn't going into IoT Edge 1.3+, but I believe it should continue in 1.2 through Debian 9's support period (June, I believe).

gordonwang0 (approved):

.github/workflows/e2e-tests-scheduled.yaml
gauravIoTEdge: @gordonwang0 , @arsing - we're reaching EOL for Debian. Mind approving this PR please?

arsing: If we don't care about 1.2 any more then it should be removed from the file entirely,

arsing (approved):

PR #408, opened 2022-05-06
Add aarch64 Build for Mariner 1.0 and 2.0

josephknierman:
Adds aarch64 build for Mariner 1.0 and 2.0. This will not build the aarch64 version in the current package build pipelines. this will instead update the corresponding build scripts to work for aarch64 machines. The mariner aarch64 builds were not added to the normal package build yaml since mariner does not support cross-compiling and requires the packages to be built in a native aarch64 machine.


ci/install-build-deps.sh
```diff
@@ -176,7 +176,20 @@ mkdir -p ~/.cargo/bin
 export PATH="$PATH:$(realpath ~/.cargo/bin)"
 
 if ! [ -f ~/.cargo/bin/rustup ]; then
-    curl -Lo ~/.cargo/bin/rustup 'https://static.rust-lang.org/rustup/dist/x86_64-unknown-linux-gnu/rustup-init'
+    baseArch=$(uname -m)
```
arsing, line 179:
```suggestion
    baseArch="$(uname -m)"
```

arsing (approved):

ci/install-build-deps.sh
josephknierman: added quotes around uname command

PR #411, opened 2022-05-12
Truncate CNs to 64 characters in conversion to X509Name

gordonwang0:
Truncate CNs to 64 characters. The conversion won't work if they exceed the allowable length.

### DIFF
--- a/cert/aziot-certd-config/src/lib.rs
+++ b/cert/aziot-certd-config/src/lib.rs
@@ -235,7 +235,7 @@ impl std::convert::TryFrom<&CertSubject> for openssl::x509::X509Name {
                         let mut cn = value.to_string();
                         cn.truncate(CN_MAX_LENGTH);
 
-                        builder.append_entry_by_text(name, value)?;
+                        builder.append_entry_by_text(name, &cn)?;
                     } else {
                         builder.append_entry_by_text(name, value)?;
                     }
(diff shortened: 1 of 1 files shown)


cert/aziot-certd-config/src/lib.rs
```diff
@@ -217,13 +217,28 @@ impl std::convert::TryFrom<&CertSubject> for openssl::x509::X509Name {
...
+                builder.append_entry_by_nid(openssl::nid::Nid::COMMONNAME, &cn)?;
+            }
             CertSubject::Subject(fields) => {
                 for (name, value) in fields {
-                    builder.append_entry_by_text(name, value)?;
+                    if name.to_lowercase() == "cn" {
+                        let mut cn = value.to_string();
+                        cn.truncate(CN_MAX_LENGTH);
+
+                        builder.append_entry_by_text(name, value)?;
```
arsing, line 238: `cn` instead of `value`, no?

gordonwang0: Good catch! Fixed.

cert/aziot-certd-config/src/lib.rs
```diff
@@ -217,13 +217,28 @@ impl std::convert::TryFrom<&CertSubject> for openssl::x509::X509Name {
...
+            CertSubject::CommonName(cn) => {
+                let mut cn = cn.to_string();
+                cn.truncate(CN_MAX_LENGTH);
+
+                builder.append_entry_by_nid(openssl::nid::Nid::COMMONNAME, &cn)?;
+            }
             CertSubject::Subject(fields) => {
                 for (name, value) in fields {
-                    builder.append_entry_by_text(name, value)?;
+                    if name.to_lowercase() == "cn" {
```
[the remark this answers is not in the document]
gordonwang0: Yeah, I figured it's too annoying to check arbitrary subject fields, especially since we only really look at CN. Maybe see if anyone else encounters a problem.

arsing (approved):

PR #418, opened 2022-06-16
Disable RedHat debug artifact creation

yophilav:
Disable RedHat debug artifact creation


arsing:
Why?

yophilav:
> Why?
Actually, never mind, this repo has the debug rpm generated by default, so RedHat creating the artifact is normal. 
In the other repo (/iotedge), only CentOs generates the debug rpm.

Issue #433, opened 2022-07-26
Max simultaneous connections for user 1000


tom-seechange:
Hi. When using aziot-edge 1.3.0 with aziot-identity 1.3.0 our modules fail to deploy with the following errors:
```

<6> 2022-07-26 13:58:22.635 +00:00 [INF] [Microsoft.Azure.Devices.Edge.Agent.Core.Planners.HealthRestartPlanner] - Unable to process module x add or update as the module identity could not be obtained
<6> 2022-07-26 13:58:22.635 +00:00 [INF] [Microsoft.Azure.Devices.Edge.Agent.Core.Planners.HealthRestartPlanner] - Unable to process module y add or update as the module identity could not be obtained
<6> 2022-07-26 13:58:22.635 +00:00 [INF] [Microsoft.Azure.Devices.Edge.Agent.Core.Planners.HealthRestartPlanner] - Unable to process module z add or update as the module identity could not be obtained
<6> 2022-07-26 13:58:22.635 +00:00 [INF] [Microsoft.Azure.Devices.Edge.Agent.Core.Planners.HealthRestartPlanner] - Unable to process module a add or update as the module identity could not be obtained
<6> 2022-07-26 13:58:22.635 +00:00 [INF] [Microsoft.Azure.Devices.Edge.Agent.Core.Planners.HealthRestartPlanner] - Unable to process module edgeHub add or update as the module identity could not be obtained
<6> 2022-07-26 13:58:22.635 +00:00 [INF] [Microsoft.Azure.Devices.Edge.Agent.Core.Planners.HealthRestartPlanner] - Unable to process module b add or update as the module identity could not be obtained
<6> 2022-07-26 13:58:22.635 +00:00 [INF] [Microsoft.Azure.Devices.Edge.Agent.Core.Planners.HealthRestartPlanner] - Unable to process module c add or update as the module identity could not be obtained
<6> 2022-07-26 13:58:22.635 +00:00 [INF] [Microsoft.Azure.Devices.Edge.Agent.Core.Planners.HealthRestartPlanner] - Unable to process module d add or update as the module identity could not be obtained
<6> 2022-07-26 13:58:22.635 +00:00 [INF] [Microsoft.Azure.Devices.Edge.Agent.Core.Planners.HealthRestartPlanner] - Unable to process module e add or update as the module identity could not be obtained
[...]

PR #438, opened 2022-08-05
Add no-socket-throttle feature to http-common

gordonwang0:
Adds a cargo feature `no-socket-throttle` to http-common. Enabling this disables socket throttling when serving connections on UNIX sockets.

This is in response to #433. The 1.3.0 update consolidated the server code between iotedge and this repo, and unintentionally introduced this repo's throttling mechanism to the Edge daemon. The Edge daemon does not need to throttle the management socket, as only users in the iotedge group can access the management socket. By default, this prevents modules from calling the management socket.

### DIFF
--- a/http-common/Cargo.toml
+++ b/http-common/Cargo.toml
@@ -28,3 +28,5 @@ url = { version = "2", features = ["serde"] }
 [dev-dependencies]
 serde_json = "1"
 
+[features]
+no-socket-throttle = []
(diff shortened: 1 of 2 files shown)


varunpuranik (approved):

arsing:
Do we not want throttling on the workload socket?

gordonwang0:
The workload socket is split so that each module now has its own socket. That mitigates the risk of modules making repeated requests and denying service to other modules.

PR #440, opened 2022-08-06
[release/1.3] Fix check-in test pipeline

gordonwang0:
- *Cf.* https://github.com/microsoft/CBL-Mariner/issues/3483.
- chrono released v0.4.20 which has broken running tarpaulin. Upstream issue: https://github.com/chronotope/chrono/issues/755.

PR #442, opened 2022-08-08
Pin cargo-tarpaulin dependencies to its lockfile

gordonwang0:
chrono released v0.4.20 which has broken running tarpaulin.

Upstream issue: https://github.com/chronotope/chrono/issues/755.


gordonwang0:
Seems this can't be merged without #440. I'll combine the two commits and close this PR.

PR #444, opened 2022-08-08
[main] Update dependencies

onalante-msft:
*N.b.* `time 0.3` is a newly introduced duplicate dependency through `serde_with`. However, it is not built since we do not enable the `time_0_3` feature.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -43,9 +43,9 @@ checksum = "c794e162a5eff65c72ef524dfe393eb923c354e350bb78b9c7383df13f3bc142"
 
 [[package]]
 name = "async-recursion"
-version = "0.3.2"
+version = "1.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d7d78656ba01f1b93024b7c3a0467f1608e4be67d725749fdcd7d2c7678fd7a2"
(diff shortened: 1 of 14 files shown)


arsing:
What about `cargo outdated` ? In case there are new major versions.

---

>```
>jq: error (at <stdin>:1): crate unicode-ident 1.0.3 has unknown license (MIT OR Apache-2.0) AND Unicode-DFS-2016
>```

I think Unicode-DFS-2016 should be fine. Check to be sure.

As for fixing it, since it's just this one crate it's easier to just special-case the whole string in https://github.com/Azure/iot-identity-service/blob/bd484b560488521cd662d4029a42b9654c2285a7/contrib/third-party-notices.sh#L127 That is,

```diff
 ($license == "MIT") or
+($license == "(MIT OR Apache-2.0) AND Unicode-DFS-2016") or
 ($license | startswith("MIT/")) or
```

---

>```
>notify:4.0.17 at /root/.cargo/registry/src/github.com-1ecc6299db9ec823/notify-4.0.17/Cargo.toml has license CC0-1.0 but its license file could not be found
>```

https://github.com/Azure/iot-identity-service/blob/bd484b560488521cd662d4029a42b9654c2285a7/contrib/third-party-notices.sh#L223 needs s/4.0.15/4.0.17/

arsing (approved):

PR #445, opened 2022-08-11
Allow socket throttle limits to be configurable

gordonwang0:
A better solution to https://github.com/Azure/iot-identity-service/issues/433 is to add a config option to allow the throttling limits to be configurable. This PR adds the `max_requests` option, which controls the throttling limit.

Reverts #438 and removes the no-socket-throttle feature.

### DIFF
--- a/cert/aziot-certd-config/src/lib.rs
+++ b/cert/aziot-certd-config/src/lib.rs
@@ -28,6 +28,13 @@ pub struct Config {
     /// Path of home directory.
     pub homedir_path: PathBuf,
 
+    /// Maximum number of simultaneous requests per user that certd will service.
+    #[serde(
+        default = "http_common::Incoming::default_max_requests",
+        skip_serializing_if = "http_common::Incoming::is_default_max_requests"
+    )]
+    pub max_requests: usize,
+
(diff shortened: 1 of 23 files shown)


gordonwang0:
I didn't add this setting to template.toml, since this should be a rarely-used setting. If needed, we can point to the documentation.

cert/aziot-certd-config/src/lib.rs
```diff
@@ -358,6 +365,7 @@ mod tests {
     fn parse_config() {
         let actual = r#"
 homedir_path = "/var/lib/aziot/certd"
+max_requests = 50
```
varunpuranik, line 368: Oh.. so now we need to make the limit 50 so that EdgeAgent's requests succeed?

gordonwang0: This is test code. The default remains 10.

arsing (approved):

cert/aziot-certd-config/src/lib.rs
varunpuranik: ah, yes, missed that.

varunpuranik (approved):

PR #447, opened 2022-08-17
Fix all identity certs to create P-256 / RSA-2048 key pairs.

### LINKED ISSUE #403: EST ID and Device ID certificates default to using different key types
I'm setting up my service to use dps provisioning using x509 certificates from an est server using an est bootstrap certificate.
I noticed that the est id certificate keys are hard coded to use 256 bit EC keys by default, but the device id certificate is set to use 2048 bit RSA keys.

https://github.com/Azure/iot-identity-service/blob/84483897dc832f3343b9aee37370bd17c6aadfbb/cert/aziot-certd/src/lib.rs#L498
https://github.com/Azure/iot-identity-service/blob/84483897dc832f3343b9aee37370bd17c6aadfbb/identity/aziot-identityd/src/identity.rs#L846

It would be nice if they were consistent so I didn't have to configure my EST server to handle both types of certificate signing requests, or even better if there was somewhere in the configuration where the preferred algorithms for these keys could be set.

arsing:
Fixes #403

### DIFF
--- a/cert/aziot-certd/src/lib.rs
+++ b/cert/aziot-certd/src/lib.rs
@@ -521,7 +521,7 @@ async fn create_cert_inner<'a>(
                                 .key_client
                                 .create_key_pair_if_not_exists(
                                     &x509.identity.pk,
-                                    Some("ec-p256:rsa-4096:*"),
+                                    Some("ec-p256:rsa-2048:*"),
                                 )
                                 .await?;
                             let cstr = CString::new(handle.0)?;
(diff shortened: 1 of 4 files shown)


gordonwang0 (approved):
Mariner builds failing are probably due to a keychain update in mariner. Damon's team is taking a look.

onalante-msft (approved):

arsing:
In shiproom we decided to do this after the next release branch is branched, and to implement a configuration option for this to reduce the chance of regression from switching from RSA to EC certs. So I've marked this is Draft for now.

buentead:
I was wondering if there is an ETA for this PR to merge into main branch as I'm waiting for the fix of [EST ID and Device ID certificates default to using different key types #403](https://github.com/Azure/iot-identity-service/issues/403). Thanks.

PR #449, opened 2022-08-25
Remove `~dev` from package version

and-rewsmith:
Remove ~dev from package version


yophilav (changes requested):
This should only happen after the `release/1.4` is splited.

yophilav (approved):

PR #456, opened 2022-09-09
[tpmd] Fix regression in TPM authentication key index

gordonwang0:
The default index was incorrectly written, and the value inherited from the configuration was not used in the key import subroutine.

### DIFF
--- a/aziotctl/config/unix/template.toml
+++ b/aziotctl/config/unix/template.toml
@@ -227,8 +227,8 @@
 # # The TPM index at which to persist the DPS authentication key. The index is
 # # taken as an offset from the base address for persistent objects
 # # (0x81000000), and must lie in the range 0x00_00_00--0x7F_FF_FF. The default
-# # value is 0x00_10_00.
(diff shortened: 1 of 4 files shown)


arsing:
There's also [this change](https://github.com/Azure/iot-identity-service/commit/ed4c0e51f0b36cb18dba371898a534711988edee#diff-e3a869ce272d8763377b11e8306bd22d3ba10b24db1fd91180c6701fb6e29658) from the first commit that you need, no?

gordonwang0:
https://github.com/Azure/iot-identity-service/pull/457

arsing:
But if you merge that then it has the change to template.toml and tpmd-config/lib.rs that was reverted in the second commit, and which you've not included here. I thought you were cherry-picking and squashing both in this PR so that the template.toml and tpmd-config/lib.rs changes canceled out.

gordonwang0:
Ok, I've cherry-picked both into this PR.

arsing (approved):

PR #459, opened 2022-09-12
Make userid's for the services configurable at build time

micahl:
This change is to unblock potential development of a snap package.  

Also related is an (out-of-date) change for the `iotedge` CLI.  See https://github.com/Azure/iotedge/pull/6671.

### DIFF
--- a/aziotctl/src/config/apply.rs
+++ b/aziotctl/src/config/apply.rs
@@ -12,11 +12,11 @@ use anyhow::Context;
 
 use aziotctl_common::config as common_config;
 
-#[derive(structopt::StructOpt)]
+#[derive(clap::Args)]
 pub(crate) struct Options {
     /// The path of the config file.
-    #[structopt(
-        short = "-c",
+    #[arg(
+        short = 'c',
         long,
         value_name = "CONFIG",
         default_value = "/etc/aziot/config.toml"
--- a/aziotctl/src/config/apply.rs
+++ b/aziotctl/src/config/apply.rs
@@ -39,22 +39,10 @@ pub(crate) fn run(options: Options) -> anyhow::Result<()> {
     // Otherwise, tell the user to re-run as root.
     // When run in a snap expect the four users to be prefixed with `snap_`.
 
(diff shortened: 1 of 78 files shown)


aziotctl/src/config/apply.rs
```diff
@@ -37,10 +37,24 @@ pub(crate) fn run(options: Options) -> anyhow::Result<()> {
     // So when running as root, get the four users appropriately.
     // Otherwise, if this is a debug build, fall back to using the current user.
     // Otherwise, tell the user to re-run as root.
-    let aziotks_user = crate::internal::common::get_system_user("aziotks")?;
-    let aziotcs_user = crate::internal::common::get_system_user("aziotcs")?;
-    let aziotid_user = crate::internal::common::get_system_user("aziotid")?;
-    let aziottpm_user = crate::internal::common::get_system_user("aziottpm")?;
+    // When run in a snap expect the four users to be prefixed with `snap_`.
+
+    let (aziotks_user, aziotcs_user, aziotid_user, aziottpm_user) = if std::env::var("SNAP").is_ok()
```
arsing, line 42: This is a runtime env var and not a compile-time one?

micahl: Hmm... I was thinking that doing something compile-time might require a larger change although I can see that from an overall cleanliness perspective it provides benefits.  Let me take a closer look as perhaps it wouldn't be as involved as I assume.

arsing: You'd replace `if std::env::var("SNAP").is_some()` with `if option_env!("SNAP").is_some()`

The question is more about: Will the snap package be reusing the binaries of the regular Ubuntu build? Is that env var automatically defined for snaps at runtime?

micahl: The binaries for a regular Ubuntu build should work as well.  Yes, the env var is automatically injected for a snap at runtime by the snapd process.

However, I'm making a change to this PR to have the userid be set at build time.  A motivating reason is to unblock further development by making it simple to change the userid's (e.g. root vs snap_* vs the aziot??) without requiring a code change every time.

micahl:
Closing in favor of https://github.com/Azure/iot-identity-service/pull/479

PR #469, opened 2022-10-19
Fix E2E test failures

micahl:
Errors addressed:
1. contrib/debian/rules wasn't setting the correct engines libdir for openssl 1.1
2. scheduled E2E test runs on the release/1.x branch don't know about Ubuntu 22.04 yet and complain it isn't known and error out
3. accidentally changed the vm_image for Ubuntu 20.04 in a prior change.  updating it and 22.04 to the most recent VM image for each respectively.

Sidenote: Including a change to have the E2E tests targeting the release/1.4 branch instead of release/1.3.

### DIFF
--- a/.github/workflows/e2e-tests-scheduled.yaml
+++ b/.github/workflows/e2e-tests-scheduled.yaml
@@ -69,13 +69,16 @@ jobs:
         # - 'platform:el8'
         - 'ubuntu:18.04'
         - 'ubuntu:20.04'
-        # The scheduled E2E test in the release/1.x branch fails until the changes are merged. Disabling for now.
-        #- 'ubuntu:22.04'
+        - 'ubuntu:22.04'
         test_name:
         - 'manual-symmetric-key'
         - 'manual-x509'
         - 'dps-symmetric-key'
         - 'dps-x509'
+        exclude:
+          # The scheduled E2E test in the release/1.x branch would fail until the changes are merged. Excluding for now.
+          - os: 'ubuntu:22.04'
+            branch: 'release/1.4'
 
       max-parallel: 10
 
--- a/.github/workflows/e2e-tests-scheduled.yaml
+++ b/.github/workflows/e2e-tests-scheduled.yaml
@@ -27,7 +27,7 @@ jobs:
       matrix:
         branch:
         - 'main'
-        - 'release/1.3'
+        - 'release/1.4'
 
       max-parallel: 10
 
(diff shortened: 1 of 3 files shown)


ci/e2e-tests/test-run.sh
```diff
@@ -603,14 +603,14 @@ case "$OS" in
         # az vm image list --all \
         #     --publisher 'Canonical' --offer '0001-com-ubuntu-server-focal' --sku '20' \
         #     --query "[?publisher == 'Canonical' && offer == '0001-com-ubuntu-server-focal'].{ sku: sku, version: version, urn: urn }" --output table
-        vm_image='Canonical:0001-com-ubuntu-server-focal:20_04-lts-gen2:20.04.202209130'
+        vm_image='Canonical:0001-com-ubuntu-server-focal:20_04-lts-gen2:20.04.202210180'
```
damonbarry, line 606: Question for @arsing: Why are we pinning to a particular version of the VM image, as opposed to using `...:latest`?

arsing: Ah, TIL.

https://learn.microsoft.com/en-us/azure/virtual-machines/linux/cli-ps-findimage

>These values can be passed individually or as an image URN, combining the values separated by the colon (:). For example: Publisher:Offer:Sku:Version. You can replace the version number in the URN with latest to use the latest version of the image.

Yes, we should switch to that then.

micahl: Updated all of them to use latest

.github/workflows/e2e-tests-scheduled.yaml
```diff
@@ -69,7 +69,8 @@ jobs:
         # - 'platform:el8'
         - 'ubuntu:18.04'
         - 'ubuntu:20.04'
-        - 'ubuntu:22.04'
+        # The scheduled E2E test in the release/1.x branch fails until the changes are merged. Disabling for now.
```
arsing, line 72: You can use https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#excluding-matrix-configurations to remove the release/1.4 - 22.04 combination instead of disabling it entirely.

arsing (changes requested):

.github/workflows/e2e-tests-scheduled.yaml
micahl: Re-using your TLA... TIL.  :)

arsing (approved):

damonbarry (approved):

micahl:
@arsing or @damonbarry there's a failure in the `Azure-IoT-IIS SDL` that references 'guardian'.  What's that?

PR #471, opened 2022-10-24
[aziotctl] Map parser outputs instead of generating help manually

onalante-msft:
This produces better help text than using the `FromStr` value parser since it also prints the possible values when invalid input is provided.

### DIFF
--- a/aziotctl/src/system.rs
+++ b/aziotctl/src/system.rs
@@ -3,6 +3,7 @@
 use std::ffi::{OsStr, OsString};
 
 use anyhow::{anyhow, Result};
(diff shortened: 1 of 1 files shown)


aziotctl/src/system.rs
```diff
@@ -3,6 +3,7 @@
 use std::ffi::{OsStr, OsString};
 
 use anyhow::{anyhow, Result};
+use clap::builder::TypedValueParser;
```
arsing, line 6:
```suggestion
use clap::builder::{PossibleValuesParser, TypedValueParser};
```

arsing (approved):

PR #472, opened 2022-10-25
[certd] Fix panic when truncating certificate CN

onalante-msft:
[`String::truncate`](https://doc.rust-lang.org/std/string/struct.String.html#method.truncate) panics when the truncation point is not at a valid UTF-8 character boundary.  To fix this, take at most 64 characters from the parsed CN string before setting the X509 CN field.

### DIFF
--- a/cert/aziot-certd-config/src/lib.rs
+++ b/cert/aziot-certd-config/src/lib.rs
@@ -375,6 +375,32 @@ mod tests {
     use super::*;
 
     #[test]
+    fn common_name_truncation_does_not_panic() {
+        let long_name_unicode_boundary =
+            "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaá";
+        let expected_truncation =
+            "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa";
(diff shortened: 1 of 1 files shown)


cert/aziot-certd-config/src/lib.rs
```diff
@@ -231,22 +231,43 @@ impl std::convert::TryFrom<&CertSubject> for openssl::x509::X509Name {
         // X.509 requires CNs to be shorter than 64 characters.
         const CN_MAX_LENGTH: usize = 64;
 
+        // TODO(rustup): feature(round_char_boundary)
+        // Ref: https://doc.rust-lang.org/std/string/struct.String.html#method.floor_char_boundary
+        fn truncate_cn_length(cn: &str) -> &str {
+            if CN_MAX_LENGTH >= cn.len() {
```
arsing, line 237:
```suggestion
            if cn.len() <= CN_MAX_LENGTH {
```

arsing (approved):
I wonder if it'd be more appropriate to split at graphemes (`unicode-segmentation`) instead of codepoints.

onalante-msft:
> I wonder if it'd be more appropriate to split at graphemes (`unicode-segmentation`) instead of codepoints.

You are correct: see the bottom of page 124 [here](https://www.rfc-editor.org/rfc/rfc5280). This is not a particularly performance-sensitive code path, so I will just do a `.chars().take(64).collect::<String>()`.

PR #475, opened 2022-10-27
[identityd] Fix certificate subject propagation

onalante-msft:
Certificate subject configuration options for EST-issued certificates were not being propagated to identityd, which meant that CSRs to issue a certificate would not be generated with the configured certificate subject.  Add additional configuration options to identityd to receive a CSR subject configuration and adjust `aziotctl config apply` to hydrate these options with the certificate issuance options extracted for `certd`.

### DIFF
--- a/identity/aziot-identityd-config/src/lib.rs
+++ b/identity/aziot-identityd-config/src/lib.rs
@@ -8,11 +8,14 @@
     clippy::large_enum_variant
 )]
 
+use std::collections::BTreeMap;
 use std::io::ErrorKind;
+
+use serde::{Deserialize, Serialize};
 
 mod check;
 
-#[derive(Clone, Debug, serde::Deserialize, serde::Serialize)]
+#[derive(Clone, Debug, Deserialize, Serialize)]
 pub struct Settings {
     pub hostname: String,
 
@@ -74,9 +77,9 @@ impl Settings {
 
 pub fn deserialize_cloud_timeout<'de, D>(deserializer: D) -> Result<u64, D::Error>
 where
-    D: serde::de::Deserializer<'de>,
+    D: serde::Deserializer<'de>,
 {
-    let result: u64 = serde::Deserialize::deserialize(deserializer)?;
+    let result: u64 = Deserialize::deserialize(deserializer)?;
 
     if result == 0 {
         return Err(serde::de::Error::custom(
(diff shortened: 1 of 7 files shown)


onalante-msft:
Testing done by inspecting generated CSR with this patch:
```diff
diff --git a/identity/aziot-identityd/src/lib.rs b/identity/aziot-identityd/src/lib.rs
index 4ec2a7d..df7169c 100644
--- a/identity/aziot-identityd/src/lib.rs
+++ b/identity/aziot-identityd/src/lib.rs
@@ -838,7 +838,9 @@ pub(crate) fn create_csr(
     csr.set_pubkey(public_key)?;
     csr.sign(private_key, openssl::hash::MessageDigest::sha256())?;
 
-    csr.build().to_pem()
+    let pem = csr.build().to_pem()?;
+    std::fs::write("/var/lib/aziot/identityd/csr.pem", &pem).unwrap();
+    Ok(pem)
 }
 
 pub struct SettingsAuthenticator {

```

gordonwang0 (approved):

identity/aziot-identityd-config/src/lib.rs
```diff
@@ -124,31 +125,98 @@ pub enum ManualAuthMethod {
...
 }
 
-#[derive(Clone, Debug, serde::Deserialize, serde::Serialize)]
-#[serde(tag = "method")]
-#[serde(rename_all = "lowercase")]
+#[derive(Clone, Debug, Deserialize, Serialize)]
+#[serde(untagged)]
+pub enum CsrSubject {
+    CommonName(String),
+    #[serde(deserialize_with = "subject_from_key_value")]
```
arsing, line 137: Do you need the `deserialize_with` ? With:

```rust
#[serde(rename = "CN")]
cn: String,

#[serde(flatten, skip_serializing_if = "BTreeMap::is_empty")]
rest: BTreeMap<String, String>,
```

... I get:

```
$ toml::from_str::<CsrSubject>("CN = \"foo\"\nDN = \"bar\"")

Ok(Subject { cn: "foo", rest: {"DN": "bar"} })

$ toml::from_str::<CsrSubject>("CN = \"foo\"")

Ok(Subject { cn: "foo", rest: {} })
```

arsing (changes requested):

identity/aziot-identityd-config/src/lib.rs
onalante-msft: Not necessarily. We can always uppercase later in the code, but the uppercasing has to happen at some point since OpenSSL expects uppercased field identifiers. I can simplify the deserialization to use rename on the `cn` field and only uppercase the `BTreeMap` keys, though.

arsing: >but the uppercasing has to happen at some point since OpenSSL expects uppercased field identifiers

Oh right, I forgot about that. Makes sense.

arsing (approved):

PR #479, opened 2022-11-04
Snapping: initial pass

alexclewontin:
Here is my initial pass of packaging the identity service as a snap. Dropping here as a draft to socialize and solicit feedback.

This includes @micahl's work on build-time userid settings.

It also includes build-time alternative implementations for service management commands (as systemctl doesn't work well in a snapped environment). It doesn't handle the `aziotctl system status` case as that doesn't map quite as cleanly, so more work will be required there.

~~This does not account for the possibility of #478, except for a little of the groundwork which is in snapcraft.yaml but commented out.~~ Incorporated those changes here.


micahl (changes requested):

aziotctl/aziotctl-common/src/system/set_log_level.rs
```diff
@@ -35,3 +41,19 @@ fn write_log_level_file(service: &str, level: log::Level) -> Result<()> {
 
     Ok(())
 }
+
+#[cfg(feature = "snapctl")]
+pub fn set_log_level(_services: &[&ServiceDefinition], level: log::Level) -> Result<()> {
+    let result = Command::new("snapctl")
```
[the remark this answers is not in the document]
alexclewontin: Effectively yes. 

All this actually does is set a arbitrary key-value pair in the snap environment (log-level isn't a special key in any way). The real work is done in contrib/snap/command-chain/launch-wrapper.sh where it reads the value, and translates it into an env var before the service gets launched. That wrapper then gets wrapped around all of the services (see command-chain in snapcraft.yaml)

I'd like a double check that this functionality is sufficient, as my initial read was the original version was just setting environment variables via a systemd service drop in file, but definitely could be wrong there.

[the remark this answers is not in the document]
micahl: Ah, I missed that in the launch-wrapper.sh.

snap/hooks/configure
```diff
@@ -0,0 +1,28 @@
...
+	toml_kvp "homedir_path" "$SNAP_COMMON/libaziot_keys_homedir"
+
+	toml_new_section "provisioning"
+	toml_kvp "source" "$(snapctl get provisioning.source)"
+	# TODO: don't print this if not set
+	toml_kvp "connection_string" "$(snapctl get provisioning.connection-string)"
+
+} > $SNAP_DATA/etc/aziot/config.toml
+
+$SNAP/bin/aziotctl config apply
```
micahl, line 28: Is there any way that someone could pass as a config value a file path (or the full content) of an existing config.toml and then this configure hook could just call `$SNAP/bin/aziotctl config apply -c 'path to file'`?

alexclewontin: We could certainly implement something along those lines. 

`snap set azure-iot-identity config-file-path='path to file'` -> `$SNAP/bin/aziotctl config apply -c 'path to file'` would be a bit limited. We could do it as long as `'path to file'` is in the users home directory, but currently there isn't a very good way to give access to arbitrary file paths (and plus for unattended install/config you'd have to somehow seed that file onto the device, and there aren't really users and home directories on ubuntu core, and so this path gets a bit messy for cases besides development work).

More interesting is the idea of doing something more like `snap set azure-iot-identity raw-config-file="$(cat 'path to config file')" -> `snapctl get raw-config-file > /tmp/cfg.toml ; $SNAP/bin/aziotctl config apply -c /tmp/cfg.toml`. Then, on unattended installs, you could also just pass in the raw file contents as a multiline default config value.

As a note, this config hook is probably the least developed part of this whole effort so far, and definitely needs quite a bit more work.

aziotctl/aziotctl-common/src/system/restart.rs
```diff
@@ -35,3 +40,33 @@ fn start(name: &str) -> Result<()> {
...
+#[cfg(feature = "snapctl")]
+pub fn restart(services: &[&ServiceDefinition]) -> Result<()> {
+    let snap_instance_name = match std::env::var("SNAP_INSTANCE_NAME") {
+        Ok(snap_instance_name) => snap_instance_name,
+        Err(_) => {
+            std::env::var("SNAP_NAME").expect("snapctl must be used within the context of a snap")
+        }
+    };
+
+    print!("Restarting {} services...", snap_instance_name);
```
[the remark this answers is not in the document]
alexclewontin: Sure. I think the only thing is that since the systemctl version just calls the stop & start functions, whereas this makes a single call to snapctl restart, it still won't have parity because the non-snapctl version will have stop & start messages whereas this will have no messages. Let me know if this is ok/what you think the best way to handle it would be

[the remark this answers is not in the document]
micahl: Does snapctl have an equivalent start/stop that could be used?

snap/hooks/configure
micahl: I think we should start with the raw-config-file option to make setting the config as much of a pass-through operation as possible.  Otherwise, I can see a lot of effort going into ensuring that the range of allowed options in the config.toml can also be expressed via `snap set azure-iot-identity ...` not to mention that whenever we decided to add/remove config options then this configure script would also need to be modified.

micahl (changes requested):

aziotctl/aziotctl-common/src/system/restart.rs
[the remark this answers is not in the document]
alexclewontin: It does, and I used that to implement the stop command. I can change this to just call stop then start

Makefile
```diff
@@ -20,6 +26,15 @@ ARCH =
 
 INSTALL_PRESET = true
 
+# Enable special features based on a specific package format
+# '' => none, 'snap' => snap package
+PACKAGE_FORMAT ?=
```
micahl, line 31: We currently create more than one package format for the same functionality (deb vs rpm).  Since this flag affects what runtime features of the platform are expected I'm not sure if PACKAGE_FORMAT adequately captures it.   How about...
```suggestion
PLATFORM_FEATURES ?=
```

micahl (approved):
Added one minor comment.  Otherwise, lgtm.  @arsing, @gordonwang0, and @onalante-msft if we don't hear otherwise before Wednesday then I'll assume everyone else is also OK with these proposed changes.  Some similar build-time changes will likely be needed for IoT Edge (e.g., make the OS userid configurable, use snapctl instead of systemctl in the iotedge cli, etc.).

Makefile
```diff
@@ -309,6 +332,9 @@ deb: dist
 	# Copy package files
 	cp -R contrib/debian /tmp/aziot-identity-service-$(PACKAGE_VERSION)/
 	sed -i -e 's/@version@/$(PACKAGE_VERSION)/g; s/@release@/$(PACKAGE_RELEASE)/g' /tmp/aziot-identity-service-$(PACKAGE_VERSION)/debian/changelog
+	sed -i -e 's/@user_aziotid@/$(USER_AZIOTID)/g; s/@user_aziotks@/$(USER_AZIOTKS)/g; s/@user_aziotcs@/$(USER_AZIOTCS)/g; s/@user_aziottpm@/$(USER_AZIOTTPM)/g' /tmp/aziot-identity-service-$(PACKAGE_VERSION)/debian/postinst
+	sed -i -e 's/@user_aziotid@/$(USER_AZIOTID)/g; s/@user_aziotks@/$(USER_AZIOTKS)/g; s/@user_aziotcs@/$(USER_AZIOTCS)/g; s/@user_aziottpm@/$(USER_AZIOTTPM)/g' 's|@socket_dir@|$(SOCKET_DIR)|g' /tmp/aziot-identity-service-$(PACKAGE_VERSION)/debian/postrm
```
micahl, line 336:
```suggestion
	sed -i -e 's/@user_aziotid@/$(USER_AZIOTID)/g; s/@user_aziotks@/$(USER_AZIOTKS)/g; s/@user_aziotcs@/$(USER_AZIOTCS)/g; s/@user_aziottpm@/$(USER_AZIOTTPM)/g s|@socket_dir@|$(SOCKET_DIR)|g' /tmp/aziot-identity-service-$(PACKAGE_VERSION)/debian/postrm
```

Makefile
```diff
@@ -309,6 +332,9 @@ deb: dist
 	# Copy package files
 	cp -R contrib/debian /tmp/aziot-identity-service-$(PACKAGE_VERSION)/
 	sed -i -e 's/@version@/$(PACKAGE_VERSION)/g; s/@release@/$(PACKAGE_RELEASE)/g' /tmp/aziot-identity-service-$(PACKAGE_VERSION)/debian/changelog
+	sed -i -e 's/@user_aziotid@/$(USER_AZIOTID)/g; s/@user_aziotks@/$(USER_AZIOTKS)/g; s/@user_aziotcs@/$(USER_AZIOTCS)/g; s/@user_aziottpm@/$(USER_AZIOTTPM)/g' /tmp/aziot-identity-service-$(PACKAGE_VERSION)/debian/postinst
+	sed -i -e 's/@user_aziotid@/$(USER_AZIOTID)/g; s/@user_aziotks@/$(USER_AZIOTKS)/g; s/@user_aziotcs@/$(USER_AZIOTCS)/g; s/@user_aziottpm@/$(USER_AZIOTTPM)/g s|@socket_dir@|$(SOCKET_DIR)|g' /tmp/aziot-identity-service-$(PACKAGE_VERSION)/debian/postrm
```
micahl, line 336:
```suggestion
	sed -i -e 's/@user_aziotid@/$(USER_AZIOTID)/g; s/@user_aziotks@/$(USER_AZIOTKS)/g; s/@user_aziotcs@/$(USER_AZIOTCS)/g; s/@user_aziottpm@/$(USER_AZIOTTPM)/g; s|@socket_dir@|$(SOCKET_DIR)|g' /tmp/aziot-identity-service-$(PACKAGE_VERSION)/debian/postrm
```

snap/snapcraft.yaml
```diff
@@ -0,0 +1,158 @@
...
+      ln -vrfs $SNAPCRAFT_PRIME/libexec/aziot-identity-service/aziotd $SNAPCRAFT_PRIME/libexec/aziot-keyd
+      ln -vrfs $SNAPCRAFT_PRIME/libexec/aziot-identity-service/aziotd $SNAPCRAFT_PRIME/libexec/aziot-tpmd
+  command-chain:
+    plugin: dump
+    source: ./contrib
+    stage: [ snap/command-chain ]
+
+apps:
+  aziotctl:
+    command: bin/aziotctl
```
micahl, line 87:
```suggestion
    command: usr/bin/aziotctl
```

alexclewontin: So the way the snap is constructed now, `bin/aziotctl` is the correct path, because I explicitly relocated `usr/bin/` to `bin/`. There's no good reason to do this except my personal preference is to keep snap directory structure as flat as possible (one bin, one lib, one shared, avoid nesting things under usr) but I should emphasize that that is my personal preference, not established best practices. So if to keep things consistent you'd like to put aziotctl under usr/bin within the snap, that would be fine with me

aziotctl/aziotctl-common/src/system/mod.rs
```diff
@@ -37,6 +38,28 @@ pub const SERVICE_DEFINITIONS: &[&ServiceDefinition] = &[
     },
 ];
 
+// Ordering is not important here, we defer it to snapd
+#[cfg(feature = "snapctl")]
+pub const SERVICE_DEFINITIONS: &[&ServiceDefinition] = &[
+    &ServiceDefinition {
+        service: "identityd",
```
micahl, line 45: @alexclewontin I was looking into why the `system logs` command for iotedge and aziotctl is always empty output.  This service name gets used in the aztioctl-common/src/system/system_logs.rs when it shells out to `journalctl -u <name>`.  I noticed that the names for services in a snap end up being `snap.<package>.<service>`.  Does that sound right to you and is that durable or would it be considered an implementation detail for snaps on which the code cannot depend?  I tried a local change with those service names and found the `system logs` command now worked as expected.  

Iterating... I'll have some proposed changes later.

alexclewontin:
Closing in favor of PR(s) against the feature/snapping branch (i.e. https://github.com/Azure/iot-identity-service/pull/520)

PR #480, opened 2022-11-04
Add RHEL 9

micahl:


micahl:
@damonbarry here's a draft PR that adds RHEL 9 support, including tests, to the identity service.

damonbarry (approved):

Issue #481, opened 2022-11-08
cargo audit finding in cert-renewal


JanZachmann:
Hi everybody

Running cargo audit points out the following vulnerability:
```
Crate:         time
Version:       0.1.44
Title:         Potential segfault in the time crate
Date:          2020-11-18
ID:            RUSTSEC-2020-0071
URL:           https://rustsec.org/advisories/RUSTSEC-2020-0071
Solution:      Upgrade to >=0.2.23
Dependency tree: 
time 0.1.44
```
which is introduced here:
https://github.com/Azure/iot-identity-service/blob/bc310c7b2254486ff0a0e3fcc4977b8487868b04/cert/cert-renewal/Cargo.toml#L9

A PR fixing the issue can be found here: https://github.com/Azure/iot-identity-service/pull/482

arsing:
That CVE has been discussed to death already. The tl;dr is:

1. chrono does not use the code path in time 0.1 that triggers the error.

2. chrono independently has the same bug as the one in time 0.1, which is a different CVE.

3. The CVE in (2) is only encountered when using chrono's local time API, not when using its UTC time API.

And, since we only use chrono's UTC time API, there is no code path in i-i-s that is affected.

PR #482, opened 2022-11-08
fixed https://github.com/Azure/iot-identity-service/issues/481

JanZachmann:


arsing:
See https://github.com/Azure/iot-identity-service/issues/481#issuecomment-1306752146

Also, chrono is used by multiple crates in the repo and this PR only replaces one of them, so this doesn't achieve what you want anyway.

PR #494, opened 2022-12-14
Make permission check more informative

jlian:
Helps customers self-serve unblock when encountering issues like https://github.com/MicrosoftDocs/azure-docs/issues/102589 and https://github.com/Azure/iotedge/issues/6723

String update only

### DIFF
--- a/aziotctl/aziotctl-common/src/config/mod.rs
+++ b/aziotctl/aziotctl-common/src/config/mod.rs
@@ -133,15 +133,11 @@ pub fn check_readable(mut path: &Path, user: &User, fix: bool) -> anyhow::Result
                 }
             } else {
                 return Err(anyhow::anyhow!(
-                    "{} is not readable by user {} (uid {}, gid {}), try sudo chown {}:{} {} && sudo chmod 700 {}",
+                    "{} is not readable by user {} (uid {}, gid {}), see https://aka.ms/iotedge/cert-permissions",
                     path.display(),
                     user.name,
                     user.uid,
-                    user.gid,
-                    user.name,
-                    user.name,
-                    path.display(),
-                    path.display()
+                    user.gid
                 ));
             }
         }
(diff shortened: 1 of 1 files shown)


aziotctl/aziotctl-common/src/config/mod.rs
```diff
@@ -133,11 +133,15 @@ pub fn check_readable(mut path: &Path, user: &User, fix: bool) -> anyhow::Result
                 }
             } else {
                 return Err(anyhow::anyhow!(
-                    "{} is not readable by user {} (uid {}, gid {})",
+                    "{} is not readable by user {} (uid {}, gid {}), try sudo chown {}:{} {} && sudo chmod 700 {}",
```
arsing, line 136: The suggestion to make it `700` is wrong.

jlian: What is the correct value? I took it from an email from @gordonwang0 

image

arsing: From the email you quoted:

>Since those are **directories**, the permissions should be **755 and 700**

>The **files** in the directories should be **644 and 600** though.

1. Your change is proposing 700 for files.

1. Your change is not proposing the other permissions that would be appropriate.

arsing: So, for (1) you'll of course need the `if is_directory` check.

For (2) you'll need to add an `is_secret: bool` parameter to `check_readable`, and then update all the callers to pass in `true` or `false`. Looking at the code, two of the callers are `ReadKeyPairs::inner_execute` which should pass `true` and `CertificateValidity::new` which should pass false. The third is `create_dir_all` which itself is called from many other places, so you'll need to update them all as well. It'll also have callers in the iotedge repo so you'll have to fix those callers too when you update the dep there.

jlian: I see. What about

```suggestion
                    "{} is not readable by user {} (uid {}, gid {}), see https://aka.ms/iotedge/directory-permissions",
```

And then we just put all the info in the docs?

jlian: Updated

varunpuranik (approved):

arsing (approved):

PR #502, opened 2023-01-24
Add exponential backoff for IotHub throttle

lfitchett:
Adds exponential backoff to identityd. This has caused several icms when many devices are updated at once. Previously the devices would attempt to connect to hub every 10-20 seconds, causing the throttle to persist forever. 

Now devices will go into an exponential backoff when throttled. This only affects 429 (too many requests) errors, all other errors use the old logic. 

As a result of the backoff change, the edged timeout has been increased to 10 minutes.

This also moves the request payload download inside the timeout. This has never been an issue before, but it is more correct.

### DIFF
--- a/aziotctl/config/unix/template.toml
+++ b/aziotctl/config/unix/template.toml
@@ -29,9 +29,13 @@
 # cloud_retries controls how many times a request may be retried should it fail.
 # The client will always send at least one attempt, so its value will be the number
 # of retries after the first attempt should that fail (i.e. cloud_retries = 2
-# means that the client will make a total of 3 attempts).
+# means that the client will make a total of 3 attempts). 
 #
-# cloud_timeout_sec = 10
+# cloud_timeout_sec has a minimum of 60s to allow hub to throttle requests. 
+# If a request is throttled, it will enter an exponentual backoff with 5 retries instead
+# of using the configured value. The configured value is used for all other errors.
+#
+# cloud_timeout_sec = 60
 # cloud_retries = 1
 
 # ==============================================================================
--- a/aziotctl/config/unix/template.toml
+++ b/aziotctl/config/unix/template.toml
@@ -31,11 +31,11 @@
 # of retries after the first attempt should that fail (i.e. cloud_retries = 2
 # means that the client will make a total of 3 attempts). 
 #
-# cloud_timeout_sec has a minimum of 60s to allow hub to throttle requests. 
-# If a request is throttled, it will enter an exponential backoff with 5 retries instead
+# cloud_timeout_sec has a minimum of 70s to allow hub to throttle requests. 
+# If a request is throttled, it will enter an exponential backoff with 4 retries instead
 # of using the configured value. The configured value is used for all other errors.
 #
-# cloud_timeout_sec = 60
+# cloud_timeout_sec = 70
 # cloud_retries = 1
 
 # ==============================================================================
--- a/aziotctl/config/unix/template.toml
+++ b/aziotctl/config/unix/template.toml
@@ -32,7 +32,7 @@
 # means that the client will make a total of 3 attempts). 
 #
 # cloud_timeout_sec has a minimum of 60s to allow hub to throttle requests. 
-# If a request is throttled, it will enter an exponentual backoff with 5 retries instead
+# If a request is throttled, it will enter an exponential backoff with 5 retries instead
 # of using the configured value. The configured value is used for all other errors.
 #
 # cloud_timeout_sec = 60
--- a/http-common/src/request.rs
+++ b/http-common/src/request.rs
@@ -465,7 +465,7 @@ mod tests {
             });
         }
 
-        tokio::time::sleep(Duration::MAX).await;
+        std::future::pending::<()>().await;
     }
 
     async fn query_hub(i: &str) -> Result<(), Box<Error>> {
--- a/http-common/src/request.rs
+++ b/http-common/src/request.rs
@@ -2,6 +2,8 @@
 
 use std::io::{Error, ErrorKind};
 
+use crate::backoff::DEFAULT_BACKOFF;
+
 const CONTENT_TYPE_JSON: &str = "application/json";
 
 pub struct HttpRequest<TBody, TConnector> {
@@ -161,9 +163,8 @@ where
         let client: hyper::Client<_, hyper::Body> = hyper::Client::builder().build(self.connector);
 
         let mut current_attempt = 1;
-        let retry_limit = self.retries + 1;
 
-        let response = loop {
+        loop {
             let mut request = hyper::Request::builder()
                 .method(&self.method)
                 .uri(&self.uri);
@@ -186,61 +187,99 @@ where
                 .body(request_body)
                 .expect("cannot fail to create request");
 
-            let err = match tokio::time::timeout(self.timeout, client.request(request)).await {
-                Ok(response) => {
-                    match response {
-                        Ok(response) => break response,
-                        Err(err) => {
-                            if err.is_connect() {
-                                // Network error.
-                                std::io::Error::new(std::io::ErrorKind::NotConnected, err)
-                            } else {
-                                std::io::Error::new(std::io::ErrorKind::Other, err)
+            let (err, backoff_exponential) =
+                match tokio::time::timeout(self.timeout, client.request(request)).await {
+                    Ok(response) => {
+                        match response {
+                            Ok(response) => {
+                                let (
+                                    http::response::Parts {
+                                        status: response_status,
+                                        headers: response_headers,
+                                        ..
+                                    },
+                                    response_body,
+                                ) = response.into_parts();
+
+                                // Make sure to download body inside the timeout
+                                let response_body = if has_response_body {
+                                    let response_body = hyper::body::to_bytes(response_body)
+                                        .await
+                                        .map_err(|err| Error::new(ErrorKind::Other, err))?;
+
+                                    Some(response_body)
+                                } else {
+                                    None
+                                };
+
+                                // if response throttled, go into exponential backoff
+                                if response_status == http::StatusCode::TOO_MANY_REQUESTS {
+                                    (
+                                        std::io::Error::new(
+                                            std::io::ErrorKind::Other,
+                                            "429: Too many requests",
+                                        ),
+                                        true,
+                                    )
+                                } else {
+                                    // Return results
+                                    return Ok((response_status, response_headers, response_body));
+                                }
+                            }
+                            Err(err) => {
+                                if err.is_connect() {
+                                    // Network error.
+                                    (
+                                        std::io::Error::new(std::io::ErrorKind::NotConnected, err),
+                                        false,
+                                    )
+                                } else {
+                                    (std::io::Error::new(std::io::ErrorKind::Other, err), false)
+                                }
                             }
                         }
                     }
+
+                    Err(timeout) => (timeout.into(), false),
+                };
+
+            if backoff_exponential {
+                if let Some(backoff_duration) =
+                    DEFAULT_BACKOFF.get_backoff_duration(current_attempt)
+                {
+                    log::warn!(
+                        "HTTP request throttled (attempt {} of {}). Sleeping for {} seconds.",
+                        current_attempt,
+                        DEFAULT_BACKOFF.max_retries() + 1,
+                        backoff_duration.as_secs()
+                    );
+                    tokio::time::sleep(backoff_duration).await;
+                } else {
+                    log::warn!(
+                        "Final HTTP request throttled (attempt {} of {}).",
+                        current_attempt,
+                        DEFAULT_BACKOFF.max_retries() + 1,
+                    );
+                    return Err(err);
+                }
+            } else {
+                log::warn!(
+                    "Failed to send HTTP request (attempt {} of {}): {}",
+                    current_attempt,
+                    self.retries + 1,
+                    err
+                );
+
+                if current_attempt > self.retries {
+                    return Err(err);
                 }
 
-                Err(timeout) => timeout.into(),
-            };
-
-            log::warn!(
-                "Failed to send HTTP request (attempt {} of {}): {}",
-                current_attempt,
-                retry_limit,
-                err
-            );
-
-            if current_attempt == retry_limit {
-                return Err(err);
+                // Wait a short time between failed requests.
+                tokio::time::sleep(tokio::time::Duration::from_secs(3)).await;
             }
 
             current_attempt += 1;
-
-            // Wait a short time between failed requests.
-            tokio::time::sleep(tokio::time::Duration::from_secs(3)).await;
-        };
-
-        let (
-            http::response::Parts {
-                status: response_status,
-                headers: response_headers,
-                ..
-            },
-            response_body,
-        ) = response.into_parts();
-
-        let response_body = if has_response_body {
-            let response_body = hyper::body::to_bytes(response_body)
-                .await
-                .map_err(|err| Error::new(ErrorKind::Other, err))?;
-
-            Some(response_body)
-        } else {
-            None
-        };
-
-        Ok((response_status, response_headers, response_body))
+        }
     }
 }
 
--- a/http-common/src/request.rs
+++ b/http-common/src/request.rs
@@ -339,11 +339,14 @@ impl HttpResponse {
 ///
 /// To run this test, first create a new device in a hub. Then use the az cli to generate a device token:
 /// `az iot hub generate-sas-token --hub-name "your-hub" --device-id "your-device-id"`
-/// 
-/// Then simply set the HUB_HOSTNAME, DEVICE_ID, and SAS_TOKEN variables and run the test using
+///
+/// Then simply set the `HUB_HOSTNAME`, `DEVICE_ID`, and `SAS_TOKEN` variables and run the test using
 /// `RUST_LOG=info cargo test test_backoff -- --nocapture --ignored`
-/// 
-/// A successful run should print lots of throttle warnings, but never error. The throttle warnings should have some jitter:
+///
+/// A successful run should print lots of throttle warnings, but never error. The throttle warnings should have some jitter.
+/// It will never return and must be manually canceled.
+///
+/// Example output:
 /// ```
 /// [2023-01-31T05:03:18Z WARN  http_common::request] HTTP request throttled (attempt 1 of 5). Sleeping for 66 seconds.
 /// Finished request 2201 (57) in 9.593452176
@@ -438,7 +441,7 @@ mod tests {
     #[ignore]
     async fn test_backoff_manual() {
         if HUB_HOSTNAME == "your-hubname-here.azurecr.io" {
-            return
+            return;
         }
 
         env_logger::init();
--- a/http-common/src/request.rs
+++ b/http-common/src/request.rs
@@ -187,63 +187,63 @@ where
                 .body(request_body)
                 .expect("cannot fail to create request");
 
-            let (err, backoff_exponential) =
-                match tokio::time::timeout(self.timeout, client.request(request)).await {
+            let mut is_timeout = false;
+            let response_future = async {
+                match client.request(request).await {
                     Ok(response) => {
-                        match response {
-                            Ok(response) => {
-                                let (
-                                    http::response::Parts {
-                                        status: response_status,
-                                        headers: response_headers,
-                                        ..
-                                    },
-                                    response_body,
-                                ) = response.into_parts();
+                        let (
+                            http::response::Parts {
+                                status: response_status,
+                                headers: response_headers,
+                                ..
+                            },
+                            response_body,
+                        ) = response.into_parts();
 
-                                // Make sure to download body inside the timeout
-                                let response_body = if has_response_body {
-                                    let response_body = hyper::body::to_bytes(response_body)
-                                        .await
-                                        .map_err(|err| Error::new(ErrorKind::Other, err))?;
+                        // Make sure to download body inside the timeout
+                        let response_body = if has_response_body {
+                            let response_body = hyper::body::to_bytes(response_body)
+                                .await
+                                .map_err(|err| Error::new(ErrorKind::Other, err))?;
 
-                                    Some(response_body)
-                                } else {
-                                    None
-                                };
+                            Some(response_body)
+                        } else {
+                            None
+                        };
 
-                                // if response throttled, go into exponential backoff
-                                if response_status == http::StatusCode::TOO_MANY_REQUESTS {
-                                    (
-                                        std::io::Error::new(
-                                            std::io::ErrorKind::Other,
-                                            "429: Too many requests",
-                                        ),
-                                        true,
-                                    )
-                                } else {
-                                    // Return results
-                                    return Ok((response_status, response_headers, response_body));
-                                }
-                            }
-                            Err(err) => {
-                                if err.is_connect() {
-                                    // Network error.
-                                    (
-                                        std::io::Error::new(std::io::ErrorKind::NotConnected, err),
-                                        false,
-                                    )
-                                } else {
-                                    (std::io::Error::new(std::io::ErrorKind::Other, err), false)
-                                }
-                            }
+                        // if response throttled, go into exponential backoff
+                        if response_status == http::StatusCode::TOO_MANY_REQUESTS {
+                            is_timeout = true;
+                            Err(std::io::Error::new(
+                                std::io::ErrorKind::Other,
+                                "429: Too many requests",
+                            ))
+                        } else {
+                            // Return results
+                            Ok((response_status, response_headers, response_body))
                         }
                     }
+                    Err(err) => {
+                        if err.is_connect() {
+                            // Network error.
+                            Err(std::io::Error::new(std::io::ErrorKind::NotConnected, err))
+                        } else {
+                            Err(std::io::Error::new(std::io::ErrorKind::Other, err))
+                        }
+                    }
+                }
+            };
 
-                    Err(timeout) => (timeout.into(), false),
-                };
+            let err = match tokio::time::timeout(self.timeout, response_future).await {
+                Ok(response) => match response {
+                    Ok(response) => return Ok(response),
+                    Err(err) => err,
+                },
 
-            if backoff_exponential {
+                Err(timeout) => timeout.into(),
+            };
+
+            if is_timeout {
                 if let Some(backoff_duration) =
                     DEFAULT_BACKOFF.get_backoff_duration(current_attempt)
                 {
--- a/http-common/src/request.rs
+++ b/http-common/src/request.rs
@@ -331,3 +331,176 @@ impl HttpResponse {
         }
     }
 }
+
+/// Using this test:
+///
+/// This test can only be run manually. It will send thousands of concurrent tests to iothub to induce a throttle.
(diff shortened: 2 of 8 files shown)


identity/aziot-identityd-config/src/lib.rs
```diff
@@ -81,9 +81,9 @@ where
 {
     let result: u64 = Deserialize::deserialize(deserializer)?;
 
-    if result == 0 {
+    if result < 60 {
         return Err(serde::de::Error::custom(
-            "cloud_timeout_sec must be greater than 0",
+            "cloud_timeout_sec must be greater than 60 seconds",
```
onalante-msft, line 86:
```suggestion
            "cloud_timeout_sec must be at least 60 seconds",
```

http-common/src/request.rs
```diff
@@ -292,3 +331,179 @@ impl HttpResponse {
...
+                    if let Err(e) = query_hub(&request_num).await {
+                        println!("Error: {}", e);
+                    }
+
+                    tokio::time::sleep(Duration::from_millis(1)).await;
+                }
+            });
+        }
+
+        tokio::time::sleep(Duration::MAX).await;
```
onalante-msft, line 468:
```suggestion
        std::future::pending::<()>().await;
```

http-common/src/request.rs
```diff
@@ -186,61 +187,99 @@ where
...
+                                let (
+                                    http::response::Parts {
+                                        status: response_status,
+                                        headers: response_headers,
+                                        ..
+                                    },
+                                    response_body,
+                                ) = response.into_parts();
+
+                                // Make sure to download body inside the timeout
```
onalante-msft, line 204: By "inside the timeout", do you mean gating the entire `to_bytes` within the timeout or just the initial connection as done here?

onalante-msft (changes requested):
Classical exponential backoff would be simpler, I think.

<details><summary>Patch</summary>

```diff
diff --git a/Cargo.lock b/Cargo.lock
index 650835c..cce9c70 100644
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2160,12 +2160,6 @@ version = "0.3.26"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6ac9a59f73473f1b8d852421e59e64809f025994837ef743615c6d0c5b305160"
 
-[[package]]
-name = "ppv-lite86"
-version = "0.2.17"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5b40af805b3121feab8a3c29f04d8ad262fa8e0561883e7653e024ae4479e6de"
-
 [[package]]
 name = "proc-macro-error"
 version = "1.0.4"
@@ -2214,18 +2208,6 @@ version = "0.8.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "34af8d1a0e25924bc5b7c43c079c942339d8f0a8b57c39049bef581b46327404"
 dependencies = [
- "libc",
- "rand_chacha",
- "rand_core",
-]
-
-[[package]]
-name = "rand_chacha"
-version = "0.3.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e6c10a63a0fa32252be49d21e7709d4d4baf8d231c2dbce1eaa8141b9b127d88"
-dependencies = [
- "ppv-lite86",
  "rand_core",
 ]
 
diff --git a/http-common/Cargo.toml b/http-common/Cargo.toml
index 21b3fd6..5adf28f 100644
--- a/http-common/Cargo.toml
+++ b/http-common/Cargo.toml
@@ -19,7 +19,7 @@ nix = "0.24"
 openssl = { version = "0.10" }
 openssl-sys = { version = "0.9" }
 percent-encoding = "2"
-rand = "0.8.5"
+rand = { version = "0.8", features = ["getrandom"], default-features = false }
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 tokio = { version = "1", features = ["net", "rt-multi-thread", "sync", "time"] }
diff --git a/http-common/src/backoff.rs b/http-common/src/backoff.rs
deleted file mode 100644
index 86b8ca0..0000000
--- a/http-common/src/backoff.rs
+++ /dev/null
@@ -1,53 +0,0 @@
-// Copyright (c) Microsoft. All rights reserved.
-
-use std::time::Duration;
-
-use rand::Rng;
-
-pub const DEFAULT_BACKOFF: Backoff<4> = Backoff {
-    pattern: [
-        BackoffInstance::from_secs(60, 10),
-        BackoffInstance::from_secs(120, 20),
-        BackoffInstance::from_secs(180, 30),
-        BackoffInstance::from_secs(300, 30),
-    ],
-};
-
-pub struct Backoff<const N: usize> {
-    pattern: [BackoffInstance; N],
-}
-
-impl<const N: usize> Backoff<N> {
-    #[allow(clippy::unused_self, clippy::cast_possible_truncation)]
-    pub fn max_retries(&self) -> u32 {
-        N as u32
-    }
-
-    /// Computes backoff for current try. Returns None if no retry attempts left
-    pub fn get_backoff_duration(&self, current_attempt: u32) -> Option<Duration> {
-        self.pattern
-            .get(current_attempt as usize - 1)
-            .map(BackoffInstance::backoff_duration)
-    }
-}
-
-pub struct BackoffInstance {
-    duration: Duration,
-    max_jitter: Duration,
-}
-
-impl BackoffInstance {
-    const fn from_secs(duration: u64, max_jitter: u64) -> Self {
-        Self {
-            duration: Duration::from_secs(duration),
-            max_jitter: Duration::from_secs(max_jitter),
-        }
-    }
-
-    fn backoff_duration(&self) -> Duration {
-        let mut rng = rand::thread_rng();
-        let jitter_multiple = rng.gen_range(0.0..1.0);
-
-        self.duration + self.max_jitter.mul_f32(jitter_multiple)
-    }
-}
diff --git a/http-common/src/lib.rs b/http-common/src/lib.rs
index cf683f6..1e35f1e 100644
--- a/http-common/src/lib.rs
+++ b/http-common/src/lib.rs
@@ -30,8 +30,6 @@ pub use request::HttpRequest;
 
 pub mod server;
 
-mod backoff;
-
 mod uid;
 
 /// Ref <https://url.spec.whatwg.org/#path-percent-encode-set>
diff --git a/http-common/src/request.rs b/http-common/src/request.rs
index 7b57ae3..efde9ee 100644
--- a/http-common/src/request.rs
+++ b/http-common/src/request.rs
@@ -1,10 +1,15 @@
 // Copyright (c) Microsoft. All rights reserved.
 
 use std::io::{Error, ErrorKind};
+use std::time::Duration;
 
-use crate::backoff::DEFAULT_BACKOFF;
+use rand::Rng;
 
 const CONTENT_TYPE_JSON: &str = "application/json";
+const MAX_REQUESTS: i32 = 4;
+const INITIAL_BACKOFF: Duration = Duration::from_secs(60);
+const BACKOFF_FACTOR: f32 = 1.5;
+const JITTER_RATIO: f32 = 0.25;
 
 pub struct HttpRequest<TBody, TConnector> {
     connector: TConnector,
@@ -162,8 +167,7 @@ where
     > {
         let client: hyper::Client<_, hyper::Body> = hyper::Client::builder().build(self.connector);
 
-        let mut current_attempt = 1;
-
+        let mut attempt = 0;
         loop {
             let mut request = hyper::Request::builder()
                 .method(&self.method)
@@ -187,98 +191,60 @@ where
                 .body(request_body)
                 .expect("cannot fail to create request");
 
-            let (err, backoff_exponential) =
-                match tokio::time::timeout(self.timeout, client.request(request)).await {
-                    Ok(response) => {
-                        match response {
-                            Ok(response) => {
-                                let (
-                                    http::response::Parts {
-                                        status: response_status,
-                                        headers: response_headers,
-                                        ..
-                                    },
-                                    response_body,
-                                ) = response.into_parts();
-
-                                // Make sure to download body inside the timeout
-                                let response_body = if has_response_body {
-                                    let response_body = hyper::body::to_bytes(response_body)
-                                        .await
-                                        .map_err(|err| Error::new(ErrorKind::Other, err))?;
-
-                                    Some(response_body)
-                                } else {
-                                    None
-                                };
-
-                                // if response throttled, go into exponential backoff
-                                if response_status == http::StatusCode::TOO_MANY_REQUESTS {
-                                    (
-                                        std::io::Error::new(
-                                            std::io::ErrorKind::Other,
-                                            "429: Too many requests",
-                                        ),
-                                        true,
-                                    )
-                                } else {
-                                    // Return results
-                                    return Ok((response_status, response_headers, response_body));
-                                }
-                            }
-                            Err(err) => {
-                                if err.is_connect() {
-                                    // Network error.
-                                    (
-                                        std::io::Error::new(std::io::ErrorKind::NotConnected, err),
-                                        false,
-                                    )
-                                } else {
-                                    (std::io::Error::new(std::io::ErrorKind::Other, err), false)
-                                }
-                            }
-                        }
-                    }
-
-                    Err(timeout) => (timeout.into(), false),
-                };
-
-            if backoff_exponential {
-                if let Some(backoff_duration) =
-                    DEFAULT_BACKOFF.get_backoff_duration(current_attempt)
-                {
-                    log::warn!(
-                        "HTTP request throttled (attempt {} of {}). Sleeping for {} seconds.",
-                        current_attempt,
-                        DEFAULT_BACKOFF.max_retries() + 1,
-                        backoff_duration.as_secs()
-                    );
-                    tokio::time::sleep(backoff_duration).await;
-                } else {
-                    log::warn!(
-                        "Final HTTP request throttled (attempt {} of {}).",
-                        current_attempt,
-                        DEFAULT_BACKOFF.max_retries() + 1,
-                    );
-                    return Err(err);
-                }
-            } else {
+            if attempt > 0 {
+                let base_backoff = INITIAL_BACKOFF.mul_f32(BACKOFF_FACTOR.powi(attempt));
+                let jitter = INITIAL_BACKOFF
+                    .mul_f32(rand::rngs::OsRng.gen_range(-JITTER_RATIO..JITTER_RATIO));
+                let wait = base_backoff + jitter;
                 log::warn!(
-                    "Failed to send HTTP request (attempt {} of {}): {}",
-                    current_attempt,
-                    self.retries + 1,
-                    err
+                    "HTTP request throttled (attempt {} of {}). Retrying in {} seconds.",
+                    attempt,
+                    MAX_REQUESTS,
+                    wait.as_secs()
                 );
+                tokio::time::sleep(wait).await;
+            }
 
-                if current_attempt > self.retries {
-                    return Err(err);
+            let response = match tokio::time::timeout(self.timeout, client.request(request)).await {
+                Ok(Ok(response)) => response,
+                Ok(Err(err)) => {
+                    break Err(Error::new(
+                        if err.is_connect() {
+                            ErrorKind::NotConnected
+                        } else {
+                            ErrorKind::Other
+                        },
+                        err,
+                    ));
+                }
+                Err(timeout) => {
+                    break Err(timeout.into());
                 }
+            };
 
-                // Wait a short time between failed requests.
-                tokio::time::sleep(tokio::time::Duration::from_secs(3)).await;
-            }
+            let (
+                http::response::Parts {
+                    status, headers, ..
+                },
+                response_body,
+            ) = response.into_parts();
+            let body = if has_response_body {
+                Some(
+                    hyper::body::to_bytes(response_body)
+                        .await
+                        .map_err(|err| Error::new(ErrorKind::Other, err))?,
+                )
+            } else {
+                None
+            };
 
-            current_attempt += 1;
+            if status != http::StatusCode::TOO_MANY_REQUESTS {
+                break Ok((status, headers, body));
+            } else if attempt < MAX_REQUESTS {
+                attempt += 1;
+            } else {
+                break Err(Error::new(ErrorKind::Other, "too many requests"));
+            }
         }
     }
 }

```
</summary>

http-common/src/request.rs
lfitchett: What do you mean by only the initial connection? Before, only the request was inside the timeout, and the   `hyper::body::to_bytes(response_body).await` was outside. I just moved it inside. Now the body download is inside the timeout

onalante-msft (approved):
Approved. Alternative implementation suggested, but not to the point of outright requesting changes.

onalante-msft:
Alternative implementation:

<details><summary>PATCH</summary>

```diff
diff --git a/Cargo.lock b/Cargo.lock
index 650835c..cce9c70 100644
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2160,12 +2160,6 @@ version = "0.3.26"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "6ac9a59f73473f1b8d852421e59e64809f025994837ef743615c6d0c5b305160"
 
-[[package]]
-name = "ppv-lite86"
-version = "0.2.17"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "5b40af805b3121feab8a3c29f04d8ad262fa8e0561883e7653e024ae4479e6de"
-
 [[package]]
 name = "proc-macro-error"
 version = "1.0.4"
@@ -2214,18 +2208,6 @@ version = "0.8.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "34af8d1a0e25924bc5b7c43c079c942339d8f0a8b57c39049bef581b46327404"
 dependencies = [
- "libc",
- "rand_chacha",
- "rand_core",
-]
-
-[[package]]
-name = "rand_chacha"
-version = "0.3.1"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "e6c10a63a0fa32252be49d21e7709d4d4baf8d231c2dbce1eaa8141b9b127d88"
-dependencies = [
- "ppv-lite86",
  "rand_core",
 ]
 
diff --git a/http-common/Cargo.toml b/http-common/Cargo.toml
index 21b3fd6..5adf28f 100644
--- a/http-common/Cargo.toml
+++ b/http-common/Cargo.toml
@@ -19,7 +19,7 @@ nix = "0.24"
 openssl = { version = "0.10" }
 openssl-sys = { version = "0.9" }
 percent-encoding = "2"
-rand = "0.8.5"
+rand = { version = "0.8", features = ["getrandom"], default-features = false }
 serde = { version = "1", features = ["derive"] }
 serde_json = "1"
 tokio = { version = "1", features = ["net", "rt-multi-thread", "sync", "time"] }
diff --git a/http-common/src/backoff.rs b/http-common/src/backoff.rs
deleted file mode 100644
index 86b8ca0..0000000
--- a/http-common/src/backoff.rs
+++ /dev/null
@@ -1,53 +0,0 @@
-// Copyright (c) Microsoft. All rights reserved.
-
-use std::time::Duration;
-
-use rand::Rng;
-
-pub const DEFAULT_BACKOFF: Backoff<4> = Backoff {
-    pattern: [
-        BackoffInstance::from_secs(60, 10),
-        BackoffInstance::from_secs(120, 20),
-        BackoffInstance::from_secs(180, 30),
-        BackoffInstance::from_secs(300, 30),
-    ],
-};
-
-pub struct Backoff<const N: usize> {
-    pattern: [BackoffInstance; N],
-}
-
-impl<const N: usize> Backoff<N> {
-    #[allow(clippy::unused_self, clippy::cast_possible_truncation)]
-    pub fn max_retries(&self) -> u32 {
-        N as u32
-    }
-
-    /// Computes backoff for current try. Returns None if no retry attempts left
-    pub fn get_backoff_duration(&self, current_attempt: u32) -> Option<Duration> {
-        self.pattern
-            .get(current_attempt as usize - 1)
-            .map(BackoffInstance::backoff_duration)
-    }
-}
-
-pub struct BackoffInstance {
-    duration: Duration,
-    max_jitter: Duration,
-}
-
-impl BackoffInstance {
-    const fn from_secs(duration: u64, max_jitter: u64) -> Self {
-        Self {
-            duration: Duration::from_secs(duration),
-            max_jitter: Duration::from_secs(max_jitter),
-        }
-    }
-
-    fn backoff_duration(&self) -> Duration {
-        let mut rng = rand::thread_rng();
-        let jitter_multiple = rng.gen_range(0.0..1.0);
-
-        self.duration + self.max_jitter.mul_f32(jitter_multiple)
-    }
-}
diff --git a/http-common/src/lib.rs b/http-common/src/lib.rs
index cf683f6..1e35f1e 100644
--- a/http-common/src/lib.rs
+++ b/http-common/src/lib.rs
@@ -30,8 +30,6 @@ pub use request::HttpRequest;
 
 pub mod server;
 
-mod backoff;
-
 mod uid;
 
 /// Ref <https://url.spec.whatwg.org/#path-percent-encode-set>
diff --git a/http-common/src/request.rs b/http-common/src/request.rs
index 7b57ae3..cd2a05e 100644
--- a/http-common/src/request.rs
+++ b/http-common/src/request.rs
@@ -1,10 +1,14 @@
 // Copyright (c) Microsoft. All rights reserved.
 
 use std::io::{Error, ErrorKind};
+use std::time::Duration;
 
-use crate::backoff::DEFAULT_BACKOFF;
+use rand::Rng;
 
-const CONTENT_TYPE_JSON: &str = "application/json";
+const CONTENT_TYPE_JSON: http::HeaderValue = http::HeaderValue::from_static("application/json");
+const INITIAL_BACKOFF: Duration = Duration::from_secs(60);
+const BACKOFF_FACTOR: f32 = 1.5;
+const JITTER_RATIO: f32 = 0.25;
 
 pub struct HttpRequest<TBody, TConnector> {
     connector: TConnector,
@@ -89,7 +93,7 @@ where
 
     pub fn add_header(
         &mut self,
-        name: hyper::header::HeaderName,
+        name: http::header::HeaderName,
         value: &str,
     ) -> Result<(), Error> {
         let value = http::HeaderValue::from_str(value)
@@ -125,18 +129,7 @@ where
     pub async fn json_response(self) -> Result<HttpResponse, Error> {
         let (status, headers, body) = self.process_request(true).await?;
 
-        let is_json_response = if let Some(content_type) = headers.get(hyper::header::CONTENT_TYPE)
-        {
-            let content_type = content_type
-                .to_str()
-                .map_err(|err| Error::new(ErrorKind::InvalidData, err))?;
-
-            content_type.contains(CONTENT_TYPE_JSON)
-        } else {
-            false
-        };
-
-        if !is_json_response {
+        if headers.get(http::header::CONTENT_TYPE) != Some(&CONTENT_TYPE_JSON) {
             return Err(Error::new(
                 ErrorKind::InvalidData,
                 "invalid Content-Type; expected JSON",
@@ -162,15 +155,19 @@ where
     > {
         let client: hyper::Client<_, hyper::Body> = hyper::Client::builder().build(self.connector);
 
-        let mut current_attempt = 1;
-
-        loop {
+        let attempt_limit = self.retries + 1;
+        let mut attempt = 0;
+        let mut backoff = 0;
+        let result = loop {
+            attempt += 1;
             let mut request = hyper::Request::builder()
                 .method(&self.method)
                 .uri(&self.uri);
+            let headers = request.headers_mut()
+                .expect("cannot fail to get request headers");
 
             let request_body = if let Some(body) = &self.body {
-                request = request.header(hyper::header::CONTENT_TYPE, CONTENT_TYPE_JSON);
+                headers.insert(http::header::CONTENT_TYPE, CONTENT_TYPE_JSON);
 
                 serde_json::to_vec(body)
                     .expect("cannot fail to serialize request")
@@ -179,107 +176,68 @@ where
                 hyper::Body::default()
             };
 
-            for (header_name, header_value) in &self.headers {
-                request = request.header(header_name, header_value);
-            }
+            headers.extend(self.headers.clone());
 
             let request = request
                 .body(request_body)
                 .expect("cannot fail to create request");
 
-            let (err, backoff_exponential) =
-                match tokio::time::timeout(self.timeout, client.request(request)).await {
-                    Ok(response) => {
-                        match response {
-                            Ok(response) => {
-                                let (
-                                    http::response::Parts {
-                                        status: response_status,
-                                        headers: response_headers,
-                                        ..
-                                    },
-                                    response_body,
-                                ) = response.into_parts();
-
-                                // Make sure to download body inside the timeout
-                                let response_body = if has_response_body {
-                                    let response_body = hyper::body::to_bytes(response_body)
-                                        .await
-                                        .map_err(|err| Error::new(ErrorKind::Other, err))?;
-
-                                    Some(response_body)
-                                } else {
-                                    None
-                                };
-
-                                // if response throttled, go into exponential backoff
-                                if response_status == http::StatusCode::TOO_MANY_REQUESTS {
-                                    (
-                                        std::io::Error::new(
-                                            std::io::ErrorKind::Other,
-                                            "429: Too many requests",
-                                        ),
-                                        true,
-                                    )
-                                } else {
-                                    // Return results
-                                    return Ok((response_status, response_headers, response_body));
-                                }
-                            }
-                            Err(err) => {
-                                if err.is_connect() {
-                                    // Network error.
-                                    (
-                                        std::io::Error::new(std::io::ErrorKind::NotConnected, err),
-                                        false,
-                                    )
-                                } else {
-                                    (std::io::Error::new(std::io::ErrorKind::Other, err), false)
-                                }
-                            }
-                        }
+            let mut wait = Duration::from_secs(3);
+            let err = match tokio::time::timeout(self.timeout, client.request(request)).await {
+                Ok(Ok(response)) => {
+                    if response.status() == http::StatusCode::TOO_MANY_REQUESTS {
+                        backoff += 1;
+                        let base_backoff = INITIAL_BACKOFF.mul_f32(BACKOFF_FACTOR.powi(backoff));
+                        let wait_ratio =
+                            1.0 + rand::rngs::OsRng.gen_range(-JITTER_RATIO..JITTER_RATIO);
+                        wait = base_backoff.mul_f32(wait_ratio);
+                        Error::new(ErrorKind::Other, "too many requests")
+                    } else {
+                        break Ok(response);
                     }
-
-                    Err(timeout) => (timeout.into(), false),
-                };
-
-            if backoff_exponential {
-                if let Some(backoff_duration) =
-                    DEFAULT_BACKOFF.get_backoff_duration(current_attempt)
-                {
-                    log::warn!(
-                        "HTTP request throttled (attempt {} of {}). Sleeping for {} seconds.",
-                        current_attempt,
-                        DEFAULT_BACKOFF.max_retries() + 1,
-                        backoff_duration.as_secs()
-                    );
-                    tokio::time::sleep(backoff_duration).await;
-                } else {
-                    log::warn!(
-                        "Final HTTP request throttled (attempt {} of {}).",
-                        current_attempt,
-                        DEFAULT_BACKOFF.max_retries() + 1,
-                    );
-                    return Err(err);
                 }
-            } else {
+                Ok(Err(err)) => Error::new(
+                    if err.is_connect() {
+                        ErrorKind::NotConnected
+                    } else {
+                        ErrorKind::Other
+                    },
+                    err,
+                ),
+                Err(err) => Error::from(err),
+            };
+
+            if attempt < attempt_limit {
                 log::warn!(
-                    "Failed to send HTTP request (attempt {} of {}): {}",
-                    current_attempt,
-                    self.retries + 1,
-                    err
+                    "Failed to send request (attempt {} of {}): {}. Retrying in {} seconds.",
+                    attempt,
+                    attempt_limit,
+                    err,
+                    wait.as_secs(),
                 );
-
-                if current_attempt > self.retries {
-                    return Err(err);
-                }
-
                 // Wait a short time between failed requests.
-                tokio::time::sleep(tokio::time::Duration::from_secs(3)).await;
+                tokio::time::sleep(wait).await
+            } else {
+                break Err(err);
             }
+        };
 
-            current_attempt += 1;
-        }
+        let (
+            http::response::Parts {
+                status, headers, ..
+            },
+            response_body,
+        ) = result?.into_parts();
+        let body = if has_response_body {
+            Some(
+                hyper::body::to_bytes(response_body)
+                    .await
+                    .map_err(|err| Error::new(ErrorKind::Other, err))?,
+            )
+        } else {
+            None
+        };
+        Ok((status, headers, body))
     }
 }
 
```
</details>

http-common/src/request.rs
onalante-msft: The body download time is not limited by the timeout with this, though. You would need to wrap the function contents up to the body download with `async { .. }` for that.

lfitchett: moved body download into timeout

identity/aziot-identityd-config/src/lib.rs
```diff
@@ -57,7 +57,7 @@ pub struct Settings {
 
 impl Settings {
     pub fn default_cloud_timeout() -> u64 {
-        10
+        60
```
arsing, line 60: We might want it to be a bit higher since 60s is the request timeout on the server side, and the client does extra work of DNS lookup, establishing the TCP connection and sending the request that we include in the request timeout on our end.

70s should be good enough.

http-common/src/request.rs
```diff
@@ -186,61 +187,99 @@ where
...
-                    match response {
-                        Ok(response) => break response,
-                        Err(err) => {
-                            if err.is_connect() {
-                                // Network error.
-                                std::io::Error::new(std::io::ErrorKind::NotConnected, err)
-                            } else {
-                                std::io::Error::new(std::io::ErrorKind::Other, err)
-                            }
+            let mut is_timeout = false;
```
arsing, line 190: `is_timeout` -> `is_throttled`

arsing (changes requested):

identity/aziot-identityd-config/src/lib.rs
lfitchett: In testing the server never took longer than 45 seconds, so the 60s was already a small buffer

arsing: Yes, that's what Viktor observed as well. But the 60s timeout is what they document, so we should take them at their word.

aziotctl/config/unix/template.toml
```diff
@@ -29,9 +29,13 @@
 # cloud_retries controls how many times a request may be retried should it fail.
 # The client will always send at least one attempt, so its value will be the number
 # of retries after the first attempt should that fail (i.e. cloud_retries = 2
-# means that the client will make a total of 3 attempts).
+# means that the client will make a total of 3 attempts). 
 #
-# cloud_timeout_sec = 10
+# cloud_timeout_sec has a minimum of 60s to allow hub to throttle requests. 
+# If a request is throttled, it will enter an exponentual backoff with 5 retries instead
```
gordonwang0, line 35:
```suggestion
# If a request is throttled, it will enter an exponential backoff with 5 retries instead
```

http-common/src/request.rs
```diff
@@ -186,61 +187,99 @@ where
...
+                                .await
+                                .map_err(|err| Error::new(ErrorKind::Other, err))?;
+
+                            Some(response_body)
+                        } else {
+                            None
+                        };
+
+                        // if response throttled, go into exponential backoff
+                        if response_status == http::StatusCode::TOO_MANY_REQUESTS {
```
gordonwang0, line 215: I think we also discussed using exponential backoff for certain 5xx errors, right?

gordonwang0 (changes requested):

http-common/src/request.rs
lfitchett: We discussed it, but decided that if hub gives a 500 error we should do the normal retry and fail quickly.

identity/aziot-identityd-config/src/lib.rs
```diff
@@ -81,9 +81,9 @@ where
 {
     let result: u64 = Deserialize::deserialize(deserializer)?;
 
-    if result == 0 {
+    if result < 70 {
         return Err(serde::de::Error::custom(
-            "cloud_timeout_sec must be greater than 0",
+            "cloud_timeout_sec must be at least 60 seconds",
```
arsing, line 86: 70

arsing, line 86:
```suggestion
            "cloud_timeout_sec must be at least 70 seconds",
```

arsing (approved):

PR #512, opened 2023-02-09
Add Ubuntu 22.04 as a target

damonbarry:
Cherry-pick f66c155113cce5089d398376d16b206f7973d6f6, 0c244f97c190fdd5bf6cd71e4f9ece96aacc025a, bd148d4ffdd4759bea77529e6f4a98f0403b400f


.github/workflows/e2e-tests-scheduled.yaml
```diff
@@ -69,11 +69,16 @@ jobs:
...
+        - 'ubuntu:22.04'
         test_name:
         - 'manual-symmetric-key'
         - 'manual-x509'
         - 'dps-symmetric-key'
         - 'dps-x509'
+        exclude:
+          # The scheduled E2E test in the release/1.x branch would fail until the changes are merged. Excluding for now.
+          - os: 'ubuntu:22.04'
+            branch: 'release/1.4'
```
damonbarry, line 81: Should this be removed?

contrib/debian/rules
```diff
@@ -27,9 +27,7 @@ PRIVATE_LIBS = /usr/lib/aziot-identity-service
 
 override_dh_auto_build:
 	# Default behavior is `make -j1`, which is not quite right for us.
-	make -j \
-		RELEASE=1 \
-		V=1
```
micahl, line 32: These lines were only removed originally as part of a workaround to perform the build outside of the dpkg-buildpackage step.  When that workaround was subsequently removed these probably should have been added back.

.github/workflows/e2e-tests-scheduled.yaml
micahl: Yes, I think so.  On a related note, I expect there are two of the i-i-s E2E tests that will consistently fail on 22.04 that involve the keys engine.

ci/install-build-deps.sh
```diff
@@ -16,7 +16,7 @@ case "$OS:$ARCH" in
 
         yum install -y centos-release-scl epel-release
         yum install -y \
-            autoconf autoconf-archive automake clang curl devtoolset-9-gcc devtoolset-9-gcc-c++ \
```
micahl, line 19: I don't recall removing this, but perhaps it isn't needed anyway if the build was alread successful in `main`.

ci/install-runtime-deps.sh
```diff
@@ -38,8 +38,9 @@ case "$OS" in
         esac
         ;;
 
-    'debian:10'|'debian:11'|'ubuntu:18.04'|'ubuntu:20.04')
-        # openssl 1.1.0 for Debian 9, 1.1.1 for the others
+    'debian:10'|'debian:11'|'ubuntu:18.04'|'ubuntu:20.04'|'ubuntu:22.04')
+        # openssl 1.1.0 for Debian 9, 1.1.1 for Debian 10/11 and Ubuntu 18.04/20.04
```
micahl, line 42: Minor nit

```suggestion
        # openssl 1.1.1 for Debian 10/11 and Ubuntu 18.04/20.04
```

micahl (changes requested):

.github/workflows/e2e-tests-scheduled.yaml
damonbarry: Removed.

> On a related note, I expect there are two of the i-i-s E2E tests that will consistently fail on 22.04 that involve the keys engine.

I assume there's no action for me to take here?

contrib/debian/rules
damonbarry: Reverted.

ci/install-build-deps.sh
damonbarry: Yeah, and the gates pass here as well...

micahl (approved):
lgtm.  Once we're publishing packages for 22.04 to PMC then I'd plan to update the docs/installation.md file to include 22.04.

PR #516, opened 2023-02-15
Disable X.509 E2E tests on Ubuntu 22.04

gordonwang0:
A bug in openssl 3, which is used in Ubuntu 22.04, causes an error when attempting to use openssl s_client with an engine. For now, disable the X.509 Ubuntu 22.04 E2E test runs so that the E2E tests can pass.

The openssl bug is tracked here: https://github.com/openssl/openssl/issues/20161
Revert this commit once the bug is fixed.

### DIFF
--- a/ci/e2e-tests/test-run.sh
+++ b/ci/e2e-tests/test-run.sh
@@ -880,7 +880,7 @@ ssh -i "$PWD/vm-ssh-key" "aziot@$vm_public_ip" "
         exit 1
     fi
 
-    if [ \"$OS\" != 'ubuntu:22.04' ] || [ \"$TEST_NAME\" != *'-x509'* ]; then
+    if [ \"$OS\" != 'ubuntu:22.04' ] || [[ \"$TEST_NAME\" != *'-x509'* ]]; then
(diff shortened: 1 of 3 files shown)


arsing:
I thought we agreed we would just skip the `get-twin.sh` call.

ci/e2e-tests/test-run.sh
```diff
@@ -880,8 +880,10 @@ ssh -i "$PWD/vm-ssh-key" "aziot@$vm_public_ip" "
         exit 1
     fi
 
-    device_twin=\"\$(~/iothub-get-twin.sh \"\$device_identity\")\"
-    printf 'Device twin: %s\n' \"\$device_twin\" >&2
+    if [ \"$OS\" != 'ubuntu:22.04' ] || [ \"$TEST_NAME\" != *'-x509'* ]; then
```
arsing, line 883: You already have `$test_name`. You don't need a new env var.

gordonwang0: Oh right. Fixed.

arsing (approved):

PR #520, opened 2023-03-13
feature/snapping: fix file permissions issues

alexclewontin:
The aziotctl_common::config::write_file function first changes ownership of a file, then changes the permissions.

Changing the permissions on a file owned by another user requires CAP_FOWNER. While this is not normally a problem for root, the assumption that root ALWAYS has CAP_FOWNER is erroneous (snaps are a counter example).

This commit simply reorders the operation, so that the file is still owned by the creator when the permissions get changed.

NB: afaict this doesn't actually have any meaningful effects on the snapped identity service by itself (because everything runs as root, so no files ever actually change ownership.) It *does* impact iotedge, which leverages this code in a way that breaks in confinement.

### DIFF
--- a/aziotctl/aziotctl-common/src/config/mod.rs
+++ b/aziotctl/aziotctl-common/src/config/mod.rs
@@ -62,13 +62,22 @@ pub fn write_file(
     let path = path.as_ref();
     let path_displayable = path.display();
 
+    // We're about to truncate the file anyway, but it is a little safer to
+    // fully remove it, as a truncating write does not actually reset ownership
+    // or user permissions (which can lead to needing CAP_FOWNER, as then we 
+    // encounter a set_permissions call on a file we may or may not own). We
+    // can just ignore any errors returned here as they either mean a) the file
+    // doesn't exist, in which case great, or b) there is some sort of permission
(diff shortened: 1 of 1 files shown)


arsing:
>It does impact iotedge, which leverages this code in a way that breaks in confinement.

Just to be clear, it breaks the iotedge snap, not regular non-snap iotedge, right? Because any uses of this code in non-snap iotedge would also happen as root.

alexclewontin:
> Just to be clear, it breaks the iotedge snap, not regular non-snap iotedge, right?

The way the code is currently only breaks in snap confinement, correct. In snap confinement, the root user is sometimes subject to DAC in ways that it is not outside of snap confinement. This is an instance of that: root inside snap confinement does not have CAP_FOWNER, but root outside snap confinement *does* have CAP_FOWNER.

The proposed changes should work well either in or outside of confinement

arsing:
One of the things that CI is complaining about is new in this PR, but the others have been there before, and I see commits being made to this feature branch despite the failures. I assume the intent is to fix them up when the branch is ready for review, and for now it's okay to just merge this manually, so doing that.

PR #521, opened 2023-03-21
feature/snapping: hostname reading fix on Ubuntu Core

alexclewontin:
This fixes how the snap reads the device's hostname on Ubuntu Core, and should be more robust in general.


alexclewontin:
Very confused with how github is calculating the diff, the first commit is already merged into feature/snapping

alexclewontin:
Ah good catch. Done

PR #535, opened 2023-06-27
Update cargo dependencies

gordonwang0:
- Update cargo dependencies to resolve CG issues.
- Update toolchain to 1.65 (consistent with main) because some new dependency versions require newer toolchain.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2242,11 +2242,11 @@ dependencies = [
 
 [[package]]
 name = "serde_with"
-version = "2.3.3"
+version = "3.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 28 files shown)


arsing:
You should update the dependency versionreqs in Cargo.toml too. eg we depend on an old version of `serde-with` that continues to pulls in an old version of base64.

arsing:
So just to be clear, you checked all the other dependency versionreqs and serde-with was the only one that was old?

gordonwang0:
Actually, I think I'll revert the serde_with version change. Last time we updated dependencies between major versions, we had some issues due to differences in behavior between major versions. Since 1.4 is supposed to be a stable release, I think we should minimize the amount of changes. (And CG isn't complaining about base64, which is the reason for doing this update.)

Let me know if you disagree.

arsing:
Oh right, I mixed up this one and #536. Keeping this one on the same major versions is fine. What I wrote applies to the main branch PR only.

arsing (approved):

PR #537, opened 2023-07-16
Enable CodeQL in this repo

damonbarry:
CodeQL detects one C# file and four C files in this repository and wants to run analysis on them. This PR enables CodeQL analysis.

For the C code, since CodeQL's autobuild can't figure out our Rust-based build, I added a custom pipeline. I adapted the YAML that GitHub generates for [advanced code scanning](https://docs.github.com/en/code-security/code-scanning/automatically-scanning-your-code-for-vulnerabilities-and-errors/configuring-code-scanning-for-a-repository#configuring-code-scanning-manually).

For C#, the single file in this repo is not production code--it's meant to be run inside an Azure Function during a test--so building it in the new pipeline isn't worth the effort. Instead, I changed the extension on the file. Hopefully CodeQL will pass over it.


arsing:
Your troubles with needing to install the deps manually are because you're building outside a container. Why are you doing that? Is it because the `make` needs to be done in the same "context" as where the `github/codeql-action/init` runs?

If so, I suggest:

Step 1: `github/codeql-action/init@v2`

Step 2: `run: sudo bash -c '. install-build-deps.sh; make'`

Step 3: `github/codeql-action/analyze@v2`

damonbarry:
> Is it because the make needs to be done in the same "context" as where the github/codeql-action/init runs?

Yes, exactly. Thanks, I'll use your suggestion.

PR #549, opened 2023-09-06
When creating an AES PKCS#11 key, test it with a dummy encryption to ensure that the token can perform AES-GCM.

arsing:
It is possible for PKCS#11 implementations to support AES keys only with AES-CBC etc but not with AES-GCM. Before this change, creating the key would succeed, and then encrypting with it would fail, at which point it would be too late.

With this change, we do a dummy encruption with the key just to be sure that it can be used with AES-GCM. If that fails, we continue to the next location, usually the filesystem ie openssl keys.

### DIFF
--- a/key/aziot-keys/src/key.rs
+++ b/key/aziot-keys/src/key.rs
@@ -669,14 +669,14 @@ fn create_inner(
                     .is_ok()
                 {
                     return Ok(());
-                } else {
-                    // Delete the object we just created...
-                    if let Some(object_label) = &uri.object_label {
-                        _ = pkcs11_session.clone().delete_key(object_label);
-                    }
-
-                    // ... and continue to the next location.
                 }
+
+                // Delete the object we just created...
+                if let Some(object_label) = &uri.object_label {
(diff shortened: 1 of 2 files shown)


gordonwang0 (commented):
I feel like this would be confusing to customers who expect key operations through their PKCS#11 library, only to find that all the keys are on the filesystem. Though I realize that the only way to get iotedge to work in this case would be filesystem keys.

Does it make sense to add an iotedge check that will raise a warning when the PKCS#11 library doesn't support the correct algorithm? This would at least provide awareness on why iotedge is using filesystem keys even if configured with a PKCS#11 library.

arsing:
It already used to silently fall back to filesystem if the PKCS#11 impl didn't support AES keys at all. We're just tightening it to also check for AES-GCM operations.

If we want to add a `check` for it it can be a separate work item.

gordonwang0 (approved):
Ok, makes sense. Approved.

damonbarry (approved):

damonbarry (approved):

PR #571, opened 2023-12-06
Fix e2e tests to not run Trusted Launch VMs since some distros don't support it.

arsing:
Azure changed the default for Gen 2 VMs to use Trusted Launch security, but the CentOS 7 and Debian 10 images don't support it so those VMs fail to deploy. This commit makes the tests create Standard VMs again.

Also update the comment describing how to select VM image URNs to account for how we specify the version as `latest` now.


damonbarry:
FYI @nlcamp

PR #577, opened 2023-12-27
Reduce http-common logging noise in logs.

robinmholt-daikin:
Currently, the aziot-edged logs are littered with the request/response data coming from the http-common server crate included from this repo. That logging level is probably more appropriately debug.  Adjust accordingly.

```bash
Dec 27 22:17:58 SOLU424200002 aziot-edged[2546]: 2023-12-27T22:17:58Z [INFO] - <-- GET /modules?api-version=2020-07-07 {"accept": "application/json", "host": "mgmt.sock:80", "connection": "close"}
Dec 27 22:17:58 SOLU424200002 aziot-edged[2546]: 2023-12-27T22:17:58Z [INFO] - --> 200 {"content-type": "application/json"}
Dec 27 22:18:03 SOLU424200002 aziot-edged[2546]: 2023-12-27T22:18:03Z [INFO] - <-- GET /modules?api-version=2020-07-07 {"accept": "application/json", "host": "mgmt.sock:80", "connection": "close"}
Dec 27 22:18:03 SOLU424200002 aziot-edged[2546]: 2023-12-27T22:18:03Z [INFO] - --> 200 {"content-type": "application/json"}
Dec 27 22:18:08 SOLU424200002 aziot-edged[2546]: 2023-12-27T22:18:08Z [INFO] - <-- GET /modules?api-version=2020-07-07 {"accept": "application/json", "host": "mgmt.sock:80", "connection": "close"}
Dec 27 22:18:08 SOLU424200002 aziot-edged[2546]: 2023-12-27T22:18:08Z [INFO] - --> 200 {"content-type": "application/json"}
Dec 27 22:18:13 SOLU424200002 aziot-edged[2546]: 2023-12-27T22:18:13Z [INFO] - <-- GET /modules?api-version=2020-07-07 {"accept": "application/json", "host": "mgmt.sock:80", "connection": "close"}
Dec 27 22:18:13 SOLU424200002 aziot-edged[2546]: 2023-12-27T22:18:13Z [INFO] - --> 200 {"content-type": "application/json"}
Dec 27 22:18:18 SOLU424200002 aziot-edged[2546]: 2023-12-27T22:18:18Z [INFO] - <-- GET /modules?api-version=2020-07-07 {"accept": "application/json", "host": "mgmt.sock:80", "connection": "close"}
Dec 27 22:18:18 SOLU424200002 aziot-edged[2546]: 2023-12-27T22:18:18Z [INFO] - --> 200 {"content-type": "application/json"}
Dec 27 22:18:23 SOLU424200002 aziot-edged[2546]: 2023-12-27T22:18:23Z [INFO] - <-- GET /modules?api-version=2020-07-07 {"accept": "application/json", "host": "mgmt.sock:80", "connection": "close"}
Dec 27 22:18:23 SOLU424200002 aziot-edged[2546]: 2023-12-27T22:18:23Z [INFO] - --> 200 {"content-type": "application/json"}
```

### DIFF
--- a/http-common/src/server.rs
+++ b/http-common/src/server.rs
@@ -234,12 +234,12 @@ macro_rules! make_service {
                 }
 
                 // TODO: When we get distributed tracing, associate these two logs with the tracing ID.
-                log::info!("<-- {:?} {:?} {:?}", req.method(), req.uri(), req.headers());
+                log::debug!("<-- {:?} {:?} {:?}", req.method(), req.uri(), req.headers());
                 let res = call_inner(self, req);
                 Box::pin(async move {
                     let res = res.await;
                     match &res {
-                        Ok(res) => log::info!("--> {:?} {:?}", res.status(), res.headers()),
+                        Ok(res) => log::debug!("--> {:?} {:?}", res.status(), res.headers()),
                         Err(err) => log::error!("-!> {:?}", err),
                     }
                     res


robinmholt-daikin:
This partially addresses https://github.com/Azure/iotedge/issues/5700 which also needs a PR in the iotedge repository to complete in order to completely quiet the logs which is https://github.com/Azure/iotedge/pull/7181

arsing:
Unfortunately almost every customer incident we need to investigate requires these logs, and our users are generally not clueful enough to raise the log level before filing incidents.

So if you want to turn them off you can do so yourself, by adding a service dropin for each service that sets `Environment=AZIOT_LOG=foo=off`, where `foo` is the module path where `make_service!` was invoked. That is, you can create `/etc/systemd/system/aziot-certd.service.d/dont-log-server-requests.conf` with the content:

```ini
[Service]
Environment=AZIOT_LOG=aziot_certd::http=off
```

... `/etc/systemd/system/aziot-edged.service/dont-log-server-requests.conf` with the content:

```ini
[Service]
Environment=AZIOT_LOG=edgelet_http_mgmt=off,edgelet_http_workload=off
```

... and so on.

robinmholt-daikin:
@arsing 

I adapted you edglet_http_mgmt=off to WARN and that works for quieting the aziot-edged.

Do you need these for debug because the error case does not capture the req.{method,uri,headers} at the same log level?  If so, I could just update the PR to add those to the log.error which is what really should be there anyway since having the log level below INFO would lose useful information if an error were to occur.

arsing:
>Do you need these for debug [...]

Yes. Knowing which requests succeed is as important as knowing which ones don't.

>[...] because the error case does not capture the req.{method,uri,headers} at the same log level?

It doesn't matter that the error log doesn't have that info, because that info is in the info log and is thus emitted by default.

robinmholt-daikin:
> It doesn't matter that the error log doesn't have that info, because that info is in the info log and is thus emitted by default.

@arsing - I don't think you are considering what I am asking.  Assume I have done something like set AZIOT_LOG=WARN.  If an error occurs, there will only be the `-|> err` message without the `<--- method, uri, headers` message.

What I am suggesting is there is a general principle that you should not rely upon another log message at a more verbose logging level to produce the information you would need to understand a message.

arsing:
>Assume I have done something like set AZIOT_LOG=WARN. If an error occurs, there will only be the `-|>` err message without the `<--- method, uri, headers` message.

Yes. And since you have demonstrated your cluefulness by overriding the log level, you will hopefully remember to remove the override before you file your issue.

FWIW, I do not consider these logs to be "verbose" in the first place, and I don't see the problem with leaving them enabled. So keep in mind that you're decreasing your and our ability to diagnose any issues you have, and you will have to re-enable all of those logs before you file an issue. Also note that changing the log level requires restarting the service(s), so you have to hope your issue continues to repro after you do that.

>What I am suggesting is there is a general principle that you should not rely upon another log message at a more verbose logging level to produce the information you would need to understand a message.

Sure, you can duplicate the request fields in the response log for the error case. It will become noisier in the default case where info logs are enabled, so hopefully nobody else comes and complains about \*that\*...

robinmholt-daikin:
> FWIW, I do not consider these logs to be "verbose" in the first place, and I don't see the problem with leaving them enabled. So keep in mind that you're decreasing your and our ability to diagnose any issues you have, and you will have to re-enable all of those logs before you file an issue. Also note that changing the log level requires restarting the service(s), so you have to hope your issue continues to repro after you do that.

@arsing, I looked at devices we have had in the field.  These are running the iotedge 1.1 security daemon, et. al.  Their iotedge.service was configured with IOTEDGE_LOG=WARN.  After 6 months on the four Alpha/Beta devices I have access to, we had 3084211, 2548403, 2659896, and 2734078 bytes of log messages over the previous 6 months for an average of 2756647 or 631 bytes per hour.

With one device and these log messages, I have 342624 bytes per hour, so I take you point, it is only 543 times as much log messages so it really is a trivial burden for everybody to bear.

As to the ability to debug without these log messages, doesn't this really beg the question, why not have log messages specific to the transitions into and out of exceptional conditions and not just periodic "This is what is going on now" messages.  The former results in you actually thinking through your code and making the logging effective for the end user to independently investigate while the later means you really need to be an expert in both the normal and abnormal to begin to diagnose a problem.

But, at this point, I can see I am not making any progress and I have a kludge which gets me to nearly where I need.  Hopefully I will have better luck with my other PR and I can leave this mess to you to either sort out or continue to ignore.

Either way, I am going to advise our team to build our own versions of these packages for all distros (we were already doing this for the unsupported distros) and include in our fork this patch.  We will, as a fall-back, leave the log level set to warn for the _http_ services as you indicated.

robinmholt-daikin:
After some more testing, I came to the following.  With the default log level, my logs rotate every approximately 11 hours.  With the IOTEDGE_LOG=WARN, I was getting a rotation at greater than 6 months which is what our security team had required of our design.

To attempt to make a less restrictive version of the logging I have created the following files:
```
# File /etc/systemd/system/aziot-certd.service.d/override.conf
[Service]
Environment=AZIOT_LOG=aziot_certd::http=WARN

# File /etc/systemd/system/aziot-edged.service.d/override.conf
[Service]
Environment=AZIOT_LOG=edgelet_http_mgmt=WARN,edgelet_http_workload=WARN


# File /etc/systemd/system/aziot-identityd.service.d/override.conf
[Service]
Environment=AZIOT_LOG=aziot_identityd::http=WARN

# File /etc/systemd/system/aziot-keyd.service.d/override.conf
[Service]
Environment=AZIOT_LOG=aziot_keyd::http=WARN
```

This does get me more information than I had before which I do find useful.  I have not run very long in this configuration, but will continue to run this way for a while to see where my logs stabilize.

I don't know how to interpret rust code into these config settings, I just used `strings` on the executables and looked for interesting patterns, then set those until the log messages went away.  This is certainly less than ideal, but it is what I am left with.

robinmholt-daikin:
@arsing in https://github.com/Azure/iot-identity-service/pull/577#issuecomment-1871249978 I took a swag at turning down the extraneous logging.  Was this done in the most limiting way possible or could I have specified something differently for the services like aziot-edged to have limited them?

What I am asking is this works for me, but not being a rust expert, I don't know if it is the best way to quiet just these extraneous log messages.

PR #609, opened 2024-04-18
Read latest version from product-versions.json

damonbarry:
The "aziot-version" check in `aziotctl check` reads the latest identity service version from https://github.com/Azure/azure-iotedge/blob/main/latest-aziot-identity-service.json. This JSON file has no ability to accomodate multiple releases (e.g., 1.4 and 1.5). However, https://github.com/Azure/azure-iotedge/blob/main/product-versions.json was recently introduced and, while it has a more complex structure, it is capable of providing information about multiple versions of the same product.

This change updates `aziotctl check` to read the latest identity service version from product-versions.json. Since the logic needs to contend with multiple versions, it now uses the MAJOR.MINOR version from the actual version string to match the expected version. For example, if the user has 1.5.0 installed and 1.5.2 and 1.4.10 are the latest versions, it will match on 1.5 and return 1.5.2 as the expected version. It does not differentiate between the same versions on different channels (e.g., stable vs. LTS); it assumes they are the same (in other words, it assumes product-versions.json will never legitimately show version 1.5.3 in the stable channel and 1.5.2 in the LTS channel).

### DIFF
--- a/aziotctl/src/internal/check/checks/aziot_version.rs
+++ b/aziotctl/src/internal/check/checks/aziot_version.rs
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft. All rights reserved.
 
 use anyhow::{anyhow, Context, Result};
+use semver::Version;
 use serde::{Deserialize, Serialize};
 
 use crate::internal::check::{CheckResult, Checker, CheckerCache, CheckerMeta, CheckerShared};
@@ -33,6 +34,7 @@ impl AziotVersion {
         shared: &CheckerShared,
         cache: &mut CheckerCache,
     ) -> Result<CheckResult> {
+        let actual_version = env!("CARGO_PKG_VERSION");
         let expected_version = if let Some(expected_aziot_version) =
             &shared.cfg.expected_aziot_version
         {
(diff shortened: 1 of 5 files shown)


aziotctl/src/internal/check/checks/aziot_version.rs
```diff
@@ -98,11 +99,40 @@ impl AziotVersion {
...
+                .collect();
+
+            let actual_channel = Version::parse(actual_version)
+                .context("could not parse actual version as semver")?;
+            let expected_version = versions
+                .iter()
+                .find(|version| {
+                    let expected_channel = Version::parse(version)
+                        .context("could not parse expected version as semver")
+                        .unwrap(); // TODO: What's the right error handling here?
```
damonbarry, line 120: I assume unwrap() is not a good option here, but not sure how best to handle this. Advice?

damonbarry:
@arsing @gordonwang0 I had to dust off my Rust skills for this, which were basic to begin with. I'm open to any feedback about idiomatic usage, etc.

aziotctl/src/internal/check/checks/aziot_version.rs
arsing: `.try_find()` is unfortunately not stable, so you can do:

```rust
let versions: Vec<String> = ...;
let parsed_versions =
    versions.iter()
    .map(|version| Ok(Version::parse(version).context("...")?))
    .collect::<Result<Vec<_>>>()?;

...

let expected_version =
    parsed_versions.iter()
    .find(|expected_channel| expected_channel.major == ...);
```

arsing (approved):

aziotctl/src/internal/check/checks/aziot_version.rs
damonbarry: Fixed.

PR #610, opened 2024-04-23
Add support for Debian 12

damonbarry:


arsing (approved):

damonbarry:
@arsing any thoughts on the packages workflow failures for debian 12 armhf and aarch64?

arsing:
We do:

```sh
export PKG_CONFIG_armv7_unknown_linux_gnueabihf="arm-linux-gnueabihf-pkg-config"
```

... for arm32v7, which is correct for all distros including Debian 11:

```
$ dpkg -L pkg-config:armhf | grep '^/usr/bin/'

/usr/bin/arm-unknown-linux-gnueabihf-pkg-config
/usr/bin/pkg-config
```

... but Debian 12 switched the pkg-config impl to `pkgconf`, and it calls the binary by a different name:

```sh
$ dpkg -L pkgconf:armhf | grep '^/usr/bin/'

/usr/bin/arm-linux-gnueabihf-pkg-config
/usr/bin/arm-linux-gnueabihf-pkgconf
/usr/bin/pkg-config
```

So the env var needs to be set to `arm-linux-gnueabihf-pkg-config` instead of `arm-unknown-linux-gnueabihf-pkg-config`

Same for aarch64.

arsing:
(nvm my previous comment, I can't read.)

I don't know where `/usr/bin/arm-linux-gnueabihf-pkg-config` comes from on Debian 11. It seems to not belong to any package but seems to appear automatically if `pkg-config` is installed while armhf architecture is enabled.

```sh
$ dpkg --add-architecture armhf && apt update -y && apt install -y pkg-config

...

$ ls -l /usr/bin/arm-linux-gnueabihf-pkg-config

lrwxrwxrwx 1 root root 34 Apr 23 21:27 /usr/bin/arm-linux-gnueabihf-pkg-config -> /usr/share/pkg-config-crosswrapper

$ dpkg -S /usr/bin/arm-linux-gnueabihf-pkg-config

dpkg-query: no path found matching pattern /usr/bin/arm-linux-gnueabihf-pkg-config
```

... and in fact, installing `pkg-config:armhf` will uninstall the `pkg-config:amd64` one, ie they're not coinstallable.

On Debian 12 it requires installing `pkg-config:armhf` explicitly and it is even coninstallable with `pkg-config:amd64`.

```sh
$ dpkg --add-architecture armhf && apt update -y && apt install -y pkg-config

...

$ ls -l /usr/bin/arm-linux-gnueabihf-pkg-config

ls: cannot access '/usr/bin/arm-linux-gnueabihf-pkg-config': No such file or directory

$ apt install -y pkg-config:armhf

...

$ ls -l /usr/bin/arm-linux-gnueabihf-pkg-config

lrwxrwxrwx 1 root root 7 Jan 22  2023 /usr/bin/arm-linux-gnueabihf-pkg-config -> pkgconf

$ dpkg -S /usr/bin/arm-linux-gnueabihf-pkg-config

pkgconf:armhf: /usr/bin/arm-linux-gnueabihf-pkg-config
```

Same thing for aarch64.

PR #617, opened 2024-06-11
Use managed identity for end-to-end tests

damonbarry:
This change upgrades the scheduled and manual end-to-end test workflows and scripts to login to Azure CLI using a managed identity that has been installed on the customer runner VMs. It eliminates the need to generate a app secret and store it in the workflow.

To test, I ran the manual workflow with these changes and confirmed that it can still create Azure resources.

### DIFF
--- a/docs-dev/e2e-tests.md
+++ b/docs-dev/e2e-tests.md
@@ -53,7 +53,7 @@ Note that the `e2e-tests-scheduled.yaml` workflow only runs in the main `Azure/i
 
 ### Running the script locally
 
-This requires you to have your own Azure subscription. Follow the same steps as the section above to create a resource group and managed identity, except for creating GitHub secrets.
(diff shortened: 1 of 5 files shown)


docs-dev/e2e-tests.md
```diff
@@ -58,7 +53,7 @@ Note that the `e2e-tests-scheduled.yaml` workflow only runs in the main `Azure/i
 
 ### Running the script locally
 
-This requires you to have your own Azure subscription. Follow the same steps as the section above to create a resource group and service principal, except for creating GitHub secrets.
+This requires you to have your own Azure subscription. Follow the same steps as the section above to create a resource group and managed identity, except for creating GitHub secrets.
```
arsing, line 56:
```suggestion
This requires you to have your own Azure subscription and an Azure VM to run the tests from. Follow the same steps as the section above to create a resource group and managed identity, except for creating GitHub secrets. Assign the managed identity to the VM.
```

arsing (approved):

PR #619, opened 2024-06-17
Disable shared key access on test storage account

damonbarry:
We create a temporary storage account for use by an Azure Function in the scheduled and manual end-to-end tests. Per best practices, we want to create it with shared key access disabled. This change updates the Azure CLI command that creates the storage account.


arsing (dismissed):

damonbarry:
Disabling shared key access on the storage account breaks the Azure Functions app we deploy to host a DPS custom allocation policy.

There is apparently a way to run the Azure Functions app with storage keys disabled (using an identity), but it seems to require a labrynth of settings and workarounds. It may take some time to get this all sorted out.

PR #626, opened 2024-08-17
Update snap base image and remove unnecessary config

damonbarry:
To be consistent with the azure-iot-edge snap (see Azure/iotedge#7351), this change updates the azure-iot-identity snap base to core24. It also removes logic from the configure hook that was unnecessarily populating the hostname field in the identityd, certd, keyd, and tpmd config files. Identityd is the only service that defines hostname in its config, and the field will be automatically set by `aziotctl config apply`, which the configure hook already calls.

### DIFF
--- a/snap/hooks/configure
+++ b/snap/hooks/configure
@@ -1,6 +1,5 @@
 #!/bin/sh
 
-
 toml_kvp() {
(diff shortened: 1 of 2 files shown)


nyanzebra (approved):

damonbarry:
FYI @alexclewontin

PR #629, opened 2024-08-22
Fix Cargo tools install in CodeQL workflow

damonbarry:
The `ubuntu-latest` GitHub runner recently updated its version of bindgen, which exposed a problem in our CodeQL workflow. We need to be using the pinned version of bindgen we install, but we were unknowingly using the pre-installed version because our install script was running in a subshell, so any exported variables were discarded.

This change updates the workflow to source the script in the current shell, and as a non-root user to match the way it calls make. The script now detects the presence of the `DISABLE_FOR_CODEQL` variable and skips parts of the script which (1) require root access or (2) assume the presence of submodules (which are intentionally not present for the CodeQL build to avoid analyzing code we don't own).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -90,7 +90,7 @@ SHELL = /bin/bash
 default:
 	# Re-generate aziot-keys.h if necessary
 	set -euo pipefail; \
-	command -v bindgen; bindgen --version \
+	command -v bindgen; bindgen --version; \
 	aziot_keys_h_new="$$(mktemp --tmpdir 'aziot-keys.h.new.XXXXXXXXXX')"; \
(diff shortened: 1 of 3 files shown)


Makefile
```diff
@@ -90,6 +90,7 @@ SHELL = /bin/bash
 default:
 	# Re-generate aziot-keys.h if necessary
 	set -euo pipefail; \
+	command -v bindgen; bindgen --version \
```
arsing, line 93:
```suggestion
	command -v bindgen; bindgen --version; \
```

arsing (approved):

PR #635, opened 2024-09-26
Use Azure feed to resolve crate dependencies

damonbarry:
To comply with security policies, this change updates the cargo projects in this repo to resolve dependencies through a public Azure feed (which upstreams to crates.io), instead of directly from crates.io. Also, this change adds a few missed members to the cargo workspace for consistency.

All packages have been saved to the feed at the version given in Cargo.lock. Whether the pipelines build our code or you build it locally, the Azure feed will be used to download dependency packages. Authentication in not required.

To add/upgrade a package in the feed, you must authenticate with write credentials. Ideally, a simple `cargo login` before `cargo build` would allow you to seamlessly update the feed, but cargo does not currently support optional authentication with fallback to anonymous. In other words, because we allow anonymous access, cargo will not authenticate. Instead, you can use the feed's REST API directly, e.g.,

```bash
package='<package name goes here>'
version='<package version goes here>'

# the user needs to have "Feed and Upstream Reader (Collaborator)" permissions on the feed
az login
auth_header=$(az account get-access-token --query "join(' ', ['Authorization: Bearer', accessToken])" --output tsv)

url="$(curl -sSL 'https://pkgs.dev.azure.com/iotedge/iotedge/_packaging/iotedge_PublicPackages/Cargo/index/config.json' | jq -r '.dl')"
url="${url/\{crate\}/$package}"
url="${url/\{version\}/$v}"

# curl with --max-time of 5 seconds because we don't actually have to download the package, we just need to nudge
# the feed to acquire the package from upstream
curl -sSL --max-time 5 --header "$auth_header" --write-out '%{http_code}\n' "$url"
```

Outside contributors who need to add/update packages can temporarily comment out the changes in .cargo/config.toml during development, then open a PR (with config.toml restored to its original state) for review. Someone with access to the feed will need to update the feed before the PR can be tested and merged.

I updated docs-dev/building.md with a new section (docs-dev/building.md#updating-a-dependency) that explains how to add/upgrade dependencies in the feed.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -4,3 +4,12 @@ USER_AZIOTKS = "aziotks"
 USER_AZIOTCS = "aziotcs"
 USER_AZIOTTPM = "aziottpm"
 SOCKET_DIR = "/run/aziot"
+
+[registries]
+iotedge_PublicPackages = { index = "sparse+https://pkgs.dev.azure.com/iotedge/iotedge/_packaging/iotedge_PublicPackages/Cargo/index/" }
+
+[registry]
+global-credential-providers = ["cargo:token", "cargo:libsecret"]
+
+[source.crates-io]
+replace-with = "iotedge_PublicPackages"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,6 +8,7 @@ members = [
 	"cert/aziot-cert-client-async",
 	"cert/aziot-cert-common",
 	"cert/aziot-cert-common-http",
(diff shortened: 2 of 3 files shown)


arsing:
Maybe check that in as a script?

arsing (approved):

nyanzebra (approved):

damonbarry:
> Maybe check that in as a script?

I added a section called "Updating a dependency" to docs-dev/building.md.

PR #645, opened 2024-12-16
Add content interface to snap for sharing aziot_keys.so

damonbarry:
Our snap does not allow dependent snaps to access aziot_keys.so, so they cannot use our OpenSSL engine library for mTLS when their client certificate's private key is backed by our key store. This change adds a content interface for the $SNAP/lib/engines-3 folder, where aziot_keys.so lives.

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -73,7 +73,7 @@ parts:
         ARCH=$CRAFT_ARCH_BUILD_FOR \
         RELEASE=1 \
         VENDOR_LIBTSS=0 \
-        OPENSSL_ENGINES_DIR="$SNAP/lib/engines-3" \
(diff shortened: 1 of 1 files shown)


snap/snapcraft.yaml
```diff
@@ -164,6 +164,11 @@ slots:
     content: aziotctl-executables
     source:
       read: [ $SNAP/bin ]
+  aziot-keys-library:
```
arsing, line 167: I suggest calling it `aziot-keys-openssl-engine` so as to not confuse it with the other "aziot-keys library" that is the implementation used by keyd to interact with hardware.

arsing (approved):

snap/snapcraft.yaml
damonbarry: Updated

PR #651, opened 2025-05-08
Add support for handling network errors from parent edgeHub

gordonwang0:
In nested mode, child devices are not able to detect if the network between the parent and IoT Hub is down. This leads to some incorrect behavior where the child assumes IoT Hub is available during its startup. edgeHub will return a 500 Internal Server Error in this case. This PR adds handling of 500 Internal Server Errors from edgeHub as network errors so identity service starts correctly when connection to IoT Hub is down.

Also improves the logging of HTTP errors by logging the HTTP error code instead of a JSON parse error.

### DIFF
--- a/http-common/src/lib.rs
+++ b/http-common/src/lib.rs
@@ -27,7 +27,7 @@ mod proxy;
 pub use proxy::{get_proxy_uri, MaybeProxyConnector};
 
 mod request;
-pub use request::HttpRequest;
+pub use request::{HttpRequest, HttpResponse};
 
 pub mod server;
 
--- a/http-common/src/request.rs
+++ b/http-common/src/request.rs
@@ -506,7 +506,7 @@ mod tests {
 
     #[derive(Debug, serde::Deserialize)]
     struct HubError {
-        #[serde(rename = "Message", alias = "errorMessage")]
+        #[serde(rename = "Message")]
         pub message: String,
     }
 
--- a/http-common/src/request.rs
+++ b/http-common/src/request.rs
@@ -293,6 +293,10 @@ pub struct HttpResponse {
 }
 
 impl HttpResponse {
+    pub fn status(&self) -> hyper::StatusCode {
+        self.status
+    }
+
     pub fn into_parts(self) -> (hyper::StatusCode, hyper::body::Bytes) {
         (self.status, self.body)
     }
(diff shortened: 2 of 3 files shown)


damonbarry (commented):
Looks good. One question: 500 is a pretty broad error, are you confident Edge hub only returns 500 when it can't connect upstream to IoT Hub?

gordonwang0:
I tested some other things like authentication errors and incorrect URLs. Those return 4xx errors. I think it's fine to have identity service treat the service as offline and continue on 5xx errors.

damonbarry:
> Looks good. One question: 500 is a pretty broad error, are you confident Edge hub only returns 500 when it can't connect upstream to IoT Hub? 

@vipeller @ancaantochi can either of you confirm? Does Edge hub ever return 500 to a downstream edge device for anything other than when it can't connect upstream? I have a vague recollection that Edge hub will return 500 for _any_ condition that it doesn't have a more specific error code for. I want to make sure that error code == 500 from Edge Hub is the right way to detect an upstream network error.

ancaantochi:
> > Looks good. One question: 500 is a pretty broad error, are you confident Edge hub only returns 500 when it can't connect upstream to IoT Hub?
> 
> @vipeller @ancaantochi can either of you confirm? Does Edge hub ever return 500 to a downstream edge device for anything other than when it can't connect upstream? I have a vague recollection that Edge hub will return 500 for _any_ condition that it doesn't have a more specific error code for. I want to make sure that error code == 500 from Edge Hub is the right way to detect an upstream network error.

There is no guarantee that 500 is returned only when not connected. There could be 500 coming from IotHub or exceptions in Edgehub which return 500. Unauthenticated or bad request is 400, but there still could an issue which is not a disconnect and return 500.

vipeller:
Can it cause any problems extending request structure? Things I don't quite see:
- if anything else uses it (it is under common)
- if it can happen that somebody uses mixed versions (if more services use the request structure, but somehow one is old, one is new)
- even if that can happen, does it cause any problems?
  - if the receiver does not know about the new property (because it is an old version), json parsers just ignore that - so it should work as before (the receiver will not know about the network error, just like before - is it true)
  - if the receiver expects the new property, but the sender does not send it, again, it is just like before (the error will not be treated specially)

So, I think that:
- odds of mixed versions are low (or not possible)
- even if it can happen, then it just works as before the fix (so things will not get worse)

Is this correct?

PR #664, opened 2026-05-05
Update Rust to v1.95 and all deps to latest versions.

arsing:
This change updates the Rust toolchain to v1.95 and all dependencies to
their latest versions.

All crates have been updated to Edition 2024. Because of that and
the new toolchain, most of the diff is from rustfmt and clippy changes.

The other big part of the change is from hyper v1.0 having a reduced API
compared to v0.14. The missing API is now provided by the hyper-util and
http-body-util crates.

The original hyper-proxy crate is abandoned and doesn't support hyper v1.
hyper-proxy2 is one fork of it, updated to support hyper v1, but seems to have
since been abandoned also. Another fork hyper-http-proxy seems to be maintained
but it dropped openssl support (only supports rustls and native-tls) so
we cannot use it, since we need to be able to create the proxy with
an engine-backed TLS cert private key. native-tls does wrap openssl on Linux,
but intentionally does not provide a conversion from `openssl::SslConnector` to
`native_tls::TlsConnector`, so using hyper-proxy2 with native-tls is
not an option. So for now we use hyper-proxy2, but may need to re-evaluate
in the future.

We have three git submodules under third-party/ for TPM test dependencies -
tpm2-tss, libtpms and swtpm. I tried to update them to their latest releases,
however this breaks the build for RedHat OSes. The newer versions require
additional C library dependencies (json-c) but the redhat/ubi* containers
don't have the corresponding devel packages. So for now I've left them
unupdated.


tpm/tss-minimal/tests/evict_control.rs
```diff
@@ -1,18 +1,17 @@
 // Copyright (c) Microsoft. All rights reserved.
 
-#![deny(rust_2018_idioms)]
-#![warn(clippy::all, clippy::pedantic)]
-
 #[test]
-fn main() -> tss_minimal::Result<()> {
-    let _ = env_logger::builder().is_test(true).try_init();
+#[ignore = "TODO: Investigate why this fails"]
```
arsing: TODO

tpm/tss-minimal/tests/from_tpm_public.rs
```diff
@@ -1,18 +1,17 @@
 // Copyright (c) Microsoft. All rights reserved.
 
-#![deny(rust_2018_idioms)]
-#![warn(clippy::all, clippy::pedantic)]
-
 #[test]
-fn main() -> tss_minimal::Result<()> {
-    let _ = env_logger::builder().is_test(true).try_init();
+#[ignore = "TODO: Investigate why this fails"]
```
arsing: TODO

tpm/tss-minimal/tests/policy_secret.rs
```diff
@@ -1,11 +1,9 @@
 // Copyright (c) Microsoft. All rights reserved.
 
-#![deny(rust_2018_idioms)]
-#![warn(clippy::all, clippy::pedantic)]
-
 #[test]
-fn main() -> tss_minimal::Result<()> {
-    let _ = env_logger::builder().is_test(true).try_init();
+#[ignore = "TODO: Investigate why this fails"]
```
arsing: TODO

Cargo.toml
```diff
@@ -58,6 +57,125 @@ members = [
...
+hyper = { version = "1", features = [
+	"client",
+	"http1",
+	"server",
+] }
+hyper-openssl = { version = "0.10", features = [
+	"client-legacy",
+	"tokio",
+] }
+hyper-proxy2 = { version = "0.1", features = ["openssl-tls"], default-features = false }
```
arsing, line 101: The original hyper-proxy is abandoned and doesn't support hyper v1.

This hyper-proxy2 is one fork of it, updated to support hyper v1, but seems to have since been abandoned.

Another fork hyper-http-proxy seems to be maintained, but dropped openssl support (only supports rustls and native-tls) so we cannot use it. native-tls wraps openssl but (intentionally) does not provide a conversion from openssl `SslConnector` to native-tls `TlsConnector`.

So for now I've used hyper-proxy2, but we might need to fork it and merge any improvements from hyper-http-proxy, or convince hyper-http-proxy to add back openssl support, or something else.

Makefile
```diff
@@ -194,7 +194,7 @@ target/openapi-schema-validated:
```
arsing, line 206: (Noting this here just to track it.)

We have three git submodules under `third-party/` for TPM test dependencies - tpm2-tss, libtpms and swtpm. I tried to update them to their latest releases, however this breaks the build for RedHat OSes. The newer versions require additional C library dependencies (json-c) but the `redhat/ubi*` containers don't have the corresponding devel packages. So for now I've left them unupdated.

tpm/tss-minimal/tests/evict_control.rs
arsing: Passes in CI.

tpm/tss-minimal/tests/from_tpm_public.rs
arsing: Passes in CI.

tpm/tss-minimal/tests/policy_secret.rs
arsing: Passes in CI.

damonbarry (approved):

gordonwang0 (approved):

PR #665, opened 2026-06-03
Fix endless reparsing of config if it's modified once after startup.

arsing:
Since aziotd responds to config changes by reparsing the config. This reparsing would itself generate a notify events (access) after 5s, which would trigger aziotd to parse the config again, which would generate an access event, ad infinitum.

This regressed in 3fc8a1441971ee315dcbfa31a0e1e17e994ef8bf ; before that commit there was a filter to ignore access events but that commit accidentally removed the filter. This commit adds it back.

### DIFF
--- a/config-common/src/watcher.rs
+++ b/config-common/src/watcher.rs
@@ -3,7 +3,7 @@
 use std::path::PathBuf;
 
 use async_trait::async_trait;
-use notify_debouncer_mini::notify;
+use notify_debouncer_full::notify;
 
 #[async_trait]
 pub trait UpdateConfig {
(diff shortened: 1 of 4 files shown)


gordonwang0 (approved):

config-common/src/watcher.rs
```diff
@@ -58,32 +58,42 @@ pub fn start_watcher<TApi>(
...
                     .expect("Watching config file should not fail.");
             }
 
-            loop {
-                let event = file_watcher_rx.recv();
-                log::debug!("Incoming file watcher event: {:?}", &event);
-                if event.is_ok() {
-                    _ = file_changed_tx.blocking_send(());
+            while let Ok(events) = file_watcher_rx.recv() {
+                let Ok(events) = events else {
```
damonbarry, line 82: I don't understand the two `let Ok(events) = ...` statements in a row. Does `file_watcher_rx.recv()` return `Result<Result<_, _>, _>`?

arsing: We're thunking them through a channel, so the outer result is the channel receiver's, to indicate an error with receiving over the channel. Specifically that error happens if the sender is dropped, which shouldn't happen unless the notifier's internal background thread exits for some reason, in which case our receiving thread may as well exit too.

damonbarry: Makes sense, thanks!

damonbarry (approved):

PR #667, opened 2026-06-26
Block the ability to created nested derived keys.

arsing:
This was considered when we first implemented derived keys, but not fully implemented:

- `fn key_handle_to_id` was parsing such keys but throwing away all the intermediate derivation data, so the effect was that `derive(derive(base, A), B)` created the same key as `derive(base, B)`.

- libaziot-keys' API allows `AZIOT_KEYS_SIGN_DERIVED_PARAMETERS::mechanism` to be `AZIOT_KEYS_SIGN_MECHANISM_DERIVED` and thus chain arbitrarily, but the actual implementation in `derive_key_for_sign` only acted on the top layer.

We considered finishing the implementation, but there is no known use case for nested derivation. If the client really needs a nested derived key, they can import each derived key except for the last. In any case, given nested derivation did not work in the current code, we know that there cannot be any clients that depend on nested derivation behaving correctly.

So this change blocks the ability to create new nested derived key handles, and also blocks the usage of any existing nested derived key handles from being used with sign etc API.

Clients that depend on the broken behavior (that `derive(derive(base, A), B)` acts like `derive(base, B)` will be broken by this change, but we want such clients to be broken, because they should not have had the ability to generate `derive(base, B)` in this way.

### DIFF
--- a/key/aziot-keyd/src/lib.rs
+++ b/key/aziot-keyd/src/lib.rs
@@ -263,6 +263,16 @@ impl Api {
         base_handle: &aziot_key_common::KeyHandle,
         derivation_data: &[u8],
     ) -> Result<aziot_key_common::KeyHandle, Error> {
+        match key_handle_to_id(base_handle, &mut self.keys)?.0 {
+            KeyId::KeyPair(_) | KeyId::Key(_) => (),
+            KeyId::Derived(..) => {
+                return Err(Error::invalid_parameter(
+                    "handle",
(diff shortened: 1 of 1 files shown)


gordonwang0 (approved):

jlian:
This is blocking 1.6 LTS right? We should merge this?

arsing:
See my comment from last Friday in the Teams chat.

If it doesn't happen by EOD I'll merge it.

PR #674, opened 2026-09-15
Use Azure Linux prerequisite bootstrap

damonbarry:
## Description

Fixes Azure Linux 3.0 package builds after the Azure Linux toolkit increased its minimum required Go version from 1.24 to 1.25.

Previously, `ci/install-build-deps.sh` installed a hard-coded Go version before cloning the Azure Linux toolkit. This caused package builds to fail whenever the toolkit’s minimum Go version advanced beyond the locally pinned version.

This change:

- Removes the hard-coded Go toolchain installation.
- Clones the tracked Azure Linux `3.0-stable` branch first.
- Invokes the toolkit’s supported `install-prereqs-and-configure` target.
- Lets the toolkit select and install its supported Go version, including upstream checksum verification.
- Defaults `USER` to `root` because the bootstrap runs inside the packaging container.

The build will continue tracking `3.0-stable`, but future Go requirement updates on that branch will no longer require corresponding manual version bumps in this repository.

## Testing

- Verified `ci/install-build-deps.sh` with `bash -n`.
- Verified the patch with `git diff --check`.
- Confirmed `install-prereqs-and-configure` exists on the current Azure Linux `3.0-stable` branch and invokes the Ubuntu prerequisite bootstrap.


damonbarry:
Merging the changes here with #675 to get through the check-in gate. Closing...

